Issue #3, opened 2021-05-07
Can not find debug.so 


st0012:
Thanks for rewriting this cool lib! I wanted to give it a try but got this error:

```
❯ rdbg foo.rb
/Users/st0012/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/debug-1.0.0.beta1/lib/debug/session.rb:7:in `require_relative': cannot load such file -- /Users/st0012/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/debug-
1.0.0.beta1/lib/debug/debug.so (LoadError)
        from /Users/st0012/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/debug-1.0.0.beta1/lib/debug/session.rb:7:in `<top (required)>'
        from /Users/st0012/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/debug-1.0.0.beta1/lib/debug/console.rb:1:in `require_relative'
        from /Users/st0012/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/debug-1.0.0.beta1/lib/debug/console.rb:1:in `<top (required)>'
        from /Users/st0012/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/debug-1.0.0.beta1/lib/debug/run.rb:1:in `require_relative'
        from /Users/st0012/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/debug-1.0.0.beta1/lib/debug/run.rb:1:in `<top (required)>'
        from <internal:/Users/st0012/.rbenv/versions/3.0.1/lib/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from <internal:/Users/st0012/.rbenv/versions/3.0.1/lib/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
```

Not sure if it requires any additional setup or only runs on the latest version of Ruby?

### My Environment

- OS: `Darwin Kernel Version 20.3.0: Thu Jan 21 00:07:06 PST 2021; root:xnu-7195.81.3~1/RELEASE_X86_64`
- Ruby: `2.7.2` and `3.0.1` both got this error
- rubygems: `3.2.15`

PR #4, opened 2021-05-08
Use recommended way to name/load c extension

st0012:
This is to solve #3 by using [Rubygems' recommended approach](https://guides.rubygems.org/gems-with-extensions) to name/load this gem's c extension.

Summary of the commit:

1. Use `require` instead of `require_relative` to load the extension file.
2. Use the recommended way to name the extension in `extconf.rb`.

According to Rubygems' guideline, we should use `"debug/debug"` when calling `create_makefile`. Then we can use `require "debug/debug" to require the extension.

https://guides.rubygems.org/gems-with-extensions/#extension-naming

Currently, `create_makefile "debug"` means that the require path would be `require "debug"`, which confuses Ruby. As a result it'd always load `debug.rb` instead of the compiled extension.

**Test Command**

```
$ gem build debug.gemspec; gem install --local *.gem; rdbg ext/debug/extconf.rb
```

**Before This Commit**

```
gem build debug.gemspec; gem install --local *.gem; rdbg ext/debug/extconf.rb
  Successfully built RubyGem
  Name: debug
  Version: 1.0.0.beta1
  File: debug-1.0.0.beta1.gem
Building native extensions. This could take a while...
Successfully installed debug-1.0.0.beta1
Done installing documentation for debug after 0 seconds
1 gem installed
/Users/st0012/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/debug-1.0.0.beta1/lib/debug/session.rb:7:in `require_relative': cannot load such file -- /Users/st0012/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/debug-1.0.0.beta1/lib/debug/debug.so (LoadError)
```

**After This Commit**

```
gem build debug.gemspec; gem install --local *.gem; rdbg ext/debug/extconf.rb
  Successfully built RubyGem
  Name: debug
  Version: 1.0.0.beta1
  File: debug-1.0.0.beta1.gem
Building native extensions. This could take a while...
Successfully installed debug-1.0.0.beta1
Done installing documentation for debug after 0 seconds
1 gem installed
[1, 2] in ext/debug/extconf.rb
=>    1| require 'mkmf'
      2| create_makefile 'debug/debug'
--> #0  ext/debug/extconf.rb:1:in `<main>'

(rdbg)
```

### DIFF
--- a/ext/debug/extconf.rb
+++ b/ext/debug/extconf.rb
@@ -1,2 +1,2 @@
 require 'mkmf'
-create_makefile 'debug'
+create_makefile 'debug/debug'
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -4,7 +4,7 @@
                          :has_return_value, :return_value, :show_line)
 end
 
-require_relative 'debug.so'
+require "debug/debug"
 
 require_relative 'source_repository'
 require_relative 'breakpoint'


ko1:
> Use require instead of require_relative to load the extension file.

Could you explain about it more?
It is useful to try the lib without installing it.

ko1:
Ah, `.so` is not located in the `require_relative` require-able path?

ko1:
Thank you!!

st0012:
@ko1 thanks for merging and releasing it so quickly. I can confirm that the beta2 version installs and runs correctly on my machine 🙂

PR #7, opened 2021-05-12
Refine exit commands

### LINKED ISSUE #6: Quick exit command
Currently, all exit/quit/kill commands require an additional confirmation step.

https://github.com/ruby/debug/blob/4a3046ba87f573f2c21cce590bce206e469dba76/lib/debug/session.rb#L236-L251

```
❯ rdbg foo.rb
[1, 2] in foo.rb
=>    1| a = 1
      2| puts(a)
--> #0  foo.rb:1:in `<main>'

(rdbg) exit
Really quit? [Y/n] Y
```

```
❯ rdbg foo.rb
[1, 2] in foo.rb
=>    1| a = 1
      2| puts(a)
--> #0  foo.rb:1:in `<main>'

(rdbg) quit!
Really kill? [Y/n] Y
```

 Without changing any existing commands, I wonder if it's ok to add the `exit!` command? It'll allow users to quickly exit the current session without hitting `Y` to confirm it.

```
❯ exe/rdbg foo.rb
[1, 1] in foo.rb
=>    1| puts(1)
=>#0    <main> at foo.rb:1

(rdbg) exit!
```

st0012:
Closes #6

### `q`

```
❯ exe/rdbg foo.rb
[1, 1] in foo.rb
=>    1| puts(1)
=>#0    <main> at foo.rb:1

(rdbg) q
Really quit? [Y/n] Y
```

### `quit`
```
❯ exe/rdbg foo.rb
[1, 1] in foo.rb
=>    1| puts(1)
=>#0    <main> at foo.rb:1

(rdbg) quit
Really quit? [Y/n] Y
```

### `q!`
```
❯ exe/rdbg foo.rb
[1, 1] in foo.rb
=>    1| puts(1)
=>#0    <main> at foo.rb:1

(rdbg) q!
```

### `quit!`
```
❯ exe/rdbg foo.rb
[1, 1] in foo.rb
=>    1| puts(1)
=>#0    <main> at foo.rb:1

(rdbg) quit!
```

### `kill`
```
❯ exe/rdbg foo.rb
[1, 1] in foo.rb
=>    1| puts(1)
=>#0    <main> at foo.rb:1

(rdbg) kill
Really kill? [Y/n] Y
```

### `kill!`
```
❯ exe/rdbg foo.rb
[1, 1] in foo.rb
=>    1| puts(1)
=>#0    <main> at foo.rb:1

(rdbg) kill!
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,9 +321,11 @@ The `<...>` notation means the argument.
   * Finish this frame. Resume the program until the current frame is finished.
 * `c[ontinue]`
   * Resume the program.
-* `q[uit]` or exit or `Ctrl-D`
+* `q[uit]` or `Ctrl-D`
   * Finish debugger (with the debuggee process on non-remote debugging).
-* `kill` or `q[uit]!`
+* `q[uit]!`
+  * Finish debugger immediately (with the debuggee process on non-remote debugging).
+* `kill(!)`
   * Stop the debuggee process.
 
 ### Breakpoint
--- a/README.md
+++ b/README.md
@@ -324,9 +324,11 @@ The `<...>` notation means the argument.
 * `q[uit]` or `Ctrl-D`
   * Finish debugger (with the debuggee process on non-remote debugging).
 * `q[uit]!`
-  * Finish debugger immediately (with the debuggee process on non-remote debugging).
-* `kill(!)`
-  * Stop the debuggee process.
+  * Same as q[uit] but without the confirmation prompt.
+* `kill`
+  * Stop the debuggee process with `Kernal#exit!`.
+* `kill!`
+  * Same as kill but without the confirmation prompt.
 
 ### Breakpoint
 
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -256,6 +256,11 @@ module DEBUGGER__
           return :retry
         end
 
+      # * `kill!`
+      #   * Stop the debuggee process immediately.
+      when 'kill!'
+        exit! (arg || 1).to_i
+
       ### Breakpoint
 
       # * `b[reak]`
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -242,13 +242,13 @@ module DEBUGGER__
         end
 
       # * `q[uit]!`
-      #   * Finish debugger immediately (with the debuggee process on non-remote debugging).
+      #   * Same as q[uit] but without the confirmation prompt.
       when 'q!', 'quit!'
         @ui.quit arg.to_i
         @tc << :continue
 
       # * `kill`
-      #   * Stop the debuggee process.
+      #   * Stop the debuggee process with `Kernal#exit!`.
       when 'kill'
         if ask 'Really kill?'
           exit! (arg || 1).to_i
@@ -257,7 +257,7 @@ module DEBUGGER__
         end
 
       # * `kill!`
-      #   * Stop the debuggee process immediately.
+      #   * Same as kill but without the confirmation prompt.
       when 'kill!'
         exit! (arg || 1).to_i
 
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -241,9 +241,15 @@ module DEBUGGER__
           return :retry
         end
 
-      # * `kill` or `q[uit]!`
+      # * `q[uit]!`
+      #   * Finish debugger immediately (with the debuggee process on non-remote debugging).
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -321,9 +321,9 @@ The `<...>` notation means the argument.
   * Finish this frame. Resume the program until the current frame is finished.
 * `c[ontinue]`
   * Resume the program.
-* `q[uit]` or exit or `Ctrl-D`
+* `q[uit](!)` or `Ctrl-D`
```
ko1, line 324: could you explain about `!` command?
I prefer to make separate lines like:

```
* q[uit] or ...
  * Finish ..
 * q[uit]!
   * Finish ...
```

st0012: 👍 updated

README.md
```diff
@@ -321,9 +321,11 @@ The `<...>` notation means the argument.
   * Finish this frame. Resume the program until the current frame is finished.
 * `c[ontinue]`
   * Resume the program.
-* `q[uit]` or exit or `Ctrl-D`
+* `q[uit]` or `Ctrl-D`
   * Finish debugger (with the debuggee process on non-remote debugging).
-* `kill` or `q[uit]!`
+* `q[uit]!`
+  * Finish debugger immediately (with the debuggee process on non-remote debugging).
+* `kill(!)`
```
ko1, line 328: `rake` command update the doc :)

lib/debug/session.rb
```diff
@@ -231,25 +231,36 @@ def wait_command
...
 
-      # * `kill` or `q[uit]!`
+      # * `q[uit]!`
+      #   * Finish debugger immediately (with the debuggee process on non-remote debugging).
+      when 'q!', 'quit!'
+        @ui.quit arg.to_i
+        @tc << :continue
+
+      # * `kill`
       #   * Stop the debuggee process.
```
ko1, line 251: This is my fault, but could you change the description about `kill` command?
Maybe the word "immediately" fits for this command. For example,

"Stop the debugger process immediately with `exit!` method"

is clear for Rubyist (I think rubyists knows what `exit!` method mean. ... is it ok?)

So the doc for `quit!` should be:

```
* q[uit]!
  * Same as `quit` command, but without asking prompt. <- please correct my poor English please :p
```

and so on will be clear.

ko1: maybe `q!` is familiar with vim's command for programmers.

README.md
st0012: wow this is awesome!

ko1:
Thank you!

lib/debug/session.rb
```diff
@@ -231,25 +231,36 @@ def wait_command
...
-      #   * Stop the debuggee process.
-      when 'kill', 'quit!', 'q!'
+      # * `q[uit]!`
+      #   * Same as q[uit] but without the confirmation prompt.
+      when 'q!', 'quit!'
+        @ui.quit arg.to_i
+        @tc << :continue
+
+      # * `kill`
+      #   * Stop the debuggee process with `Kernal#exit!`.
```
st0012, line 251: @ko1 I think it's `Kernel#exit!`, right?

I can't speak for other people, but as a Rubyist with 8 years of experience, I don't remember I've ever used the `exit!` method in my codebase. I think for people who use Ruby for scripting they should be familiar with it. But Rails or web-focused Ruby developers may not be familiar with this method either. 

TBH, when I saw `exit` the first thing that came into my mind was the `exit` command in `pry` or `irb`. That's why I initially proposed `exit!` in the issue 😂

ko1: BTW, maybe pry and irb's `exit` command is only eval "exit" string and calls `Kernel#exit` method by eval.

st0012: In `pry`, there's an [exit command class](https://github.com/pry/pry/blob/0aae8c94ad03a732659ed56dcd5088469a15eebf/lib/pry/commands/exit.rb), which I always thought is the source of the `exit` command. But when I check the source of `exit` inside the repl, I can only see 

```
> show-source exit

From: process.c (C Method):
Owner: Kernel
Visibility: private
Signature: exit(*arg1)
Number of lines: 5

static VALUE
f_exit(int c, const VALUE *a, VALUE _)
{
    return rb_f_exit(c, a);
}
```

This is mind blowing 🤯

ko1: Ah, my mistake :p

PR #8, opened 2021-05-12
Ignore all bundle files

st0012:
Currently, running `rake` or `rake compile` would generate `lib/debug/debug.bundle`.
Those files are generated by the extension complication and should not be committed into the codebase.


ko1:
and also `.so` files.

PR #9, opened 2021-05-14
[WIP] Trial implementation of test code for debbuger

ono-max:


test/minitest_runner.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+$LOAD_PATH.unshift(__dir__)
+
+require 'minitest'
```
hsbt, line 5: Can you use "test-unit" instead of `minitest` ?

`ruby/ruby` migrates `minitest` to `test-unit` in this year.

ono-max: @hsbt 
Could you tell me why `ruby/ruby` migrates `minitest` to `test-unit`?

hsbt: see https://github.com/rubygems/rubygems/pull/3141

Basically, `ruby`ruby` uses `test-unit`, not `minitest`. rubygems and psych only uses minitest namespace. I did migrate them to `test-unit` in May, 2021.

ono-max: Thank you!

ono-max: 6d42955

PR #10, opened 2021-05-14
Handle evaled source code

casperisfine:
Fixes the issue reported here: https://bugs.ruby-lang.org/issues/17863#note-1

Otherwise it would fail with:

```
No such file or directory @ rb_sysopen - eval (Errno::ENOENT)
```

I could explictly check for `"eval"` but I think it's better to handle all potential file system issues, because eval interfaces allow to provide arbitrary paths.

cc @ko1

### DIFF
--- a/lib/debug/source_repository.rb
+++ b/lib/debug/source_repository.rb
@@ -10,7 +10,10 @@ module DEBUGGER__
 
       case
       when path = iseq.absolute_path
-        src = File.read(path)
+        src = begin
+          File.read(path)
+        rescue SystemCallError
+        end
       when iseq.path == '-e'
         path = '-e'
       else


ko1:
Thank you for the report.

I'm not sure how to reproduce it. Maybe it is related to the `eval`, but I couldn't reproduce the error with step-in into eval-source code and `binding.bp`. Could you share the small repro-code?

casperisfine:
```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem "debug", "1.0.0.beta3"
end

module Foo
  module_eval <<~RUBY, '--eval--', 1
    def self.bar
      binding.bp
    end
  RUBY
end

Foo.bar
```

```
#<Thread:0x00007fc65f04d5f8 /Users/byroot/.gem/ruby/3.0.1/gems/debug-1.0.0.beta3/lib/debug/session.rb:78 run> terminated with exception (report_on_exception is true):
/Users/byroot/.gem/ruby/3.0.1/gems/debug-1.0.0.beta3/lib/debug/source_repository.rb:13:in `read': No such file or directory @ rb_sysopen - --eval-- (Errno::ENOENT)
	from /Users/byroot/.gem/ruby/3.0.1/gems/debug-1.0.0.beta3/lib/debug/source_repository.rb:13:in `add'
	from /Users/byroot/.gem/ruby/3.0.1/gems/debug-1.0.0.beta3/lib/debug/session.rb:825:in `on_load'
	from /Users/byroot/.gem/ruby/3.0.1/gems/debug-1.0.0.beta3/lib/debug/session.rb:88:in `block in initialize'
["/Users/byroot/.gem/ruby/3.0.1/gems/debug-1.0.0.beta3/lib/debug/thread_client.rb",
 599,
 #<Errno::ENOENT: No such file or directory @ rb_sysopen - --eval-->,
 ["/Users/byroot/.gem/ruby/3.0.1/gems/debug-1.0.0.beta3/lib/debug/source_repository.rb:13:in `read'",
  "/Users/byroot/.gem/ruby/3.0.1/gems/debug-1.0.0.beta3/lib/debug/source_repository.rb:13:in `add'",
  "/Users/byroot/.gem/ruby/3.0.1/gems/debug-1.0.0.beta3/lib/debug/session.rb:825:in `on_load'",
  "/Users/byroot/.gem/ruby/3.0.1/gems/debug-1.0.0.beta3/lib/debug/session.rb:88:in `block in initialize'"]]
/Users/byroot/.gem/ruby/3.0.1/gems/debug-1.0.0.beta3/lib/debug/source_repository.rb:13:in `read': No such file or directory @ rb_sysopen - --eval-- (Errno::ENOENT)
	from /Users/byroot/.gem/ruby/3.0.1/gems/debug-1.0.0.beta3/lib/debug/source_repository.rb:13:in `add'
	from /Users/byroot/.gem/ruby/3.0.1/gems/debug-1.0.0.beta3/lib/debug/session.rb:825:in `on_load'
	from /Users/byroot/.gem/ruby/3.0.1/gems/debug-1.0.0.beta3/lib/debug/session.rb:88:in `block in initialize'
```

ko1:
Thank you, I could reproduce and I solved with other patch.

PR #11, opened 2021-05-14
Handle bare commands as print

casperisfine:
That's how byebug and many other debuggers currently behave, and IMHO is a much nicer experience.

```
(rdbg) defined? Kernel
=> "constant"
```

### DIFF
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -552,9 +552,7 @@ module DEBUGGER__
 
       ### END
       else
-        @ui.puts "unknown command: #{line}"
-        @repl_prev_line = nil
-        return :retry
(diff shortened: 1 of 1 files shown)


casperisfine:
Hum, actually my implementation don't work :/

casperisfine:
Ok, I fixed it.

ko1:
This is intentional.

> old debug.rb evaluate unrecognized command as Ruby script, but new debug.rb doesn't allow this spec.

casperisfine:
> This is intentional.

I know, I just hoped you'd reconsider as to make the transition from the existing popular debugger smoother for users. But that's your call. 

Thanks for this new debugger either way.

PR #12, opened 2021-05-14
Update rake for Ruby 3.0 compatibility

casperisfine:
```
$ bundle exec rake
Failed to remove lib/debug/debug.bundle: wrong number of arguments (given 2, expected 1)
Failed to remove tmp: wrong number of arguments (given 2, expected 1)
rake aborted!
ArgumentError: wrong number of arguments (given 2, expected 1)
~/.gem/ruby/3.0.1/gems/rake-12.3.3/exe/rake:27:in `<top (required)>'
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,6 @@ source 'https://rubygems.org'
 
 gemspec
 
-gem "rake", "~> 12.0"
+gem "rake"
 gem "rake-compiler"
 gem "minitest", "~> 5.0"


Gemfile
```diff
@@ -2,6 +2,6 @@ source 'https://rubygems.org'
 
 gemspec
 
-gem "rake", "~> 12.0"
```
st0012, line 5: shouldn't we change this to `~> 13.0` instead of removing the version restriction completely?

casperisfine: I've done that update on so many repos by now that I know think that this kind of development dependencies shouldn't be locked.

But that's @ko1's call.

hsbt: It should be set `>= 13.0` if we choose the explicitly versioning. We ignore the pessimistic versioning for reducing the maintenance cost.

ko1:
Thank you!

PR #15, opened 2021-05-17
Require console for UI_Console

adam12:
Fixes scenarios where session isn't started immediately.

    require 'debug/session'
    DEBUGGER__.console # raises NameError

### DIFF
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -935,6 +935,7 @@ module DEBUGGER__
   end
 
   def self.console
+    require_relative 'console'
     initialize_session UI_Console.new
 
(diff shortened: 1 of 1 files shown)


lib/debug/session.rb
```diff
@@ -14,6 +14,7 @@
 require_relative 'breakpoint'
 require_relative 'thread_client'
 require_relative 'config'
+require_relative 'console'
```
ko1, line 17: Could you move it to the beginning of `def self.console`?

adam12: Done.

ko1:
Thanks!

PR #17, opened 2021-05-18
Extract frame info computation out of TheadClient

st0012:
I know it's probably risky to do such refactoring without test cases but I couldn't help it 😄

Besides reducing the length of `thread_client.rb`, once we move frame string computation out of `ThreadClient`, we can eliminate many method parameters and reduce variable declarations.

(This is a proof of concept so I don't extract all the methods yet.)

### DIFF
--- a/lib/debug/frame_info.rb
+++ b/lib/debug/frame_info.rb
@@ -62,7 +62,7 @@ module DEBUGGER__
       nil
     end
 
-    def to_string
+    def to_client_output
       loc_str = pretty_location
 
       if binding && iseq
--- a/lib/debug/frame_info.rb
+++ b/lib/debug/frame_info.rb
@@ -1,0 +1,97 @@
+module DEBUGGER__
+  class FrameInfo
+    HOME = ENV['HOME'] ? (ENV['HOME'] + '/') : nil
+
+    def path
+      location.path
+    end
+
+    def pretty_path
+      use_short_path = CONFIG[:use_short_path]
+
+      case
+      when use_short_path && path.start_with?(dir = RbConfig::CONFIG["rubylibdir"] + '/')
+        path.sub(dir, '$(rubylibdir)/')
+      when use_short_path && Gem.path.any? do |gp|
+          path.start_with?(dir = gp + '/gems/')
+        end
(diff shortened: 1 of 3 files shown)


lib/debug/session.rb
```diff
@@ -10,6 +11,7 @@
   require "debug/debug"
 end
 
+require_relative 'frame_info'
```
ko1, line 14: should be at thread_client.rb?

lib/debug/frame_info.rb
```diff
@@ -0,0 +1,97 @@
...
+      end
+    end
+
+    def get_singleton_class obj
+      obj.singleton_class # TODO: don't use it
+    rescue TypeError
+      nil
+    end
+
+    def to_string
```
ko1, line 65: `to_string` is not common name, I guess. Maybe `format` or something is better because it can accept configuration.

ko1 (commented):
Does it change the code or no-code change (almost copy)?

st0012:
@ko1 It does change some implementation, if that's what you mean. For example, some methods can now directly read `location` after moved into `FrameInfo`, so they don't need to take that from arguments anymore. 

But it shouldn't introduce any behavior change.

ko1:
I don't care about parameter/variable changes.

> But it shouldn't introduce any behavior change.

Cool.

lib/debug/frame_info.rb
st0012: I am thinking `to_s` because it's essentially the string form of the frame, wdyt? Another option I have is `to_client_output`.

st0012:
@ko1 btw, do you have any recommended flow to test the `ThreadClient`'s functionalities? right now I just invoke the debugger in my other Ruby projects and try:
- step a few frames
- use `bt` command in different frames
- call `info` a few times

but I think there's probably more I should test?

ko1:
I don't have good answer...

FYI: `rdbg -e 'b <location>;; c ;; bt ;; q!' -e c target.rb` is easy to try.

st0012:
@ko1 this is impressively convenient. thanks for the great tip!

ko1:
Thanks!

PR #18, opened 2021-05-19
Follow up refactoring of #17

st0012:
- Extract some remaining frame computation from `ThreadClient` to `FrameInfo`.
- Remove unnecessary local variable assignments.
- Encapsulate several `FrameInfo` helper methods.

It'll be easier to review by commits. And same as #17, this PR doesn't introduce any behavioral change.


ko1:
Thanks!

PR #21, opened 2021-05-20
Trial implementation of test code for debbuger -multi processes

ono-max:
This is the trial implementation of test code for debugger by using multi processes.

- Use module PTY to create pseudo terminals for debugger.
- Write test codes for `next` and `step` and `break`

### How to use

```
# all test will be executed
$ ruby bin/test-unit.rb

# specify test file
$ ruby bin/test-unit.rb debug/break_test.rb
```

### Demo
[Image from Gyazo](https://gyazo.com/2bbe47cb20b9a49f48a0c6c0ad146d9e)

### Reference

The following test code is based on [byebug ](https://github.com/deivid-rodriguez/byebug)project

### DIFF
--- a/bin/test-unit.rb
+++ b/bin/test-unit.rb
@@ -1,0 +1,9 @@
+base_dir = File.expand_path(File.join(File.dirname(__FILE__), ".."))
+lib_dir  = File.join(base_dir, "lib")
+test_dir = File.join(base_dir, "test")
+
+$LOAD_PATH.unshift(lib_dir)
+
+require 'test/unit'
+
+exit Test::Unit::AutoRunner.run(true, test_dir)
--- a/test/debug/break_test.rb
+++ b/test/debug/break_test.rb
@@ -1,0 +1,160 @@
+# frozen_string_literal: true
+
+require_relative '../support/test_case'
+
+module DEBUGGER__
+  #
+  # Tests adding breakpoints to methods
+  #
+  class BreakAtMethodsTest < TestCase
+    def program
+      <<~RUBY
+        module Foo
+          class Bar
+            def self.a
+              "hello"
+            end
+
+            def b(n)
+              2.times do
+                n
+              end
+            end
+          end
+          module Baz
+            def self.c
+              1
+            end
+          end
+          Bar.a
+          bar = Bar.new
+          bar.b(1)
+          Baz.c
+        end
+      RUBY
+    end
+
+    def test_break_with_namespaced_instance_method_stops_at_correct_place
+      debug_code(program) do
+        type 'break Foo::Bar#b'
+        type 'continue'
+        assert_line_num 8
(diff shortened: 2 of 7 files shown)


bin/minitest
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require_relative '../test/minitest_runner'
```
ko1, line 4: is this file needed?

ono-max: No, I will remove it.

ono-max: f592487

ono-max: Done

bin/test-unit.rb
```diff
@@ -0,0 +1,9 @@
+base_dir = File.expand_path(File.join(File.dirname(__FILE__), ".."))
+lib_dir  = File.join(base_dir, "lib")
+test_dir = File.join(base_dir, "test")
+
+$LOAD_PATH.unshift(lib_dir)
+
+require 'test/unit'
+
+exit Test::Unit::AutoRunner.run(true, test_dir)
```
ono-max, line 9: copy and paste codes in reference
https://test-unit.github.io/test-unit/en/file.how-to.html

test/debug/break_test.rb
```diff
@@ -0,0 +1,160 @@
...
+      debug_code(program) do
+        type 'break 3'
+        type 'continue'
+        assert_line_num 3
+        type 'quit'
+        type 'y'
+      end
+    end
+  end
+end
```
ono-max, line 160: I refer to codes in byebug.
https://github.com/deivid-rodriguez/byebug/blob/master/test/commands/break_test.rb

test/support/utils.rb
```diff
@@ -0,0 +1,69 @@
...
+          read.expect(/(.*)\n|\(rdbg\)/) do |sentence|
+            print sentence[0]
+            if sentence[0] == '(rdbg)'
+              cmd = @queue.pop
+              if cmd.is_a?(Proc)
+                cmd.call(result)
+                cmd = @queue.pop
+              end
+              write.puts(cmd)
+              quitted = false if cmd == 'quit'
```
st0012, line 47: just curious about what `quitted` means here. I thought if `quitted == true`, the terminal would stop. but it looks like it's the opposite?

ono-max: Oh, that was my mistake. Thanks!

ko1:
@deivid-rodriguez

Hi, we made a test based on your byebug tests.
Now we write your license in a comment.
How about that? Or is it enough "thanks" comment in README.md?

Thank you!

deivid-rodriguez:
README thanks is enough! :)

ko1:
> README thanks is enough! :)

Thank you for your quick reply!

Issue #13, opened 2021-05-16
Proposal on backtrace (bt command) format


st0012:
I really like the useful information added to the backtrace (like arguments, block parameters...etc.). It's something I've been looking for for a long time.

And because I saw @ko1 asking for feedback on the backtrace format [in his tweet](https://twitter.com/_ko1/status/1391682538020630529), I think I can provide my 2 cents here.

### Example

This is the example script I'll use to generate the output for my proposal:

```ruby
class Foo
  def first_call
    second_call(20)
  end

  def second_call(num)
    third_call_with_block do |ten|
      forth_call(num, ten)
    end
  end

  def third_call_with_block(&block)
    @ivar1 = 10; @ivar2 = 20

    yield(10)
  end

  def forth_call(num1, num2)
    require "debug" # <= breakpoint
  end
end

Foo.new.first_call
```

And this is the output it generates for the `bt` command.

```
=>#0    Foo#forth_call(num1=20, num2=10) at foo.rb:21 #=> true
  #1    block{|ten=10|} in second_call at foo.rb:8
  #2    Foo#third_call_with_block(block=#<Proc:0x00007fc645174d10 foo.rb:7>) at foo.rb:15
  #3    Foo#second_call(num=20) at foo.rb:7
  #4    first_call at foo.rb:3
  #5    <main> at foo.rb:24
```

### Introduction of Components

I know most people reading this issue are already contributor to it and are familiar with the information contained in the backtrace. But I still want to do a short introduction to those who're not that familiar with it yet.

#### Trace Type - Call With Arguments

```
=>#0    Foo#forth_call(num1=20, num2=10) at foo.rb:21 #=> true
```

- `=>#0` - prompt + frame index
- `Foo#forth_call` - class + callee
- `(num1=20, num2=10)` - args
- `at foo.rb:21` - location
- `#=> true` - return value

#### Trace Type - Block Evaluation

```
  #1    block{|ten=10|} in second_call at foo.rb:8
```

- `#1` - prompt + frame index
- `block{|ten=10|} in second_call` - block label & parameters
- `at foo.rb:8` - location

## So, the Format

Ok, let's back to the backtrace:

```
=>#0    Foo#forth_call(num1=20, num2=10) at foo.rb:21 #=> true
  #1    block{|ten=10|} in second_call at foo.rb:8
  #2    Foo#third_call_with_block(block=#<Proc:0x00007fc645174d10 foo.rb:7>) at foo.rb:15
[...]

PR #23, opened 2021-05-22
Introduce BacktraceFormatter

st0012:
This class will help us format/colorize the backtrace more easily. These are the responsibilities of each class involved in backtrace display:

- `ThreadClient`
  - Retrieve frames
  - Compute line prefix and combine it with the trace to become a full backtrace line
  - Print backtrace lines
- `FrameInfo`
  - Store each frame's information
  - Compute components' string representation (currently, `return_value_str`, `call_identifier_str`, and `location_str`)
- `BacktraceFormatter`
  - Assemble formatted traces with `FrameInfo`'s component strings
  - Colorize components (demonstration PR: https://github.com/st0012/debug/pull/1)

**This change doesn't change the output**

```
❯ exe/rdbg -e 'b 20;; c ;; bt ;; info ;; q!' -e c target.rb
[1, 10] in target.rb
=>    1| class Foo
      2|   def first_call
      3|     second_call(20)
      4|   end
      5|
      6|   def second_call(num)
      7|     third_call_with_block do |ten|
      8|       forth_call(num, ten)
      9|     end
     10|   end
=>#0    <main> at target.rb:1
(rdbg:init) b 20
#1 line bp /Users/st0012/projects/debug/target.rb:20 (return)
(rdbg:init) c
[15, 23] in target.rb
     15|     yield(10)
     16|   end
     17|
     18|   def forth_call(num1, num2)
     19|     num1 + num2
=>   20|   end
     21| end
     22|
     23| Foo.new.first_call
=>#0    Foo#forth_call(num1=20, num2=10) at target.rb:20 #=> 30
  #1    block{|ten=10|} in second_call at target.rb:8
  # and 4 frames (use `bt' command for all frames)

Stop by #1 line bp /Users/st0012/projects/debug/target.rb:20 (return)
(rdbg:init) bt
=>#0    Foo#forth_call(num1=20, num2=10) at target.rb:20 #=> 30
  #1    block{|ten=10|} in second_call at target.rb:8
  #2    Foo#third_call_with_block(block=#<Proc:0x00007fbc021380d8 target.rb:7>) at target.rb:15
  #3    Foo#second_call(num=20) at target.rb:7
  #4    first_call at target.rb:3
  #5    <main> at target.rb:23
(rdbg:init) info
=>#0    Foo#forth_call(num1=20, num2=10) at target.rb:20 #=> 30
 %self => #<Foo:0x00007fbc02138380>
 %return => 30
 num1 => 20
 num2 => 10
 @ivar1 => 10
 @ivar2 => 20
(rdbg:init) q!
```

### DIFF
--- a/lib/debug/backtrace_formatter.rb
+++ b/lib/debug/backtrace_formatter.rb
@@ -1,0 +1,27 @@
+module DEBUGGER__
+  class BacktraceFormatter
+    attr_reader :frames
+
+    def initialize(frames)
+      @frames = frames
+    end
+
+    def formatted_traces(max)
+      traces = []
+      max += 1 if @frames.size == max + 1
+      max.times do |i|
+        break if i >= @frames.size
+        traces << formatted_trace(i)
+      end
+
+      traces
+    end
+
+    def formatted_trace(i)
+      frame = @frames[i]
+      result = "#{frame.call_identifier_str} at #{frame.location_str}"
+      result += " #=> #{frame.return_str}" if frame.return_str
+      result
+    end
+  end
+end
--- a/lib/debug/backtrace_formatter.rb
+++ b/lib/debug/backtrace_formatter.rb
@@ -20,7 +20,7 @@ module DEBUGGER__
     def formatted_trace(i)
       frame = @frames[i]
       result = "#{frame.call_identifier_str} at #{frame.location_str}"
-      result += " #=> #{frame.return_str}" if frame.return_str
+      result += " #=> #{frame.return_value_str}" if frame.return_value_str
       result
     end
   end
--- a/lib/debug/frame_info.rb
+++ b/lib/debug/frame_info.rb
@@ -63,7 +63,7 @@ module DEBUGGER__
       end
     end
 
-    def return_str
+    def return_value_str
       if binding && iseq && has_return_value
         short_inspect(return_value)
       end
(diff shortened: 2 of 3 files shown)


ko1:
To me it seems like overkill to introduce a new class.

st0012:
@ko1 I see. But would you consider accepting colorization I've tests in this PR? https://github.com/st0012/debug/pull/1

st0012:
Also, given the amount of frame information we have, it'd be better to have a formatter class extracted for customization like mentioned in https://github.com/ruby/debug/issues/13#issuecomment-845889708.

Personally, I'm very interested in maintain this area (presenting frame information) and I've been doing similar work in [my own project](https://github.com/st0012/object_tracer) for more than a year.

ko1:
* Proc should be enough for custom fornatter
* Coloring should be used by other features, so should not be in this specific class.

st0012:
> Proc should be enough for custom fornatter

I thought a class was necessary because some formats requires computation on all the frames, like the `With Separator (Aligned)` format proposed in #13. But since we probably won't use it as the default format, I agree that using a class seems redundant 👍 

> Coloring should be used by other features, so should not be in this specific class.

Yeah that seems more reasonable. Perhaps we can reuse the coloring logic implemented in `irb`? I've opened https://github.com/ruby/debug/pull/24 as an example.

PR #24, opened 2021-05-23
Use IRB's coloring helper to colorize source

st0012:
Since `IRB` already has coloring helpers, I figure it'd better to use it instead of having this library's own one? If this is a reasonable approach, I can use the same helper to colorize the backtrace in another PR. Some notes:

1. Because this library will be Ruby `3.1`'s default, the `irb` should always have the coloring feature (added after `2.7`).
2. I assume that if the debugger can start, the frame's source file must be "complete". So I choose the default complete mode.
 
Result:
image

### DIFF
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -1,5 +1,6 @@
 require 'objspace'
 require 'pp'
+require 'irb'
 require_relative 'frame_info'
 
 module DEBUGGER__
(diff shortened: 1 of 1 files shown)


ko1:
Thanks.

* The source should be cached. I'll fix it.
* This should be configurable. I'll fix it.
* There is `NO_COLOR` rule https://no-color.org/ but not sure we support it. At least, I'm discussing irb devs.

ko1:
image

irb's coloring doesn't support crlf line break. I'll fix it too.

st0012:
> The source should be cached

yeah I think we can store the colorized lines inside the frame. is that what you're thinking too? if that sounds good I can implement it.

> This should be configurable

Will adding the `--colorize` flag (like `irb`) and `RUBY_DEBUG_COLORIZE` env work?

> irb's coloring doesn't support crlf line break.

ah sorry I didn't even notice that it'd be a problem 😕

ko1:
Not sure `--nocolorize` is the best (`--no-xxx` is another manner).
Not so important, so wait to add it.

PR #25, opened 2021-05-23
Colorize frame info with irb coloring helper

st0012:
I figure the call identifier should use the same color as the identifier token (e.g. method name), so I used blue for it. The other 2 components' colors were arbitrarily picked.

image

### DIFF
--- a/lib/debug/frame_info.rb
+++ b/lib/debug/frame_info.rb
@@ -73,12 +73,6 @@ module DEBUGGER__
       "#{pretty_path}:#{location.lineno}"
     end
 
-    def to_client_output
-      result = "#{call_identifier_str} at #{location_str}"
-      result += " #=> #{return_str}" if return_str
-      result
-    end
-
     private
 
     SHORT_INSPECT_LENGTH = 40
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -13,6 +13,12 @@ module DEBUGGER__
 
     attr_reader :location, :thread, :mode, :id
 
+    DEFAULT_FRAME_FORMATTER = lambda do |frame|
+      result = "#{frame.call_identifier_str} at #{frame.location_str}"
+      result += " #=> #{frame.return_str}" if frame.return_str
+      result
+    end
+
     def initialize id, q_evt, q_cmd, thr = Thread.current
       @id = id
       @thread = thr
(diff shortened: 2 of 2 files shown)


st0012:
I accidentally deleted the branch from my fork 😅

ko1:
Could you survey pry and byebug for colors? If there is no problem the same colors help users.

st0012:
@ko1 sure 👍

st0012:
@ko1 I don't think `byebug` supports any kind of coloring. I don't see any color when I use it and I can't find anything about the color in its document. Here's its backtrace output, which is very similar to this project's but without locals.

byebug backtrace

As for `pry`, it only prints Ruby standard stacktrace and doesn't colorize it (which is not necessary imo).

pry-backtrace

ko1:
Thank you. Let's try!

ko1:
image

😢 

I'll check it.

ko1:
image
`[:BLUE, :BOLD]` helps me. it is same as method name.

ko1:
can we colorize parameters?
image

gdb sample:

image

ko1:
Do you have white background terminal like macos default?
It's on my putty/windows appearance with white backround:

image

st0012:
> can we colorize parameters?

yes, that's also my plan. I'd also colorize block identifier differently.

> Do you have white background terminal like macos default?

yes, I'll test different options and make an improvement on colors.

ko1:
RED file:line is too heavy for me (it makes me some alerts).
YELLOW is fine for me, and green is also fine.

st0012:
@ko1 ok, I'll keep it yellow and apply bold style on it too.

st0012:
@ko1 regarding coloring parameters, what do you think about this setup?

bright
dark

ko1:
image

on my putty terminal, bold yellow for file:line is too bright for me...

ko1:
How about green which uses gdb?

st0012:
@ko1 sure, let me change it 👍

st0012:
with a background that dark, bold green is still too bright to me

截圖 2021-05-27 下午12 58 03

but green without bold looks fine

截圖 2021-05-27 下午12 59 20

截圖 2021-05-27 下午1 00 11

it's still a bit dazzling with a white bright background, but I think we won't be able to find a perfect color without adding a new color code.
截圖 2021-05-27 下午1 00 00

PR #30, opened 2021-05-25
Store use_colorize config value in ivar like other options

st0012:
Just a small thing I found.


ko1:
I should check the CONFIG everytime to change the configuration in future debug command.

So other configuration should respect the configuration...

st0012:
@ko1 I see, thanks for the explanation 👍

PR #32, opened 2021-05-26
Colorize parameters

st0012:
### Method trace without arguments now also has class signature

**Before**

```
  #5    Foo#second_call(num=20) at target.rb:7
  #6    first_call at target.rb:3
```

**After**

```
  #5    Foo#second_call(num=20) at target.rb:7
  #6    Foo#first_call at target.rb:3
```

This change simplifies implementation and makes the trace look better imo.

###  Parameters are now colored differently 

截圖 2021-05-26 上午11 21 56
截圖 2021-05-26 上午11 24 16

## Implementation Details

- Because `FrameInfo` doesn't handle data representation, it shouldn't contain any colorization logic. Which means the `call_identifier_str` now needs to be assembled in `ThreadClient`.

### DIFF
--- a/lib/debug/frame_info.rb
+++ b/lib/debug/frame_info.rb
@@ -47,13 +47,13 @@ module DEBUGGER__
       if binding && iseq
         if iseq.type == :block
           if (argc = iseq.argc) > 0
-            args = parameters_info iseq.locals[0...argc]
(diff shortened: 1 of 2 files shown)


ko1:
Thanks

PR #34, opened 2021-05-27
Colorize output of info command

st0012:
截圖 2021-05-27 下午4 10 16
截圖 2021-05-27 下午4 11 42
截圖 2021-05-27 下午4 12 06


ko1:
I heard that there is an pretty print coloring feature in irb, so maybe we need to use it.

st0012:
@ko1 cool, let me check it 😄

PR #37, opened 2021-05-27
Use the gem's executable instead of system's for running test terminal

st0012:
When I ran tests locally with `bundle exec rake test`, I got

```
/Users/st0012/.rbenv/versions/3.0.1/lib/ruby/3.0.0/bundler/runtime.rb:302:in `check_for_activated_spec!': You have already activated debug 1.0.0.beta3, but your Gemfile requires debug 1.0.0.beta4. Prepending `bundle e
xec` to your command may solve this. (Gem::LoadError)
```

And it turned out that tests use the system's `debug` binary for running the pseudo terminal, which I think should be the gem folder's executable instead?

### DIFF
--- a/test/support/utils.rb
+++ b/test/support/utils.rb
@@ -28,7 +28,7 @@ module DEBUGGER__
     end
 
     def create_psuedo_terminal
-      PTY.spawn("ruby -r debug/run #{temp_file_path}") do |read, write, pid|
+      PTY.spawn("exe/rdbg #{temp_file_path}") do |read, write, pid|
         quit = false
         result = nil
 


st0012:
@ono-max do you think this is a correct solution? or I missed something in my local setup?

ono-max:
@st0012 
Can you try
`$ ruby bin/test-unit.rb` ?

ono-max:
> @ono-max do you think this is a correct solution? or I missed something in my local setup?

@st0012 
I can't reproduce the issue. Could you check your local setup again?

st0012:
@ono-max I think it's caused by I have multiple versions of this gem installed. I think you should be able to reproduce it by installing an older version like: `gem install debug -v 1.0.0.beta.3`.
I can solve the issue by removing older versions, but the better way should be to avoid this issue from the test setup.

ono-max:
Can you update debug gem?
It may help you.

st0012:
I think it's been fixed with https://github.com/ruby/debug/commit/db5a16a9090950e96bd22678bf7bbd45c5e13b6d
@ko1 thanks!

PR #39, opened 2021-05-28
Improve test setup

st0012:
few things I found when writing tests for the `backtrace` command:

- there's a typo: `psuedo` should be `pseudo`.
- the pseudo terminal should return full sentence and line number extraction should only happen in the matcher.
- the pseudo terminal should also disable coloring for better text matching.

### DIFF
--- a/test/support/utils.rb
+++ b/test/support/utils.rb
@@ -20,7 +20,7 @@ module DEBUGGER__
       @queue = Queue.new
       block.call
       write_temp_file(program)
-      create_psuedo_terminal
+      create_pseudo_terminal
     end
 
     def take_number(sentence)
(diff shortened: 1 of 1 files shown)


ko1:
Thanks.
@ono-max if you have any comments or questions, please ask it!

ko1:
BTW with @ono-max we talk about the design of this test and maybe it will be rewritten soon.

st0012:
should I halt the backtrace tests? I want to at least write test cases for each frame type first

st0012:
it looks like the current design is not able to read multiple lines of output, so I can only access the top trace. I guess I'll have to wait then 😅

ko1:
Absolutely.

PR #40, opened 2021-05-28
Colorize info output

st0012:
**Coloring Enabled**

colorized

**Coloring Disabled**

not colorized

### DIFF
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -27,13 +27,17 @@ module DEBUGGER__
       IRB::ColorPrinter.pp(obj, "")
     end
 
+    def colorize_cyan(str)
+      colorize(str, [:CYAN, :BOLD])
+    end
+
     def colorize_blue(str)
       colorize(str, [:BLUE, :BOLD])
     end
 
     def assemble_arguments(args)
       args.map do |arg|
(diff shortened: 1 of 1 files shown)


lib/debug/thread_client.rb
```diff
@@ -288,22 +301,24 @@ def show_by_editor path = nil
...
       end
       if current_frame&.has_return_value
-        puts " %return => #{current_frame.return_value}"
+        puts " #{colorize_cyan("%return")} => #{colored_inspect(current_frame.return_value)}"
       end
       if b = current_frame&.binding
         b.local_variables.each{|loc|
-          puts " #{loc} => #{b.local_variable_get(loc).inspect}"
+          value = b.local_variable_get(loc)
+          puts " #{colorize_cyan(loc)} => #{colored_inspect(value)}"
```
ko1, line 312: not strong opinion:

`colorize_cyan(loc)` is needed?
If value is colored, colored name seems verbose.

image

but gdb colored :)

st0012: since we use cyan for argument names in the trace, I think it'd be consistent to color info keys the same way too.
I didn't colorize them at first, but it's kinda odd to have them white like this 🤔 

截圖 2021-05-28 下午1 02 08

PR #41, opened 2021-05-29
Add tests for watch command

st0012:
I found several things when playing with the `watch` command:

- It mainly uses `line` events to detect the change instead of the `return` events (I'm not sure why though). This causes the program to always stop a line after the change. So in tests we need to add an extra line after the changed line. Otherwise it doesn't stop.
- It always evaluates the expression from top-level scope. So it's not able to track an instance variable inside an object like `watch @name` and users need to use `watch obj.name` instead. I think we should improve this.

But before we do anything on it, we should have some test cases to check its behavior. That's why I open this PR.

### DIFF
--- a/test/debug/watch_test.rb
+++ b/test/debug/watch_test.rb
@@ -1,0 +1,85 @@
+# frozen_string_literal: true
+
+require_relative '../support/test_case'
+
+module DEBUGGER__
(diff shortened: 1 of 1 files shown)


ko1:
Thanks.
Please file an issue for other points.

PR #42, opened 2021-05-29
display line numbers for test programs

ono-max:
Displaying line numbers in test programs is easy to write test code.

### Reference

This method is based on [byebug ](https://github.com/deivid-rodriguez/byebug)project

### DIFF
--- a/test/debug/break_test.rb
+++ b/test/debug/break_test.rb
@@ -9,28 +9,28 @@ module DEBUGGER__
   class BreakAtMethodsTest < TestCase
     def program
       <<~RUBY
-        module Foo
-          class Bar
-            def self.a
-              "hello"
-            end
-
-            def b(n)
-              2.times do
-                n
-              end
-            end
-          end
-          module Baz
-            def self.c
-              1
-            end
-          end
-          Bar.a
-          bar = Bar.new
-          bar.b(1)
-          Baz.c
-        end
+         1| module Foo
+         2|   class Bar
+         3|     def self.a
+         4|       "hello"
+         5|     end
+         6|
+         7|     def b(n)
+         8|       2.times do
+         9|         n
+        10|       end
+        11|     end
+        12|   end
+        13|   module Baz
+        14|     def self.c
+        15|       1
+        16|     end
+        17|   end
+        18|   Bar.a
+        19|   bar = Bar.new
+        20|   bar.b(1)
+        21|   Baz.c
+        22| end
       RUBY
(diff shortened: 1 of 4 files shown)


test/support/utils.rb
```diff
@@ -75,5 +75,9 @@ def create_pseudo_terminal
       p "terminal finished without reading 'y'"
       raise
     end
+
+    def strip_line_num(str)
+      str.gsub(/ *\d+\| ?/, '')
```
olleolleolle, line 80: Perhaps an anchor in the regular expression would help?

```suggestion
      str.gsub(/^\s*\d+\| ?/, '')
```

ono-max: Yes, you are right. Thanks!

st0012:
I think it's a great idea to have line number 👍 
But since it needs to be typed manually, I feel `:` is easier to type than `|` (no need to move my pinky finger that much). Just my personal preference though 🙂

test/debug/break_test.rb
```diff
@@ -8,29 +8,29 @@ module DEBUGGER__
...
-            def self.c
-              1
-            end
-          end
-          Bar.a
-          bar = Bar.new
-          bar.b(1)
-          Baz.c
-        end
+      strip_line_num <<~RUBY
```
st0012, line 11: for consistency reason, we may want **all** program code to have line numbers. and if that makes sense to you, maybe we can strip the line at a single place (perhaps in `debug_code`)? then we won't need to call this in every `program` method.

ko1:
> But since it needs to be typed manually, I feel `:` is easier to type than `|` (no need to move my pinky finger that much). Just my personal preference though 🙂

BTW do you prefer `:` for `list` command?
pry, irb uses `:` instead of `|`.

st0012:
> > But since it needs to be typed manually, I feel `:` is easier to type than `|` (no need to move my pinky finger that much). Just my personal preference though 🙂
> 
> BTW do you prefer `:` for `list` command?
> pry, irb uses `:` instead of `|`.

in that case I actually prefer `|`. but I'm ok with `:` for consistency reason as well.

test/debug/break_test.rb
ono-max: I agree with your opinion. Thank you!
https://github.com/ruby/debug/pull/42/files#diff-ee14bf65f14927f7778a246f90c5cec0c581be3e2658c2a8a3726c59db5be3f9R22

PR #47, opened 2021-05-30
Rescue exceptions raised when inspecting objects

### LINKED ISSUE #45: Exceptions inside frame inspection are fatal
I'm really not sure if this is a bug, but I presume it is. The example is contrived, but I was able to trigger this by putting `URI` in a strange state (for which I was debugging :joy:).

Target
```ruby
class Baz
  Error = Class.new(StandardError)

  def initialize
    raise Error
  end

  def to_s
    raise
  end

  def inspect
    to_s
  end
end

Baz.new
```

Command
```
exe/rdbg -e 'catch Baz::Error ;; c ;; up ;; i' -e c target.rb
```

I expect the frame information to be shown, but instead my original exception is raised again and the debugger exits.

I'm really not sure what the best solution is here. Rescue then display `rescued SomeException when inspecting`?  Or perhaps `debug` can have it's own object here that's used in it's place?

The patch below solves the issue.

```diff
diff --git a/lib/debug/thread_client.rb b/lib/debug/thread_client.rb
index 16cd2b7..808341b 100644
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -29,6 +29,8 @@ module DEBUGGER__
       else
         obj.pretty_inspect
       end
+    rescue => ex
+      colored_inspect("(rescued #{ex.inspect} during inspection)")
     end
 
     def colorize_cyan(str)
```

adam12:
Exceptions raised when inspecting frame variables cause the debugger to
exit with the original Exception.

Instead, note to the user that an exception was rescued and any possible
result of `#to_s` on the failing object.

Closes #45

image

### DIFF
--- a/lib/debug/color.rb
+++ b/lib/debug/color.rb
@@ -17,12 +17,12 @@ module DEBUGGER__
         obj.pretty_inspect
       end
     rescue => ex
-      err_msg = "(rescued #{ex.inspect} during inspection)"
+      err_msg = "#{ex.inspect} rescued during inspection"
       string_result = obj.to_s rescue nil
 
       # don't colorize the string here because it's not from user's application
       if string_result
-        "#{string_result} #{err_msg}"
+        %Q{"#{string_result}" from #to_s because #{err_msg}}
       else
         err_msg
       end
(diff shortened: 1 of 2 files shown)


st0012 (approved):
@adam12 thank you!

ko1:
Thank you.

PR #48, opened 2021-05-31
modify the way to get readable IO

ono-max:


lib/debug/thread_client.rb
```diff
@@ -103,6 +104,7 @@ def inspect
     end
 
     def puts str = ''
+      generate_info
```
ono-max, line 107: I'd like to discuss where this method should be called.

lib/debug/test_console.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    def ask prompt
+      setup_interrupt do
+        puts prompt
+        (gets || '').strip
+      end
+    end
+
+    def puts str = nil
+      @backtrace.push(str)
```
ko1, line 22: What is `backtrace`? `backlog`?

lib/debug/session.rb
```diff
@@ -185,7 +186,11 @@ def wait_command_loop tc
 
     def wait_command
       if @initial_commands.empty?
-        line = @ui.readline
+        line =  if ENV['RUBY_DEBUG_TEST_MODE']
+                  @ui.readline(@internal_info)
```
ko1, line 190: small tips: You don't need deep indent by writing assignments twice:

```ruby
if ENV['RUBY_DEBUG_TEST_MODE']
  line = @ui.readline(@internal_info)
...
```

lib/debug/session.rb
```diff
@@ -953,12 +958,20 @@ def self.require_location
...
+    UI_Console.prepend(TestUI_Console)
+  end
+
   # start methods
 
   def self.console **kw
     set_config(kw)
 
     require_relative 'console'
+    prepend_console_for_test
```
ko1, line 974: `prepend` is implementation (HOW TO).
It should be `prepare` or something goal (WHAT DOES IT).

BTW, only 4 lines, this method can be inlined here.

lib/debug/session.rb
ko1, line 190: Changing parameters of `@ui.readline` is nod good idea because `#readline` is defined for each UI classes.
Why don't you write a `@ui.puts info` first?

lib/debug/thread_client.rb
```diff
@@ -95,8 +96,12 @@ def << req
       @q_cmd << req
     end
 
+    def generate_info
+      @internal_info = { location: current_frame.location_str, line: current_frame.location.lineno }
+    end
+
     def event! ev, *args
-      @q_evt << [self, @output, ev, *args]
+      @q_evt << [self, @output, ev, @internal_info, *args]
```
ko1, line 104: why don't you call `generate_info` here?

ono-max: Thanks, I'll call it in `event!` method.

lib/debug/session.rb
ono-max: That's a good idea.

lib/debug/session.rb
ono-max: I will be inline them.

lib/debug/thread_client.rb
ono-max: https://github.com/ruby/debug/pull/48/files#diff-4208807b6db8555c61e42002b3e9c879003c5cfdb32d2b1a285a614b22fdf9d5R103

lib/debug/session.rb
ono-max: https://github.com/ruby/debug/pull/48/files#diff-c87adef8ace662a03249d206e6812162256bbcecabe18cd51861d1e1cd9cdfcaR189

lib/debug/session.rb
ono-max: https://github.com/ruby/debug/pull/48/files#diff-c87adef8ace662a03249d206e6812162256bbcecabe18cd51861d1e1cd9cdfcaR965-R968

lib/debug/test_console.rb
ono-max: I changed `backtrace` to `backlog`.

test/support/utils.rb
```diff
@@ -47,41 +48,44 @@ def create_pseudo_terminal(boot_options: "-r debug/run")
...
-        until quit
-          read.expect(/(.*)\n|\(rdbg\)/) do |sentence|
-            debug_print sentence[0]
-            if sentence[0] == '(rdbg)'
-              cmd = @queue.pop
-              if cmd.is_a?(Proc)
-                cmd.call(result)
+        lines = []
+        begin
+          Timeout.timeout(1) do
```
ko1, line 56: How to determine "1 second"?

ono-max: @ko1 
I think 1 second is enough time. What do you think about it?

ko1: Good idea to ask it :)

For me, 1 second is too short. If it runs complex tests, they will take more time. Also on the low-power computers, this  limitation should be more problematic. Tests is expected to finish sometime, so you need to check the "unexpected freeze". In many cases maybe 10 seconds are fine.

Introducing a way to change this timeout will help for slow computers (RUBY_DEBUG_TIMEOUT_SCALE for example), but now it is too much.

ono-max: Thanks, I will change it!

test/support/utils.rb
```diff
@@ -47,41 +47,44 @@ def create_pseudo_terminal(boot_options: "-r debug/run")
...
-          debug_print sentence[0]
+        rescue Errno::EIO => e
+          if @queue.empty?
+            # result of `gets` return this exception in some platform
+            # https://github.com/ruby/ruby/blob/master/ext/pty/pty.c#L729-L736
+          else
+            p e
+            pp lines
+          end
+        rescue Timeout::Error => e
```
ko1, line 83: On the timeout, the test should fail. Please try after merge it.

ko1: Also, please try the timeout on your computer by adding `sleep 11` or something similar.

ko1:
Congratulations!

PR #50, opened 2021-05-31
Support adding commands when using binding.bp

st0012:
This closes #43 

**Default (without commands)**

```
❯ exe/rdbg target.rb
[1, 10] in target.rb
=>    1| class Foo
      2|   def first_call
      3|     second_call(20)
      4|   end
      5|
      6|   def second_call(num)
      7|     third_call_with_block do |ten|
      8|       forth_call(num, ten)
      9|     end
     10|   end
=>#0    <main> at target.rb:1

(rdbg) c
[14, 23] in target.rb
     14|
     15|     yield(10)
     16|   end
     17|
     18|   def forth_call(num1, num2)
=>   19|     binding.bp()
     20|     num1 + num2
     21|   end
     22| end
     23|
=>#0    Foo#forth_call(num1=20, num2=10) at target.rb:19 #=> <DEBUGGER__::LineBreakpoint line bp /Use...
  #1    block {|ten=10|} in second_call at target.rb:8
  # and 4 frames (use `bt' command for all frames)

Stop by #0 line bp /Users/st0012/projects/debug/lib/debug/session.rb:1005 (return)

(rdbg)
```

**With commands**

```
❯ exe/rdbg target.rb
[1, 10] in target.rb
=>    1| class Foo
      2|   def first_call
      3|     second_call(20)
      4|   end
      5|
      6|   def second_call(num)
      7|     third_call_with_block do |ten|
      8|       forth_call(num, ten)
      9|     end
     10|   end
=>#0    <main> at target.rb:1

(rdbg) c
[14, 23] in target.rb
     14|
     15|     yield(10)
     16|   end
     17|
     18|   def forth_call(num1, num2)
=>   19|     binding.bp(commands: "bt ;; info")
     20|     num1 + num2
     21|   end
     22| end
     23|
=>#0    Foo#forth_call(num1=20, num2=10) at target.rb:19 #=> <DEBUGGER__::LineBreakpoint line bp /Use...
  #1    block {|ten=10|} in second_call at target.rb:8
  # and 4 frames (use `bt' command for all frames)

Stop by #0 line bp /Users/st0012/projects/debug/lib/debug/session.rb:1005 (return)
(rdbg:init) bt
=>#0    Foo#forth_call(num1=20, num2=10) at target.rb:19 #=> <DEBUGGER__::LineBreakpoint line bp /Use...
  #1    block {|ten=10|} in second_call at target.rb:8
[...]

PR #51, opened 2021-05-31
Add binding bp test & update readme

st0012:
I decide not to add these changes in https://github.com/ruby/debug/pull/50 in case it's rejected or needs more time for review. This could be merged right away.

### DIFF
--- a/test/support/utils.rb
+++ b/test/support/utils.rb
@@ -16,6 +16,12 @@ module DEBUGGER__
       })
     end
 
+    def assert_line_text(expected)
+      @queue.push(Proc.new { |result|
+        assert_match(expected, result)
+      })
+    end
+
     def debug_code(program, &block)
       @queue = Queue.new
       block.call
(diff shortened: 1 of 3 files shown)


test/support/utils.rb
```diff
@@ -16,6 +16,12 @@ def assert_line_num(expected)
       })
     end
 
+    def assert_line_text(expected)
+      @queue.push(Proc.new { |result|
+        assert_match(expected, result)
+      })
+    end
```
ono-max, line 23: To prevent this file from getting large, we need to create a class for assertion.
I will do that in new PR 👍

st0012: sounds good 👍 
I guess you mean `module` though? generally the assertion helpers are defined in assertion modules, which are included in the test classes (e.g. [irb's assertion module](https://github.com/ruby/irb/blob/9c19899fe5/test/lib/test/unit/assertions.rb)).

ono-max: Oh, that's right. Thank you!

PR #52, opened 2021-06-01
README: Build status Badge only about "master"

olleolleolle:
This change makes the badge go non-green only when the default branch ("master") has a failing CI. And the link now focuses the build on that branch.


ko1:
Thanks!

PR #53, opened 2021-06-01
Add test case for binding.bp(command:)

st0012:


test/debug/bp_test.rb
```diff
@@ -24,4 +24,31 @@ def test_breakpoint_fires_correctly
...
+
+      Foo.new.bar
+      RUBY
+    end
+
+    def test_breakpoint_execute_command_argument_correctly
+      debug_code(program) do
+        type 'continue'
+        assert_line_text(/Foo#baz/)
+        type 'quit'
```
ono-max, line 50: I'm sorry, I forgot reviewing this part in https://github.com/ruby/debug/pull/51. To finish debugger correctly, we should use `type y`
```suggestion
        type 'quit'
        type 'y'
```

PR #55, opened 2021-06-01
Fix MethodBreakpoint

st0012:
There are several errors fixed in this PR. The first triggers this:

```
/Users/st0012/projects/debug/lib/debug/source_repository.rb:42:in `read': no implicit conversion of nil into String (TypeError)
```

The 2nd and 3rd issue are explained in the commit message and comments.

### DIFF
--- a/lib/debug/breakpoint.rb
+++ b/lib/debug/breakpoint.rb
@@ -353,15 +353,19 @@ module DEBUGGER__
       rescue ArgumentError => e
         raise if retried
         retried = true
+        sig_method_name = @sig_method_name
 
         # maybe C method
         @klass.module_eval do
-          orig_name = @sig_method_name + '__orig__'
-          alias_method orig_name, @sig_method_name
-          define_method(@sig_method_name) do |*args|
+          orig_name = sig_method_name + '__orig__'
+          alias_method orig_name, sig_method_name
+          define_method(sig_method_name) do |*args|
             send(orig_name, *args)
           end
         end
+
+        # re-collect the method object after the above patch
+        search_method
         retry
       end
     rescue Exception
--- a/lib/debug/source_repository.rb
+++ b/lib/debug/source_repository.rb
@@ -53,7 +53,7 @@ module DEBUGGER__
         iseq.instance_variable_get(:@debugger_si)
       elsif @files.has_key?(path = iseq.absolute_path)
         @files[path]
-      else
+      elsif path
         add_path(path)
       end
     end
--- a/test/debug/breakpoint_test.rb
+++ b/test/debug/breakpoint_test.rb
@@ -6,9 +6,9 @@ module DEBUGGER__
   class MethodBreakpointTest < TestCase
     def program
       <<~RUBY
-      a = 1
-
-      a.abs
+      1| a = 1
+      2|
+      3| a.abs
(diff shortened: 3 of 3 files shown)


lib/debug/breakpoint.rb
```diff
@@ -348,15 +348,19 @@ def try_enable quiet: false
       rescue ArgumentError => e
         raise if retried
         retried = true
+        sig_method_name = @sig_method_name
```
st0012, line 351: Instance variable like `@sig_method_name` is `nil` inside `module_eval`. We need to assign it to a local for accessing it in `module_eval`.

lib/debug/breakpoint.rb
```diff
@@ -348,15 +348,19 @@ def try_enable quiet: false
...
-          define_method(@sig_method_name) do |*args|
+          orig_name = sig_method_name + '__orig__'
+          alias_method orig_name, sig_method_name
+          define_method(sig_method_name) do |*args|
             send(orig_name, *args)
           end
         end
+
+        # re-collect the method object after the above patch
+        search_method
```
st0012, line 363: We need to re-collect the method object after applying the patch for C methods. Otherwise it'd still trigger the same argument error.

st0012:
@ko1 this is the first time I touch breakpoints so I may have misunderstanding on them. can you point out the issue in comments and give me another chance to fix it? thanks 😄

lib/debug/source_repository.rb
```diff
@@ -49,11 +49,13 @@ def add_path path
     def get_si iseq
       return unless iseq
 
+      path = iseq.absolute_path
```
ko1, line 52: I prefer https://github.com/ruby/debug/pull/55/files#diff-f199a045ec16c0d96b45c5260edd5f1cc070f5e6d9d30a9ec6d989b9b4a59178L54 with `elsif path` instead of `else`.

ko1: and is it independent patch?

st0012: > with elsif path instead of else
sorry I don't get it. I'm using `elsif path`, do you want me to change back to `else`?

> and is it independent patch?
yes

st0012: initially I submitted this PR for this patch, but then the CI failed because of the other 2 issues. so I had to fix them at once.

ko1: I prefer:

```ruby
      elsif @files.has_key?(path = iseq.absolute_path)
        @files[path]
      elsif path
        add_path(path)
```

because it can reduce `iseq.absolute_path` call.

st0012: I see, pushed 👍

st0012:
@ko1 thank you ❤️

test/debug/breakpoint_test.rb
```diff
@@ -0,0 +1,32 @@
...
+          # it doesn't show any source before Ruby 3.0
+          assert_line_text(/<main>/)
+        end
+
+        type 'quit'
+        type 'y'
+      end
+    end
+  end
+end
```
ono-max, line 32: @st0012 
Could you tell me why you created new file instead of  adding to break_test.rb?

Issue #49, opened 2021-05-31
Improve the watch command


st0012:
Description copied from #41:

I found several things when playing with the `watch` command:

- It mainly uses `line` events to detect the change instead of the `return` events (I'm not sure why though). This causes the program to always stop a line after the change. So in tests we need to add an extra line after the changed line. Otherwise it doesn't stop.
- It always evaluates the expression from top-level scope. So it's not able to track an instance variable inside an object like `watch @name` and users need to use `watch obj.name` instead. I think we should improve this.

ko1:
> Otherwise it doesn't stop.

Could you show the example?

I agree it should stop also `return` events, but it will stops just another frames.
or another frame can't evaluate because of different bindings?

> I think we should improve this.

Maybe this request is, "I want to watch the change of an ivar of *THIS* object". So it needs to specify two parameters, "Object" and "Ivar name".

I'm planning to introduce `watch @ivar of obj`. Does it make sense?
Current `watch` command only support "global visible" variables.

Another topic of "watch" command is, should we keep the binding of the frame.
If we keep the binding of "watch" command, we can evaluate "@ivar" of binding, and you don't need to specify "of self". It seems convenient, but it seems too much feature. No conclusion.

st0012:
> Could you show the example?

if the change happens at the last line if the program, it doesn't stop there. that's why the tests need an extra line at the end

https://github.com/ruby/debug/blob/f44de3a06ec57d9d9312b40c98eb23532e0e1202/test/debug/watch_test.rb#L8-L14

> I'm planning to introduce watch @ivar of obj. Does it make sense?

yeah this feature sounds good 👍 but syntax-wise I'd go for `watch obj @ivar`.

> Maybe this request is, "I want to watch the change of an ivar of THIS object". So it needs to specify two parameters, "Object" and "Ivar name".

from a user's perspective, if the breakpoint is already inside an object's method, I'd expect `watch @var` to just work. for example:

```ruby
class Foo
  def initialize
    @var = 0
    binding.bp(command: "watch @var")
  end

  def bar
    @var = 10
  end
end

f = Foo.new
f.bar
```

I know this could be hard to implement and may not a be priority right now. but I believe it provides a better user experience.

ko1:
or only allow for ivar (or gvar) like `watch @ivar` ?
It's simple. Nobody can misuse `local_var`.

st0012:
so only:
- `watch @ivar`
- `watch $gvar`

but no `watch local`?
[...]

PR #58, opened 2021-06-02
Replace WatchExprBreakpoint with WatchIVarBreakpoint

st0012:
Based on the discussion in https://github.com/ruby/debug/issues/49, having `WatchExprBreakpoint` for watch all expression changes costs too much on performance but provide limited benefit to the users. But because watching instance variables for state changes is still valuable for debugging, we're going to support this particular case.

So in this PR, I made these changes:
- Replace `WatchExprBreakpoint` with `WatchIVarBreakpoint`.
- `watch expr` is no longer supported.
- `watch @ivar` is more accurate and cheaper.

---

## About WatchIVarBreakpoint

Currently, when watching instance variables, the breakpoint isn't aware of the instance/class it's tracking on. So it could incorrectly track instance variable change (with the same name) in a different class. 

For example, if you debug this script:

```ruby
class Student
  attr_accessor :name

  def initialize(name)
    @name = name
    binding.bp(command: "watch @name ;; c") # should only watch Student's @name change
  end
end

class Teacher
  attr_accessor :name

  def initialize(name)
    @name = name
  end
end

s1 = Student.new("John")
t1 = Teacher.new("Jane") # this shouldn't trigger a stop
s1.name = "Josh"
"foo"
```

You'll find that it stops at `Teacher`'s `@name` assignments as well, even though the script should only monitor `Student`'s `@name`.

```
✦ ❯ exe/rdbg target.rb
[1, 10] in target.rb
=>    1| class Student
      2|   attr_accessor :name
      3|
      4|   def initialize(name)
      5|     @name = name
      6|     binding.bp(command: "watch @name ;; c")
      7|   end
      8| end
      9|
     10| class Teacher
=>#0    <main> at target.rb:1

(rdbg) c
[1, 10] in target.rb
      1| class Student
      2|   attr_accessor :name
      3|
      4|   def initialize(name)
      5|     @name = name
=>    6|     binding.bp(command: "watch @name ;; c")
      7|   end
      8| end
      9|
     10| class Teacher
=>#0    Student#initialize(name="John") at target.rb:6
  #1    [C] Class#new at target.rb:18
  #2    <main> at target.rb:18
(rdbg:init) watch @name
@name = John
#0 watch bp: @name = John
(rdbg:init) c
[14, 22] in target.rb
     14|     @name = name
     15|   end
     16| end
     17|
     18| s1 = Student.new("John")
=>   19| t1 = Teacher.new("Jane")
     20| s1.name = "Josh"
     21| "foo"
     22|
=>#0    <main> at target.rb:19

Stop by #0 watch bp: @name = John ->

(rdbg) c
[10, 19] in target.rb
     10| class Teacher
     11|   attr_accessor :name
     12|
     13|   def initialize(name)
     14|     @name = name
=>   15|   end
     16| end
     17|
     18| s1 = Student.new("John")
     19| t1 = Teacher.new("Jane")
=>#0    Teacher#initialize(name="Jane") at target.rb:15 #=> "Jane"
  #1    [C] Class#new at target.rb:19
  #2    <main> at target.rb:19

Stop by #0 watch bp: @name =  -> Jane

(rdbg) c
[15, 22] in target.rb
     15|   end
     16| end
     17|
     18| s1 = Student.new("John")
     19| t1 = Teacher.new("Jane")
=>   20| s1.name = "Josh"
     21| "foo"
     22|
=>#0    <main> at target.rb:20

Stop by #0 watch bp: @name = Jane ->

(rdbg) c
```

### After 

```
✦ ❯ exe/rdbg target.rb
[1, 10] in target.rb
=>    1| class Student
      2|   attr_accessor :name
      3|
      4|   def initialize(name)
      5|     @name = name
      6|     binding.bp(command: "watch @name ;; c")
      7|   end
      8| end
      9|
     10| class Teacher
=>#0    <main> at target.rb:1

(rdbg) c
[1, 10] in target.rb
      1| class Student
      2|   attr_accessor :name
      3|
      4|   def initialize(name)
      5|     @name = name
=>    6|     binding.bp(command: "watch @name ;; c")
      7|   end
      8| end
      9|
     10| class Teacher
=>#0    Student#initialize(name="John") at target.rb:6
  #1    [C] Class#new at target.rb:18
  #2    <main> at target.rb:18
(rdbg:init) watch @name
#<Student:0x00007fc3c68cc8c8> @name = John
#0 watch bp: #<Student:0x00007fc3c68cc8c8> @name = John
(rdbg:init) c
[16, 22] in target.rb
     16| end
     17|
     18| s1 = Student.new("John")
     19| t1 = Teacher.new("Jane")
     20| s1.name = "Josh"
=>   21| "foo"
     22|
=>#0    <main> at target.rb:21

Stop by #0 watch bp: #<Student:0x00007fc3c68cc8c8> @name = John -> Josh

(rdbg) c
```

Another benefit of the change is that we can show what instance's ivar is changed:

```
#<Student:0x00007fc3c68cc8c8> @name = John
```

(I plan to work on styling it in another PR)

### DIFF
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -485,9 +485,6 @@ module DEBUGGER__
                   end
                 puts "#{object} #{eval_src} = #{result}"
                 result = WatchIVarBreakpoint.new(eval_src, object)
-              else
-                puts "#{eval_src} = #{result}"
-                result = WatchExprBreakpoint.new(eval_src, result)
               end
 
               result_type = :watch
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -484,7 +484,7 @@ module DEBUGGER__
                     current_frame.self
                   end
                 puts "#{object} #{eval_src} = #{result}"
-                result = WatchIVarBreakpoint.new(eval_src, object)
+                result = WatchIVarBreakpoint.new(eval_src, object, result)
               end
 
               result_type = :watch
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -476,8 +476,20 @@ module DEBUGGER__
             result = failed_results
           when :watch
             if @success_last_eval
-              puts "#{eval_src} = #{result}"
-              result = WatchExprBreakpoint.new(eval_src, result)
+              if eval_src.match?(/@\w+/)
+                object =
+                  if b = current_frame.binding
+                    b.receiver
+                  else
+                    current_frame.self
+                  end
+                puts "#{object} #{eval_src} = #{result}"
+                result = WatchIVarBreakpoint.new(eval_src, object)
+              else
+                puts "#{eval_src} = #{result}"
+                result = WatchExprBreakpoint.new(eval_src, result)
+              end
+
               result_type = :watch
             else
               result = nil
--- a/test/debug/watch_test.rb
+++ b/test/debug/watch_test.rb
@@ -49,6 +49,45 @@ module DEBUGGER__
     end
   end
 
+  class InstanceIVarWatchingTest < TestCase
+    def program
+      <<~RUBY
+        class Student
+          attr_accessor :name
+
+          def initialize(name)
+            @name = name
+            binding.bp(command: "watch @name")
+          end
+        end
+
+        class Teacher
+          attr_accessor :name
+
+          def initialize(name)
+            @name = name
+          end
+        end
+
+        s1 = Student.new("John")
+        t1 = Teacher.new("Jane") # it shouldn't stop for this
+        s1.name = "Josh"
+        "foo"
+      RUBY
+    end
+
+    def test_debugger_only_stops_when_the_ivar_of_instance_changes
+      debug_code(program) do
+        type 'continue'
+        assert_line_text('Student#initialize(name="John")')
+        type 'continue'
+        assert_line_num(21) # stops when assigned to "Josh"
+        type 'quit'
+        type 'y'
+      end
+    end
+  end
+
   class MethodWatchingTest < TestCase
     def program
       <<~RUBY
--- a/test/debug/watch_test.rb
+++ b/test/debug/watch_test.rb
@@ -29,27 +29,27 @@ module DEBUGGER__
   class ObjectInstanceVariableWatchingTest < TestCase
     def program
       <<~RUBY
-        class Student
-          attr_accessor :name
-
-          def initialize(name)
-            @name = name
-            binding.bp(command: "watch @name")
-          end
-        end
-
-        class Teacher
-          attr_accessor :name
-
-          def initialize(name)
-            @name = name
-          end
-        end
-
-        s1 = Student.new("John")
-        t1 = Teacher.new("Jane") # it shouldn't stop for this
-        s1.name = "Josh"
-        "foo"
+         1| class Student
+         2|   attr_accessor :name
+         3|
+         4|   def initialize(name)
+         5|     @name = name
+         6|     binding.bp(command: "watch @name")
+         7|   end
+         8| end
+         9|
+        10| class Teacher
+        11|   attr_accessor :name
+        12|
+        13|   def initialize(name)
+        14|     @name = name
+        15|   end
+        16| end
+        17|
+        18| s1 = Student.new("John")
+        19| t1 = Teacher.new("Jane") # it shouldn't stop for this
+        20| s1.name = "Josh"
+        21| "foo"
       RUBY
     end
 
--- a/test/debug/watch_test.rb
+++ b/test/debug/watch_test.rb
@@ -3,30 +3,7 @@
 require_relative '../support/test_case'
 
 module DEBUGGER__
-  class LocalVariableWatchingTest < TestCase
-    def program
-      <<~RUBY
-        1| a = 1
-        2| b = 2
-        3| c = 3
-        4| a = 2
-        5| foo = "foo" # stops here
-      RUBY
-    end
-
-    def test_debugger_stops_when_the_expression_changes
-      debug_code(program) do
-        type 'step'
-        type 'watch a'
-        type 'continue'
-        assert_line_num(5) # stops at the next line
-        type 'quit'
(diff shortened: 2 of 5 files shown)


test/debug/watch_test.rb
```diff
@@ -49,6 +49,45 @@ def test_debugger_stops_when_the_expression_changes
...
+
+          def initialize(name)
+            @name = name
+          end
+        end
+
+        s1 = Student.new("John")
+        t1 = Teacher.new("Jane") # it shouldn't stop for this
+        s1.name = "Josh"
+        "foo"
```
ono-max, line 75: Can you add line number to this program?

st0012: I'd do it later because it's quite painful to modify test content with all the test numbers 🙂 
can have some kind of rake task that can print out test script with line numbers? so it'd be like:

1. write tests  without line numbers
2. run the rake task (or just run `rake test`)
3. it fails tests that have no line numbers & print out the updated version of the scripts
4. developers copy & paste those test scripts

ko1:
This proposal is strongly connected with "how to design `watch` command".  https://github.com/ruby/debug/issues/49

So wait for merging.

st0012:
@ko1 btw this breakpoint is immune from the issue I mentioned in https://github.com/ruby/debug/pull/61#discussion_r645261653

test/debug/watch_test.rb
ono-max: > I'd do it later because it's quite painful to modify test content with all the test numbers 🙂
Ok, I got it.

> can have some kind of rake task that can print out test script with line numbers? so it'd be like:
Yeah, I agree with your opinion 😄

st0012: @ono-max I've been working on the tool set I mentioned (pretty naive though). I will open a PR shortly.

lib/debug/thread_client.rb
```diff
@@ -470,8 +470,17 @@ def wait_next_action
...
-              result = WatchExprBreakpoint.new(eval_src, result)
+              if eval_src.match?(/@\w+/)
+                object =
+                  if b = current_frame.binding
+                    b.receiver
+                  else
+                    current_frame.self
+                  end
+                puts "#{object} #{eval_src} = #{result}"
+                result = WatchIVarBreakpoint.new(eval_src, object)
```
ko1, line 481: Don't you need a `result`?

target.rb
```diff
@@ -0,0 +1,22 @@
+class Student
```
ko1, line 1: I also use this filename :)

lib/debug/thread_client.rb
st0012: I thought the result should be the breakpoint's `key`? which will be handled later automatically.
assigning anything else than a bp results in `undefined method `key' ` error.

target.rb
st0012: oops I'll remove it.

test/debug/watch_test.rb
st0012: Updated

lib/debug/thread_client.rb
ko1: No, `@current` in `WIVarBP#initialzie`.
BtW, on 480 `result` is displayed but `result` is not evaluated.

st0012: > No, @current in WIVarBP#initialzie.

If the eval_type is `watch`, `Session` always calls `bp.key` here:

https://github.com/ruby/debug/blob/3f3757328e0b1da3e43a960e8d32533b975ebeee/lib/debug/session.rb#L105-L108

> BtW, on 480 result is displayed but result is not evaluated.

It's evaluated at the beginning of upper level case statement:

https://github.com/ruby/debug/blob/3f3757328e0b1da3e43a960e8d32533b975ebeee/lib/debug/thread_client.rb#L440-L447

st0012: of course, we can change how `watch` bp should be presented. but because that needs to touch the eval value processing flow, I think it's better to do it in another PR.

ko1: Sorry `result` is evaluated.

I couldn't catch why you are talking about `@key`. or `#key`. `@current` is not related `key` method.

my idea is:

```ruby
def initialize ivar, object, current
  @current = current
  ...
```

and call `new` with `result`.

st0012: sorry that I thought you meant re-assigning another value to `result`.

given that the `WatchIVarBreakpoint` already has the `object` and `@ivar`, I think it's not necessary to take the value in the constructor. I decided to use `@current = object.instance_variable_get(@ivar)` to fetch it manually.

https://github.com/ruby/debug/pull/58/files#diff-f5007904d2ba2c3f10b7abf495c914f613c8805950216a5c4544c08f1946578eR276

ko1: I know it is evaluated manually, and my question is why it is better.

st0012: since `@current = object.instance_variable_get(@ivar)` is a cheap and non-repeating operating, I think it's worth doing so to reduce a required constructor input.
another argument could be that it avoids the potential issue the having incorrect current value passed in.

I don't think it's a "better" option than what you proposed, but I don't think it's a wrong decision either.

ko1: > I think it's worth doing so to reduce a required constructor input.

Could you explain more?
It is first time to ask this kind of argument.

> another argument could be that it avoids the potential issue the having incorrect current value passed in.

If the `result` in thread_client.rb is different from ivar value, it should be a bug (because it was displayed), and they should be fixed.

st0012: ok I'm not going to insist on the current solution, will change it later.

st0012: I've updated it.

ko1: Thank you. For DAP support I'm working now and Ractor support in future, I need to change these code and I needed to confirm the reason *why*.

st0012: I see, thanks for the explanation 👍

PR #60, opened 2021-06-03
Add irb as dependency for 2.x Ruby users

st0012:
Ruby 2.x's irb doesn't have the `irb/color_printer` that the debugger relies on. So users of those versions need to manually add `irb` to their gemfiles. I think we can save them the work by adding `irb` as this project's dependency.

CI didn't fail because we have `irb` in our gemfile as a development dependency.

### DIFF
--- a/debug.gemspec
+++ b/debug.gemspec
@@ -24,4 +24,6 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
   spec.extensions    = ['ext/debug/extconf.rb']
+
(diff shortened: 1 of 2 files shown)


debug.gemspec
```diff
@@ -24,4 +24,6 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
   spec.extensions    = ['ext/debug/extconf.rb']
+
+  spec.add_dependency "irb", "~> 1.3" # for its color_printer class, which was added after 1.3
```
ko1, line 28: Maybe `~> 1.3` is not needed.

Issue #59, opened 2021-06-03
Label breakpoint output


st0012:
I've been playing breakpoints for a while. And one thing that bothers me is that it's not always easy to find the breakpoint messages. Because they are usually surrounded by debugger's prompts or other output:

**Current BP messages**

<img width="398" alt="before bp label" src="https://user-images.githubusercontent.com/5079556/120581826-0e405880-c45e-11eb-9ffc-6d85ec1d9f11.png">

But colorizing them like backtrace or frame info may just cause more confusion. 

So instead, I want to use reverse coloring effect to add something called "breakpoint label", which looks like this:

**Proposed messages**

<img width="390" alt="after bp label" src="https://user-images.githubusercontent.com/5079556/120581843-15676680-c45e-11eb-891d-0d6ad618f1ad.png">

It also looks great with a bright background.

<img width="287" alt="截圖 2021-06-03 上午11 29 08" src="https://user-images.githubusercontent.com/5079556/120582341-eac9dd80-c45e-11eb-93cb-ae649e5ffe74.png">

### Notes

- This will be applied to the content of `Breakpoint#to_s`.
- I also plan to colorize some breakpoint (like watch breakpoint)'s value. The color scheme will be the same as backtrace.
  - It's also possible to use colored inspect on changed values, but I'm not sure if that'd be an overkill or be too noisy? @ko1 wdyt?

ko1:
Let's try!

BTW `BP` is of course "Break Point". Is it readable?

st0012:
I use `BP` because the current output uses `bp`, but I don't think that's clear to users. 

To me, `BP` implies there are 2 words "Break Point", which is not how we usually spell `breakpoint`. However just calling it `B` is more confusing.

So if it's ok to you, I prefer using `Breakpoint`. The lines aren't long in general, so I don't think the extra characters would could issues.

st0012:
I think I'm wrong here. `Breakpoint` doesn't make the line too long but is verbose.

<img width="479" alt="截圖 2021-06-03 下午5 45 51" src="https://user-images.githubusercontent.com/5079556/120624713-8bd28b80-c493-11eb-9760-b2fdbf624e7a.png">

I'll use `BP` for now and change it later if others have different opinions.

ko1:
How about `#0 Breakpoint: <color>Watch</color>` ?

st0012:
> How about `#0 Breakpoint: <color>Watch</color>` ?

Result:

[...]

PR #61, opened 2021-06-03
Label breakpoint messages

st0012:
Closes https://github.com/ruby/debug/issues/59

Screenshot:
截圖 2021-06-03 下午6 08 18

### DIFF
--- a/lib/debug/breakpoint.rb
+++ b/lib/debug/breakpoint.rb
@@ -1,3 +1,4 @@
+require "irb/color"
 module DEBUGGER__
   class Breakpoint
     attr_reader :key
@@ -55,9 +56,24 @@ module DEBUGGER__
         ""
       end
     end
+
+    def self.generate_label(name)
+      colorize(" BP - #{name} ", [:YELLOW, :BOLD, :REVERSE])
+    end
+
+    def self.colorize str, color
+      if CONFIG[:use_colorize]
+        IRB::Color.colorize str, color
+      else
+        str
+      end
+    end
   end
 
   class LineBreakpoint < Breakpoint
+    LABEL = generate_label("Line")
+    PENDING_LABEL = generate_label("Line (pending)")
+
     attr_reader :path, :line, :iseq
 
     def initialize path, line, cond: nil, oneshot: false, hook_call: true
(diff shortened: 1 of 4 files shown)


lib/debug/color.rb
```diff
@@ -0,0 +1,29 @@
...
+      else
+        obj.pretty_inspect
+      end
+    end
+
+    def colorize_cyan(str)
+      colorize(str, [:CYAN, :BOLD])
+    end
+
+    def colorize_blue(str)
```
st0012, line 25: not directly related, but I plan to name coloring methods after the type of data they're used on. for example, this could be `colorize_sig`.

lib/debug/breakpoint.rb
```diff
@@ -283,14 +302,17 @@ def setup
 
     def to_s
       if defined? @prev
-        "watch bp: #{@expr} = #{@prev} -> #{@current}"
+        "#{LABEL} #{@expr} = #{@prev} -> #{@current}"
```
st0012, line 305: btw I found that the debugger enters an infinite loop (deadlock) when I call any color helper inside a breakpoint instance. it's part of the reason I store the label in a constant.

st0012: @ko1 do you know the possible cause of this issue? it doesn't matter what I colorize, `IRB::Color.colorize` seems to trigger a stop when called here.

截圖 2021-06-03 下午10 30 06

ko1: Not sure why...
Using `rescue Exception` on this method to check the exception can help you.

st0012: The cause is not the coloring helper. Calling any non-internal or non-debugger method like `pp` (I'll just call it "the outsider method" can also cause this issue. It happens like this:

1. A breakpoint is triggered by a state change (`@name = nil => John`)
2. The session would use `show_bps` to show breakpoints.
3. When `WatchExprBreakpoint#to_s` is called, it then calls the outsider method.
4. The outsider method will trigger `WatchExprBreakpoint`'s tracepoint call and then `watch_eval`.  
5. It'll then spot another state change (`@name = John => nil`) because `WatchExprBreakpoint` is class agnostic.
6. This triggers another breakpoint suspend call and back to the 2nd step. The infinite loops begins.

I guess this just provided another reason to drop the expression breakpoint.

ko1: I see. Maybe it should be fixed, but now we can drop it for a while.

PR #63, opened 2021-06-04
Require & generate line numbers for test scripts

st0012:
Test programs without line numbers will fail with a line-numbered version. So developers only needs to copy & paste them.

Example:

```
Error: test_breakpoint_execute_command_argument_correctly(DEBUGGER__::BindingBPWithCommandTest):
  RuntimeError: line numbers are required in test script. please update the script with:

     1| class Foo
     2|   def bar
     3|     binding.bp(command: "continue")
     4|     baz
     5|   end
     6|
     7|   def baz
     8|     binding.bp
     9|   end
    10| end
    11|
    12| Foo.new.bar
```

After this PR, the workflow would be:

1. Write plain Ruby script.
2. Run tests (fail for test/program logic).
3. Fix the test/debugger.
4. Run tests (fail for missing line numbers).
5. Fix the scripts by copy & paste.

### DIFF
--- a/test/debug/bp_test.rb
+++ b/test/debug/bp_test.rb
@@ -6,13 +6,13 @@ module DEBUGGER__
   class BindingBPTest < TestCase
     def program
       <<~RUBY
-      class Foo
-        def bar
-          binding.bp
-        end
-      end
-
-      Foo.new.bar
+     1| class Foo
+     2|   def bar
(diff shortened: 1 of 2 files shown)


st0012:
@ono-max this is the tool I mentioned 🙂

ko1:
Great.

PR #67, opened 2021-06-04
Add manual session start test

st0012:


test/support/utils.rb
```diff
@@ -42,12 +42,14 @@ def debug_print msg
 
     RUBY = RbConfig.ruby
 
-    def create_pseudo_terminal
+    def create_pseudo_terminal(start_with_session: true)
```
ko1, line 45: or `boot_option: '-r require debug/run'`

st0012: sure, I guess this will help testing other modes as well 👍

ko1:
Thanks.

PR #68, opened 2021-06-06
Reorganize control flow tests

st0012:
I think testing different control flow commands against the same script can help developers understand how they work. But maintaining the same program script across test files could be a hassle and it's not necessary at this stage. So in the PR I put all control flow command tests into the same file and rewrote some of them.


ono-max:
>  But maintaining the same program script across test files could be a hassle and it's not necessary at this stage
I agree with your opinion. But I would like to discuss more about how to resolve this problem. In my opinion, we should create a class to organize the script. It's because we don't have to put all control flow command test into the same file.

st0012:
Most of the control flow tests can and should share the same scripts. It has these benefits:

- It's easy to read and compare how the behave. 
- It saves space.

In contrary, if we share test program in other ways (say in a class), readers also need to open those files to understand the whole test case, which is inconvenient.

We can, and should, separate some of the test cases when the file grows bigger or when we have more command-specific edge cases to cover. But right now it's not necessary to keep them separated. 

For example, after merging those tests, I can add tests for the `finish` command with 2 methods, instead of adding a new file and new test classes. https://github.com/ruby/debug/pull/68/commits/ae83aaac7470bc25f7cc8c46e1055eb9810c542d

test/debug/control_flow_commands_test.rb
```diff
@@ -0,0 +1,184 @@
...
+        type 'b 11'
+        type 'c'
+        assert_line_num 11
+        type 's'
+        assert_line_num 3
+        type 'fin'
+        assert_line_num 12
+        type 's'
+        assert_line_num 7
+        type 'fin'
```
st0012, line 88: @ko1 not directly related, but I found it confusing to have `fin` instead of `f` as `finish`'s shortcut. maybe we can change it to make it consistent with other commands?

st0012: nvm, I found that it's been taken by `frame` 🤦

test/debug/control_flow_commands_test.rb
```diff
@@ -0,0 +1,184 @@
...
+        type 'b 11'
+        type 'c'
+        assert_line_num 11
+        type 'b 13'
+        type 'c'
+        assert_line_num 13
+        type 'quit'
+        type 'y'
+      end
+    end
```
ono-max, line 75: BTW, it's better to write the breakpoint test in `break_test.rb`.

st0012: it's not to test breakpoints. current test framework doesn't support exiting without `quit`. so it has to stop at another breakpoint for quitting. I'd be happy to avoid the second breakpoint if using `continue` to leave the terminal is an option 🙂

st0012: actually, can you open a PR and support that? it'd be much appreciated 🙏

ono-max: > actually, can you open a PR and support that? it'd be much appreciated 🙏


You mean, this one?
https://github.com/ruby/debug/pull/48

ono-max:
I plan to write test codes for `quit`, `finish`, `continue` such as byebug.(https://github.com/deivid-rodriguez/byebug/blob/master/test/commands/finish_test.rb will make it easier for you to make sense code amount.)

 If I wrote them, it would be huge, wouldn't it?
I'd like to hear your opinion on it.

test/debug/control_flow_commands_test.rb
st0012: that looks good, thanks 👍

st0012:
@ono-max I won't call it huge, why? it's a ~180-line file with 3 program script and 12 test cases that cover 3 commands. this is like what I want to do here.

ono-max:
All right. Let's give a shot

> btw I think our current commands format is easier to read than byebug's

Yeah, this is one of my favorite part lol

test/debug/control_flow_commands_test.rb
ko1: YES.

ko1:
@ono-max can I merge it?

ono-max:
@ko1 
Sure 👍

ko1:
Thanks.

PR #71, opened 2021-06-08
[Fix bug] modify assert_line_text method to be able to make correct assertions

ono-max:


test/support/utils.rb
```diff
@@ -18,11 +18,14 @@ def assert_line_num(expected)
 
     def assert_line_text(expected)
       @queue.push(Proc.new {
-        assert_block do
-          @internal_info['backlog'].each do |line|
-            line.include? expected
+        result = false
+        @internal_info['backlog'].each do |line|
+          if line.include? expected
```
st0012, line 23: can you use `match?` instead of `include?` so it can also take regexp as expectation?

ono-max: https://github.com/ruby/debug/pull/71/files#diff-ee14bf65f14927f7778a246f90c5cec0c581be3e2658c2a8a3726c59db5be3f9R23

st0012 (approved):
thanks 👍

test/support/utils.rb
```diff
@@ -18,11 +18,14 @@ def assert_line_num(expected)
...
-          @internal_info['backlog'].each do |line|
-            line.include? expected
+        result = false
+        @internal_info['backlog'].each do |line|
+          if line.match? expected
+            result = true
+            break
           end
         end
+        assert_true(result)
```
ko1, line 28: It should show failure message. When the test fails, there is no hints and it is difficult to debug.

ono-max: https://github.com/ruby/debug/pull/71/files#diff-ee14bf65f14927f7778a246f90c5cec0c581be3e2658c2a8a3726c59db5be3f9R21

ono-max: Thanks, I will fix it.

test/support/utils.rb
```diff
@@ -18,11 +18,15 @@ def assert_line_num(expected)
 
     def assert_line_text(expected)
       @queue.push(Proc.new {
-        assert_block do
-          @internal_info['backlog'].each do |line|
-            line.include? expected
+        msg = "Expected to include #{expected}}"
```
st0012, line 21: it still doesn't provide much information:

```
  Expected to include => 20}.
  <true> expected but was
  <false>
```

how about this?

```ruby
assert_match(expected, @internal_info['backlog'].join)
```

```
  </=>\ 20/> was expected to be =~
  <"=>#0\tFoo#first_call at /var/folders/yg/hnbymwxd5pn7v94_clc59y6r0000gn/T/debugger20210608-65153-b857w.rb:4 #=> 30\n" +
  "  #1\t<main> at /var/folders/yg/hnbymwxd5pn7v94_clc59y6r0000gn/T/debugger20210608-65153-b857w.rb:23\n">.
```

ono-max: That's a good idea! Thanks 👍

ono-max: https://github.com/ruby/debug/pull/71/files#diff-ee14bf65f14927f7778a246f90c5cec0c581be3e2658c2a8a3726c59db5be3f9R21

st0012:
btw it'd be great if we can merge this sooner than later because the current matcher doesn't take regexp object and never fails when passing string 😬

st0012:
@ono-max @ko1 thanks 😄

PR #73, opened 2021-06-08
improve test framework in backlog

ono-max:
- rename 'lines' to 'backlog'
- use "last_backlog" instead of "internal_info[:backlog]"
- define create_message method to output backlog when assertion failed

### DIFF
--- a/test/support/utils.rb
+++ b/test/support/utils.rb
@@ -12,20 +12,18 @@ module DEBUGGER__
 
     def assert_line_num(expected)
       @queue.push(Proc.new {
-        assert_equal(expected, @internal_info['line'])
+        assert_equal(expected, @internal_info['line'], create_message("Expected line number to be #{expected}, but was #{@internal_info['line']}"))
       })
     end
 
     def assert_line_text(expected)
       @queue.push(Proc.new {
-        assert_match(expected, @internal_info['backlog'].join)
+        assert_match(expected, @last_backlog[2..].join, create_message("Expected to include #{expected}"))
       })
     end
 
-    def assert_fail_with_log msg, lines
-      assert false,  "#{msg}\n" +
-                     "[DEBUG SESSION LOG]\n" +
-                     lines.map{|l| "> #{l}"}.join
+    def create_message fail_msg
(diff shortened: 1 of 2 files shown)


test/support/utils.rb
```diff
@@ -54,37 +52,41 @@ def create_pseudo_terminal(boot_options: "-r debug/run")
       timeout_sec = (ENV['RUBY_DEBUG_TIMEOUT_SEC'] || 10).to_i
 
       PTY.spawn("#{RUBY} #{boot_options} #{temp_file_path}") do |read, write, pid|
-        lines = []
+        backlog = []
```
ko1, line 55: why don't you make `@backlog` instead of `backlog` local variable?

ono-max: In my opinion, local variable is better than instance variable to keep the scope of the variable small.
What do you think about this?

ono-max: @ko1 
In my opinion, local variable is better than instance variable to keep the scope of the variable small.
What do you think about this?

ko1: In this case, `@last_backlog` and `@internal_info` are already used and I don't have any good reason why only `@backlog` is passed as an argument.

ko1: "All in arguments" is another reasonable option for me.

ko1: It doesn't make sense to me. The visibility control is not related to frequency. 

* `backlog` is only visible where the value is passed. Well controlled.
* `@internal_info` and `@last_backlog` are visible from any test class methods.

Visibility of these values should be same. At least I feel it should be same because there is no reason to change the visibility.

ko1: In general, well-controlled visibility is good idea. So passing `backlog` as an argument is nice.
However, `@internal_info` and `@last_backlog` are already exposed, and maybe nobody modify them from test methods (in other words, we can believe test writers). In this background, there is no reason to prohibit exposing `backlog` and it reduces the code size.

ono-max: I fully persuaded myself of your opinion. Thanks!

PR #74, opened 2021-06-08
Add tests for the info command

st0012:


test/debug/info_test.rb
```diff
@@ -0,0 +1,67 @@
...
+        type 'q!'
+      end
+    end
+
+    def test_prints_the_other_thread
+      debug_code(program) do
+        type 'b 6'
+        type 'c'
+        type 'info threads'
+        assert_line_text(/#1 \(sleep\)@.*:2 sleep_forever/)
```
st0012, line 62: @ko1 this test case only fails in the Ruby 2.7 test suite and it's not reproduceable to me.

```
--> #0 (sleep)@/tmp/debugger20210608-1615-18rjf3w.rb:6:in `<main>'
The following threads are not managed yet by the debugger:     #<Thread:0x000056409a6dbc40 /tmp/debugger20210608-1615-18rjf3w.rb:2 run>"
```

I'm not sure how a thread can escape the `thread_begin` tp event. or maybe it's registered incorrectly...? 
anyway, I think I'm not able to diagnose the issue, can you give it a look? thanks

ko1: Thank you. Tests help us.

ko1: Could you try on master?
I didn't check it details, but it can be fixed (and it can be not fixed.

st0012: I have rebased it on master but it still fails on 2.7 😕

ko1: Thank you I'll check it.

st0012: @ko1 (part of) the cause seems to be a race condition between the test thread's (the one in `foo`) registration  (`#thread_client_create`) and the `info threads` command. 

Currently, it's fixed by postponing the `info` call a bit.

PR #75, opened 2021-06-09
Add a helper to start debug session with the test program

st0012:
When writing tests, sometimes we may test the script manually. But currently this could take a few steps:

1. Copy the test program into a file, say `target.rb`
2. Remove all the line numbers.
3. Run `exe/rdbg -e target.rb` from the terminal.

And whenever we change the script or switched to another test case, we need to go through these steps again. 

So in the PR, I introduced a new helper called `_start_manual_debug_session(program)`. You can place it in any test case:

```ruby

def program
  # ....
end

def test_something_about_debugger
  _start_manual_debug_session(program) # just add this line
  debug_code(program) do
    # your original test code
  end
end
```

And then run the test case as you usually do, it'll enter a debug session with your test script 😄 

截圖 2021-06-10 下午5 11 51

(btw, the remote debugging session is pretty neat. I love it 😍 )

### DIFF
--- a/test/support/utils.rb
+++ b/test/support/utils.rb
@@ -95,11 +95,19 @@ module DEBUGGER__
     def manual_debug_code(program)
       print("[Starting a Debug Session with @#{caller.first}]\n")
       write_temp_file(strip_line_num(program))
-      inject_lib_to_load_path
-      pid = spawn("#{RUBY} -r debug/open #{temp_file_path}")
-      sleep(0.5) # wait for the debug server to start
+
       require_relative "../../lib/debug/client"
-      connect
+
+      socket_path = DEBUGGER__.create_unix_domain_socket_name
+      inject_lib_to_load_path
+      ENV["RUBY_DEBUG_SOCK_PATH"] = socket_path
+      pid = spawn("#{RUBY} -r debug/open #{temp_file_path}")
+
+      while !File.exist?(socket_path)
+        sleep 0.1
+      end
+
+      DEBUGGER__::Client.new([socket_path]).connect
     ensure
       Process.kill('QUIT', pid)
     end
--- a/test/support/utils.rb
+++ b/test/support/utils.rb
@@ -43,9 +43,7 @@ module DEBUGGER__
     RUBY = RbConfig.ruby
 
     def create_pseudo_terminal(boot_options: "-r debug/run")
-      lib = "#{__dir__}/../../lib"
-
-      ENV['RUBYOPT'] = "-I #{lib}"
+      inject_lib_to_load_path
       ENV['RUBY_DEBUG_USE_COLORIZE'] = "false"
       ENV['RUBY_DEBUG_TEST_MODE'] = 'true'
 
@@ -92,6 +90,23 @@ module DEBUGGER__
     end
 
     private
+
+    # use this to start a debug session with the test program
+    def manual_debug_code(program)
+      print("[Starting a Debug Session with @#{caller.first}]\n")
+      write_temp_file(strip_line_num(program))
+      inject_lib_to_load_path
+      pid = spawn("#{RUBY} -r debug/open #{temp_file_path}")
+      sleep(0.5) # wait for the debug server to start
+      require_relative "../../lib/debug/client"
+      connect
+    ensure
+      Process.kill('QUIT', pid)
+    end
+
+    def inject_lib_to_load_path
+      ENV['RUBYOPT'] = "-I #{__dir__}/../../lib"
+    end
 
     LINE_NUMBER_REGEX = /^\s*\d+\| ?/
 


test/support/utils.rb
```diff
@@ -36,6 +36,18 @@ def debug_code(program, **options, &block)
       check_line_num!(program)
     end
 
+    # use this to start a debug session with the test program
+    def start_debugger_with(program)
+      write_temp_file(strip_line_num(program))
+      inject_lib_to_load_path
+      pid = spawn("#{RUBY} -r debug/open #{temp_file_path}")
+      sleep(0.5) # wait for the debug server to start
```
ono-max, line 44: I would like to know how to decide seconds.

st0012: I guess you can say it's arbitrary 😛
because there's no way to communicate about the server state atm, the connection will fail right away when the server is not up yet. given that this is just an internal helper to help us write tests, I think it's not worth changing the library's behavior for it.

or perhaps @ko1 has a plan on adding retry & timeout here? https://github.com/ruby/debug/blob/2eed8dfe57cfc8b752001f10515bec0f2cff73e9/lib/debug/client.rb#L75

ono-max:
I'm concerned that this method should be placed in `utils.rb` because I suppose that test files are script files that can be tested automatically such as `rake test`.

st0012:
I don't get the concern. it's not suppose to be committed to the final test scripts, it's only used when "writing" the scripts.
or are you worry about the method visibility? I just defined it as a private method though

ono-max:
Ok, it looks fine to me.

test/support/utils.rb
```diff
@@ -91,6 +89,22 @@ def create_pseudo_terminal(boot_options: "-r debug/run")
 
     private
 
+    # use this to start a debug session with the test program
+    def start_debugger_with(program)
```
ko1, line 93: It is difficult to know the feature from debug log, so could you try to make clearer name for this feature?

test/support/utils.rb
```diff
@@ -91,6 +89,22 @@ def create_pseudo_terminal(boot_options: "-r debug/run")
...
+      write_temp_file(strip_line_num(program))
+      inject_lib_to_load_path
+      pid = spawn("#{RUBY} -r debug/open #{temp_file_path}")
+      sleep(0.5) # wait for the debug server to start
```
ko1, line 97: it is advanced topic: if you use PTY you can wait for the debugger start.

test/support/utils.rb
```diff
@@ -91,6 +89,22 @@ def create_pseudo_terminal(boot_options: "-r debug/run")
...
+      require_relative "../../lib/debug/client"
+      connect
```
ko1, line 99: This is my task but `client.rb` is not expected, so I need to change it to `DEBUGGER__.connect` or something similar.

ko1: Another  comment is, if there is another remote debug session, it will fail, so `debug/open` with specific path file `RUBY_DEBUG_SOCK_PATH` and connect to it is good idea.

test/support/utils.rb
st0012: do you think `_start_manual_debug_session` is a good name? I put an underscore to indicate that it's a helper.

test/support/utils.rb
st0012: if you don't mind, I'll make this improvement in another PR

test/support/utils.rb
st0012: thanks for the hint. I'll update it 👍

test/support/utils.rb
```diff
@@ -91,6 +89,22 @@ def create_pseudo_terminal(boot_options: "-r debug/run")
 
     private
 
+    # use this to start a debug session with the test program
+    def start_debugger_with(program)
+      write_temp_file(strip_line_num(program))
+      inject_lib_to_load_path
+      pid = spawn("#{RUBY} -r debug/open #{temp_file_path}")
```
ko1, line 96: It is kind to show the prompt (ex) `"[PLEASE DEBUG with DEBUGGER @#{caller.fisst}]"`)

test/support/utils.rb
ko1: if you specify the socket file path, it is enough:

```ruby
while !File.exist?(path)
  sleep 0.1
end
```

test/support/utils.rb
st0012: added (you can see it from the description's screenshot)

test/support/utils.rb
st0012: added 👍

test/support/utils.rb
ko1: how about `manual_debug_code`?

test/support/utils.rb
```diff
@@ -91,6 +89,31 @@ def create_pseudo_terminal(boot_options: "-r debug/run")
...
     private
 
+    # use this to start a debug session with the test program
+    def _start_manual_debug_session(program)
+      print("[Starting a Debug Session with @#{caller.first}]\n")
+      write_temp_file(strip_line_num(program))
+
+      require_relative "../../lib/debug/client"
+
+      socket_path = "#{DEBUGGER__.unix_domain_socket_dir}/test-program-debug-session"
```
ko1, line 99: could you use `DEBUGGER__.create_unix_domain_socket_name` to make a unique socket file?

PR #76, opened 2021-06-09
Test for remote debugger

ono-max:
I implement the test framework for remote debugger.

- Assign the argument `remote: false` in `debug_code` method when using `require "debug/run"` or `DEBUGGER__.console`.
- Test for remote and local are in each method by one. That's why if first test(local) failed, second test(remote) won't work.
  - Test for local test will be executed first, and Test for remote(UNIX domain socket mode and TCP/IP mode) will be executed next.

### DIFF
--- a/test/debug/session_test.rb
+++ b/test/debug/session_test.rb
@@ -17,7 +17,7 @@ module DEBUGGER__
     end
 
     def test_session_starts_manually
-      debug_code(program, boot_options: "") do
+      debug_code(program, boot_options: "", remote: false) do
         assert_line_num(5)
         type 'quit'
         type 'y'
@@ -25,7 +25,7 @@ module DEBUGGER__
     end
 
     def test_later_breakpoint_fires_correctly
-      debug_code(program, boot_options: "") do
+      debug_code(program, boot_options: "", remote: false) do
         assert_line_num(5)
         type 'c'
         assert_line_num(6)
@@ -49,7 +49,7 @@ module DEBUGGER__
     end
 
     def test_session_starts_manually
-      debug_code(program, boot_options: "") do
+      debug_code(program, boot_options: "", remote: false) do
         assert_line_num(5)
         type 'quit'
         type 'y'
@@ -57,7 +57,7 @@ module DEBUGGER__
     end
 
     def test_later_breakpoint_fires_correctly
-      debug_code(program, boot_options: "") do
+      debug_code(program, boot_options: "", remote: false) do
         assert_line_num(5)
         type 'c'
         assert_line_num(6)
--- a/test/support/utils.rb
+++ b/test/support/utils.rb
@@ -86,6 +86,12 @@ module DEBUGGER__
           assert_empty_queue(exception: e)
         rescue Timeout::Error => e
           assert false, create_message("TIMEOUT ERROR (#{timeout_sec} sec)")
+        ensure
+          if @server_pid
+            Process.kill(:KILL, @server_pid)
+            Process.waitpid(@server_pid)
+            @server_pid = nil
+          end
         end
       end
     end
--- a/test/support/utils.rb
+++ b/test/support/utils.rb
@@ -14,16 +14,17 @@ module DEBUGGER__
       "#{fail_msg}\n[DEBUG SESSION LOG]\n> " + @backlog.join('> ')
     end
 
-
     def combine_regexps(regexps)
       Regexp.new(regexps.map(&:source).reduce(:+))
     end
 
+    # This method will execute both local and remote mode by default.
     def debug_code(program, **options, &block)
       @queue = Queue.new
-      block.call
+      @scenario = block
+      @scenario.call
       write_temp_file(strip_line_num(program))
-      create_pseudo_terminal(**options)
+      setup_terminal(**options)
       check_line_num!(program)
     end
 
@@ -35,19 +36,51 @@ module DEBUGGER__
     end
 
     RUBY = RbConfig.ruby
-    REPL_RPOMPT = /\(rdbg\)/
+    RUBY_DEBUG_TEST_PORT = '12345'
 
-    def create_pseudo_terminal(boot_options: "-r debug/run")
+    def setup_terminal(boot_options: "-r debug/run", remote: true)
+      ENV['RUBY_DEBUG_TEST_LOCAL_ONLY_MODE'] = 'false' # for fast debug test
       inject_lib_to_load_path
+
+      if remote && ENV['RUBY_DEBUG_TEST_LOCAL_ONLY_MODE'] == 'false'
+        setup_terminal(remote: false) # local test will be executed first
+        @mutex = Mutex.new
+        repl_prompt = /\(rdb\)/
+
+        # run test on Unix domain socket mode
+        boot_options = '-r debug/open'
+        cmd = "#{__dir__}/../../exe/rdbg -A"
+        new_child_process("#{RUBY} #{boot_options} #{temp_file_path}")
+        create_pseudo_terminal(cmd, repl_prompt)
+
+        # run test on TCP/IP mode
+        cmd = "#{__dir__}/../../exe/rdbg -A #{RUBY_DEBUG_TEST_PORT}"
+        new_child_process("#{__dir__}/../../exe/rdbg -O --port=#{RUBY_DEBUG_TEST_PORT} #{temp_file_path}")
+      else
+        # run test on local mode
+        repl_prompt = /\(rdbg\)/
+        cmd = "#{RUBY} #{boot_options} #{temp_file_path}"
+      end
+      create_pseudo_terminal(cmd, repl_prompt)
+    end
+
+    def new_child_process(cmd)
+      @scenario.call
+      @mutex.synchronize do
+        r, _, @server_pid = PTY.spawn(cmd, :in=>'/dev/null', :out=>'/dev/null')
+        r.read(1) # wait for the remote server to boot up
+      end
+    end
+
+    def create_pseudo_terminal(cmd, repl_prompt)
       ENV['RUBY_DEBUG_USE_COLORIZE'] = "false"
       ENV['RUBY_DEBUG_TEST_MODE'] = 'true'
 
       timeout_sec = (ENV['RUBY_DEBUG_TIMEOUT_SEC'] || 10).to_i
 
-      PTY.spawn("#{RUBY} #{boot_options} #{temp_file_path}") do |read, write, pid|
+      PTY.spawn(cmd) do |read, write, _|
         @backlog = []
         @last_backlog = []
-        ask_cmd = ['quit', 'delete', 'kill']
         begin
           Timeout.timeout(timeout_sec) do
             while (line = read.gets)
@@ -67,7 +100,7 @@ module DEBUGGER__
                 write.puts(cmd)
                 @last_backlog.clear
                 next # INTERNAL_INFO shouldn't be pushed into @backlog and @last_backlog
-              when REPL_RPOMPT
+              when repl_prompt
                 # check if the previous command breaks the debugger before continuing
                 if error_index = @last_backlog.index { |l| l.match?(/REPL ERROR/) }
                   raise "Debugger terminated because of: #{@last_backlog[error_index..-1].join}"
--- a/test/support/utils.rb
+++ b/test/support/utils.rb
@@ -28,6 +28,7 @@ module DEBUGGER__
     end
 
     DEBUG_MODE = false
+    ASK_CMD = %w[quit delete kill]
 
     def debug_print msg
       print msg if DEBUG_MODE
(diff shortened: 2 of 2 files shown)


test/support/utils.rb
```diff
@@ -53,11 +57,28 @@ def create_pseudo_terminal(boot_options: "-r debug/run")
 
       timeout_sec = (ENV['RUBY_DEBUG_TIMEOUT_SEC'] || 10).to_i
 
-      PTY.spawn("#{RUBY} #{boot_options} #{temp_file_path}") do |read, write, pid|
+      PTY.spawn("sh") do |read, write, pid|
+        write.puts("export PS1='Hello Prompt\n'")
+        if boot_options == "-r debug/open"
```
st0012, line 62: I think the remote server can be started with just one line? like https://github.com/ruby/debug/pull/75/files#diff-ee14bf65f14927f7778a246f90c5cec0c581be3e2658c2a8a3726c59db5be3f9R43

test/support/utils.rb
```diff
@@ -28,23 +28,27 @@ def assert_fail_with_log msg, lines
                      lines.map{|l| "> #{l}"}.join
     end
 
-    def debug_code(program, **options, &block)
+    def debug_code(program, boot_options: "-r debug/run", remote_debug: true, &block)
```
st0012, line 31: perhaps we can introduce another method like `remote_debug_code` and keep this one simple? we can also name things better in that method without causing any confusion.

test/support/utils.rb
ono-max: Yeah, I will try it.

st0012:
If we can further improve the current test framework and not relying on the internal info for testing (e.g. run tests by purely recording  input/output), we can avoid patching either the server nor client for remote testing. This approach has some benefits:

- We don't need to maintain any extra layers/patches for the testing mode so the lib codebase would be clean.
- Remote debugging (TCP/IP) will be a lot easier to implement because the extra patches won't be not needed.
- It'll be able to test irb integration without extra patches on it.

@ono-max @ko1 wdyt?

ono-max:
@st0012 
We'll discuss it today. Just a moment, please.

test/support/utils.rb
```diff
@@ -51,7 +56,15 @@ def create_pseudo_terminal(boot_options: "-r debug/run")
 
       timeout_sec = (ENV['RUBY_DEBUG_TIMEOUT_SEC'] || 10).to_i
 
-      PTY.spawn("#{RUBY} #{boot_options} #{temp_file_path}") do |read, write, pid|
+      if boot_options == "-r debug/open"
+        r, w, server_pid = PTY.spawn("#{RUBY} #{boot_options} #{temp_file_path}")
+        r.close
+        sleep(0.1)
+        cmd = "exe/rdbg -A"
```
ko1, line 63: should not use this path.

ono-max:
@st0012 
I'm sorry for the late reply.

What you're trying to say is that "we don't want to create extra patches", right?
I can output the internal info without extra patches such as https://github.com/ruby/debug/pull/84, and I don't need to write codes for it.

test/support/utils.rb
ono-max: Could you tell me more about this? I don't get it.

test/support/utils.rb
ono-max: https://github.com/ruby/debug/pull/76/files#diff-ee14bf65f14927f7778a246f90c5cec0c581be3e2658c2a8a3726c59db5be3f9R57

ko1:
could you introduce `local only` mode for fast debug?

ko1:
Also, now UNIX domain socket mode is supported. Could you add an option to test on TCP/IP mode?

test/support/utils.rb
```diff
@@ -14,39 +14,54 @@ def create_message fail_msg
...
-    def create_pseudo_terminal(boot_options: "-r debug/run")
+    def create_pseudo_terminal(boot_options:)
       inject_lib_to_load_path
       ENV['RUBY_DEBUG_USE_COLORIZE'] = "false"
       ENV['RUBY_DEBUG_TEST_MODE'] = 'true'
 
       timeout_sec = (ENV['RUBY_DEBUG_TIMEOUT_SEC'] || 10).to_i
 
-      PTY.spawn("#{RUBY} #{boot_options} #{temp_file_path}") do |read, write, pid|
+      if boot_options == "-r debug/open"
```
ko1, line 50: It has bad smell.

ono-max: I modified it.

ono-max:
> could you introduce local only mode for fast debug?

I did it by an environmental variable.

> Also, now UNIX domain socket mode is supported. Could you add an option to test on TCP/IP mode?

I supported it.

ono-max:
@st0012 
You haven't replied me. Can I resolve conversation with you?

st0012:
@ono-max sorry I missed the notifications. given the scope of this PR, I'll leave the rest of the review to @ko1, so you can resolve the comments now 🙂

test/support/utils.rb
```diff
@@ -14,39 +14,66 @@ def create_message fail_msg
...
+
+      if remote && ENV['RUBY_DEBUG_LOCAL_ONLY_MODE'] == 'false'
+        setup_terminal(remote: false) # local test will be executed first
+        @mutex = Mutex.new
+        boot_options = "-r debug/open"
+        repl_prompt = /\(rdb\)/
+        cmd = "#{__dir__}/../../exe/rdbg -A"
+        new_child_process("#{RUBY} #{boot_options} #{temp_file_path}")
+        create_pseudo_terminal(cmd, repl_prompt)
+        cmd = "#{__dir__}/../../exe/rdbg -A 12345"
```
ko1, line 51: PORT should be given by an option, maybe `RUBY_DEBUG_TEST_PORT` or something.

test/support/utils.rb
```diff
@@ -14,39 +14,66 @@ def create_message fail_msg
...
 
     RUBY = RbConfig.ruby
-    REPL_RPOMPT = /\(rdbg\)/
 
-    def create_pseudo_terminal(boot_options: "-r debug/run")
+    def setup_terminal(boot_options: "-r debug/run", remote: true)
+      ENV['RUBY_DEBUG_LOCAL_ONLY_MODE'] = 'false' # for fast debug test
       inject_lib_to_load_path
+
+      if remote && ENV['RUBY_DEBUG_LOCAL_ONLY_MODE'] == 'false'
```
ko1, line 43: `RUBY_DEBUG_TEST` prefix is better to make clear the purpose.

ono-max: https://github.com/ruby/debug/pull/76/files#diff-ee14bf65f14927f7778a246f90c5cec0c581be3e2658c2a8a3726c59db5be3f9R41

test/support/utils.rb
ono-max: https://github.com/ruby/debug/pull/76/files#diff-ee14bf65f14927f7778a246f90c5cec0c581be3e2658c2a8a3726c59db5be3f9R38

test/debug/session_test.rb
```diff
@@ -17,15 +17,15 @@ def program
     end
 
     def test_session_starts_manually
-      debug_code(program, boot_options: "") do
+      debug_code(program, boot_options: "", remote: false) do
```
st0012, line 20: @ono-max why is `remote: false` needed here? changing `boot_options` shouldn't start remote debugging anyway?

ono-max: Because this test doesn't need to be executed as remote.

> changing boot_options shouldn't start remote debugging anyway?

I don't get it. If I run this test without `remote: false`, it failed. Could you tell me more?

ono-max: @st0012 
Because this test doesn't need to be executed as remote.

> changing boot_options shouldn't start remote debugging anyway?

I don't get it. If I ran this test without `remote: false`, it failed. Could you tell me more?

```shell
ruby test/debug/session_test.rb
Loaded suite test/debug/session_test
Started
.F
================================================================================
Failure: test_session_starts_manually(DEBUGGER__::ConsoleStartTest):
  Expected line number to be 5, but was 1

  [DEBUG SESSION LOG]
  > [1, 7] in /var/folders/kv/w1k6nh1x5fl7vx47b2pd005w0000gn/T/debugger20210701-11874-3zc8c.rb
  > =>    1| a = 1
  >       2| b = 2
  >       3| require "debug"
  >       4| DEBUGGER__.console
  >       5| c = 3
  >       6| binding.bp
  >       7| "foo"
  > =>#0	<main> at /var/folders/kv/w1k6nh1x5fl7vx47b2pd005w0000gn/T/debugger20210701-11874-3zc8c.rb:1
/Users/ono-max/workspace/debug/test/support/assertions.rb:8:in `block in assert_line_num'
/Users/ono-max/workspace/debug/test/support/utils.rb:87:in `block (2 levels) in create_pseudo_terminal'
/Users/ono-max/.rbenv/versions/3.0.0/lib/ruby/3.0.0/timeout.rb:97:in `block in timeout'
/Users/ono-max/.rbenv/versions/3.0.0/lib/ruby/3.0.0/timeout.rb:35:in `block in catch'
/Users/ono-max/.rbenv/versions/3.0.0/lib/ruby/3.0.0/timeout.rb:35:in `catch'
/Users/ono-max/.rbenv/versions/3.0.0/lib/ruby/3.0.0/timeout.rb:35:in `catch'
/Users/ono-max/.rbenv/versions/3.0.0/lib/ruby/3.0.0/timeout.rb:112:in `timeout'
/Users/ono-max/workspace/debug/test/support/utils.rb:79:in `block in create_pseudo_terminal'
/Users/ono-max/workspace/debug/test/support/utils.rb:75:in `spawn'
/Users/ono-max/workspace/debug/test/support/utils.rb:75:in `create_pseudo_terminal'
/Users/ono-max/workspace/debug/test/support/utils.rb:58:in `setup_terminal'
/Users/ono-max/workspace/debug/test/support/utils.rb:45:in `setup_terminal'
/Users/ono-max/workspace/debug/test/support/utils.rb:26:in `debug_code'
test/debug/session_test.rb:20:in `test_session_starts_manually'
     17:     end
     18:
     19:     def test_session_starts_manually
  => 20:       debug_code(program, boot_options: "") do
     21:         assert_line_num(5)
     22:         type 'quit'
     23:         type 'y'
================================================================================
..
Finished in 0.83998 seconds.
4 tests, 9 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
75% passed
4.76 tests/s, 10.71 assertions/s
```

st0012: sorry that I initially thought it's a toggle like choosing either local or remote mode. but it looks like all tests are now tested in both local and remote mode, and `remote: false` is to skip the remote part, right?

if that's the case, can you document this behavior (`debug_code` runs both local and remote mode) above the `debug_code` method?

test/support/utils.rb
```diff
@@ -14,39 +14,67 @@ def create_message fail_msg
...
-    def create_pseudo_terminal(boot_options: "-r debug/run")
+    def setup_terminal(boot_options: "-r debug/run", remote: true)
+      ENV['RUBY_DEBUG_TEST_LOCAL_ONLY_MODE'] = 'false' # for fast debug test
       inject_lib_to_load_path
+
+      if remote && ENV['RUBY_DEBUG_TEST_LOCAL_ONLY_MODE'] == 'false'
+        setup_terminal(remote: false) # local test will be executed first
+        @mutex = Mutex.new
+        boot_options = "-r debug/open"
+        repl_prompt = /\(rdb\)/
```
st0012, line 48: it'd be great if we can add some comments to this section to make things clearer

```ruby
        # run test in local mode
        cmd = "#{__dir__}/../../exe/rdbg -A"
        new_child_process("#{RUBY} #{boot_options} #{temp_file_path}")

        # run test in remote mode
        repl_prompt = /\(rdb\)/
        create_pseudo_terminal(cmd, repl_prompt)
        cmd = "#{__dir__}/../../exe/rdbg -A #{RUBY_DEBUG_TEST_PORT}"
        new_child_process("#{__dir__}/../../exe/rdbg -O --port=#{RUBY_DEBUG_TEST_PORT} #{temp_file_path}")
```

test/debug/session_test.rb
ono-max: > all tests are now tested in both local and remote mode, and remote: false is to skip the remote part, right?

Exactly 😄 

> if that's the case, can you document this behavior (debug_code runs both local and remote mode) above the debug_code method?

Sure.

test/support/utils.rb
ono-max: https://github.com/ruby/debug/pull/76/files#diff-ee14bf65f14927f7778a246f90c5cec0c581be3e2658c2a8a3726c59db5be3f9R50-R62

test/debug/session_test.rb
ono-max: https://github.com/ruby/debug/pull/76/files#diff-ee14bf65f14927f7778a246f90c5cec0c581be3e2658c2a8a3726c59db5be3f9R21

ono-max:
@st0012 
I modified code based on your comments. Are they ok for you?

st0012:
@ono-max yes. thanks for the quick update 👍

PR #80, opened 2021-06-10
Add tests for require 'debug/run' scenario

st0012:


test/debug/session_test.rb
```diff
@@ -34,4 +34,36 @@ def test_later_breakpoint_fires_correctly
...
+       3| require "debug/run"
+       4|
+       5| c = 3
+       6| binding.bp
+       7| "foo"
+      RUBY
+    end
+
+    def test_session_starts_manually
+      debug_code(program, boot_options: "") do
```
ko1, line 52: FYI: `nil` is also ok for `boot_options`

ko1 (approved):

PR #81, opened 2021-06-11
add missing "#"

ono-max:
I add missing `#`.

```shell
[2, 11] in target.rb
      2|   class Bar
      3|     def self.a
      4|       "hello"
      5|     end
      6|     def b(n)
=>    7|       2.times do
      8|         n
      9|       end
     10|     end
     11|   end
=>#0	Foo::Bar#b(n=1) at target.rb:7
  #1	<module:Foo> at target.rb:19
  #2	<main> at target.rb:1

Stop by # BP - Check  n==1 0
```

### DIFF
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -935,7 +935,7 @@ module DEBUGGER__
       case type
       when :suspend_bp
         bp, i = *args
-        puts "\nStop by \##{i} #{bp}" if bp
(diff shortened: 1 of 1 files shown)


ko1:
Please write *current behavior* and *expected behavior* in a description and it will help understanding.

ko1:
talked at zoom.

PR #82, opened 2021-06-11
Use markdown header for usage options

st0012:
It's kinda hard to distinguish different usage options in a large code block containing both text, Ruby code and shell commands. I think it'd be easier to read if we give different options a header.

### Before

截圖 2021-06-16 上午10 37 41

### After

截圖 2021-06-16 上午10 37 13


st0012:
@ko1 wdyt?

PR #83, opened 2021-06-11
Fix appearance of "Stop by~"

ono-max:
swap locations of variables "bp" and "i"

## Now
```shell
[2, 11] in target.rb
      2|   class Bar
      3|     def self.a
      4|       "hello"
      5|     end
      6|     def b(n)
=>    7|       2.times do
      8|         n
      9|       end
     10|     end
     11|   end
=>#0	Foo::Bar#b(n=1) at target.rb:7
  #1	<module:Foo> at target.rb:19
  #2	<main> at target.rb:1

Stop by # BP - Check  n==1 0
```

## What I expect

```shell
[2, 11] in target.rb
      2|   class Bar
      3|     def self.a
      4|       "hello"
      5|     end
      6|     def b(n)
=>    7|       2.times do
      8|         n
      9|       end
     10|     end
     11|   end
=>#0	Foo::Bar#b(n=1) at target.rb:7
  #1	<module:Foo> at target.rb:19
  #2	<main> at target.rb:1

Stop by #0  BP - Check  n == 1
```

### DIFF
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -934,7 +934,7 @@ module DEBUGGER__
     def event type, *args
       case type
       when :suspend_bp
-        bp, i = *args
+        i, bp = *args
         puts "\nStop by \##{i} #{bp}" if bp
       when :suspend_trap
         puts "\nStop by #{args.first}"


lib/debug/session.rb
```diff
@@ -935,7 +935,7 @@ def event type, *args
       case type
       when :suspend_bp
         bp, i = *args
```
st0012, line 937: I think it's because the local assignments are wrong. it should be this instead:

```suggestion
        i, bp = *args
```

the current message is actually correct.

st0012 (changes requested):

lib/debug/session.rb
st0012: the caller:

https://github.com/ruby/debug/blob/bbf475f976703788e6622b682b5a58d68297395a/lib/debug/session.rb#L108

ono-max: You are right, Thanks!

ono-max:
Don't mind closing pull request. It was just my mistake.

PR #84, opened 2021-06-12
Improve test framework - remove test_console.rb

ono-max:
It's better to remove `test_console.rb` because we don't have to maintain extra patches for tests. I did two things.
- change the place where INTERNAL_INFO is generated as a json
- modify PTY block


lib/debug/breakpoint.rb
```diff
@@ -253,6 +253,8 @@ def setup
       @tp = TracePoint.new(:line){|tp|
         next if tp.path.start_with? __dir__
         next if tp.path.start_with? '<internal:'
+        # Skip when `require 'json'` and `JSON.generate` are called during tests
+        next if tp.path.start_with? RbConfig.expand '$(libdir)'
```
st0012, line 257: doesn't this mean all Ruby stdlib will be skipped? this changes how debugger works so probably should have more explanation.
it's also not clear to me that why `JSON.generate` would trigger the `CheckBreakpoint`. we may want to fix this instead of skipping all the stdlib.

test/support/utils.rb
```diff
@@ -18,7 +18,7 @@ def assert_line_num(expected)
 
     def assert_line_text(expected)
       @queue.push(Proc.new {
-        assert_match(expected, @last_backlog[2..].join, create_message("Expected to include #{expected}"))
+        assert_match(expected, @last_backlog[1..].join, create_message("Expected to include #{expected}"))
```
st0012, line 21: I'm curious about the first element of `@last_backlog`. is it the input value?

lib/debug/test_console.rb
```diff
@@ -1,29 +0,0 @@
-# frozen_string_literal: true
```
st0012, line 1: 🎉

lib/debug/session.rb
```diff
@@ -188,7 +191,10 @@ def wait_command_loop tc
 
     def wait_command
       if @initial_commands.empty?
-        @ui.puts(@internal_info) if ENV['RUBY_DEBUG_TEST_MODE']
+        if ENV['RUBY_DEBUG_TEST_MODE']
+          require 'json'
```
st0012, line 195: `require` should not be called inside a method that'll be repeatedly called. we can place it right before this method (e.g. line 191) and add a comment to say it's needed for generating internal info.

test/support/utils.rb
ono-max: Yes, it's input value such as `(rdbg) step`.

lib/debug/session.rb
ono-max: Thanks! I will place it outside of this method.

lib/debug/breakpoint.rb
ono-max: > doesn't this mean all Ruby stdlib will be skipped? this changes how debugger works so probably should have more explanation.

Exactly. I changed it.

> it's also not clear to me that why JSON.generate would trigger the CheckBreakpoint.

FYI, `if State === opts` in `generate` method is one of the triggers in this tracepoint.
https://github.com/flori/json/blob/master/lib/json/common.rb#L297

test/support/utils.rb
st0012: I see. thanks for answering

lib/debug/breakpoint.rb
ko1: good catch. could you give me a call-stack when it is called? Maybe it should be change the test framework design.

lib/debug/session.rb
```diff
@@ -186,9 +186,10 @@ def wait_command_loop tc
       @tc = nil
     end
 
+    require 'json' if ENV['RUBY_DEBUG_TEST_MODE']
```
ko1, line 189: In general (without any reasons), `require` should be located at the beginning of file and it make easy to know what kind of libraries are uses in this file.

ono-max: Thanks, I put `require` in beginning of file.

lib/debug/breakpoint.rb
ono-max:
```shell
["/Users/[user_name]/workspace/debug/lib/debug/breakpoint.rb:14:in `safe_eval'", "/Users/[user_name]/workspace/debug/lib/debug/breakpoint.rb:260:in `block in setup'", "/Users/[user_name]/.rbenv/versions/3.0.0/lib/ruby/3.0.0/json/common.rb:297:in `generate'", "/Users/[user_name]/workspace/debug/lib/debug/session.rb:193:in `wait_command'", "/Users/[user_name]/workspace/debug/lib/debug/session.rb:177:in `block (2 levels) in wait_command_loop'", "/Users/[user_name]/workspace/debug/lib/debug/session.rb:176:in `loop'", "/Users/[user_name]/workspace/debug/lib/debug/session.rb:176:in `block in wait_command_loop'", "/Users/[user_name]/workspace/debug/lib/debug/session.rb:816:in `block in stop_all_threads'", "<internal:trace_point>:196:in `enable'", "/Users/[user_name]/workspace/debug/lib/debug/session.rb:815:in `stop_all_threads'", "/Users/[user_name]/workspace/debug/lib/debug/session.rb:175:in `wait_command_loop'", "/Users/[user_name]/workspace/debug/lib/debug/session.rb:116:in `session_server_main'", "/Users/[user_name]/workspace/debug/lib/debug/session.rb:75:in `block in initialize'"]
```

ono-max: Are they what you want?

ko1: I'll fix it later.

ko1:
Conflicts are found. could you check it?

ono-max:
@ko1 
I did.

PR #85, opened 2021-06-13
Make sure duplicated breakpoint is disabled

st0012:
The debugger currently shows a warning message for duplicated breakpoints and doesn't add it to the breakpoints list. But line or catch breakpoints would still be activated and stop the program.

To fix this issue, we need to disable duplicated breakpoints. However, `binding.bp` also uses line breakpoints with a fixed internal location, simply excluding duplicated breakpoints will break it.

So the solution requires 2 checks:

1. Check if the breakpoint is duplicated by key.
2. Check if it's a normal breakpoint (e.g. added by the `break` command) instead of a `oneshot` (binding.bp or DEBUGGER__.console) breakpoint.

If both are satisfied, the debugger should warn the user and disable the breakpoint.

It also fixes the issue that multiple `binding.bp` raises duplicated breakpoints warning, which doesn't make sense.

### DIFF
--- a/lib/debug/breakpoint.rb
+++ b/lib/debug/breakpoint.rb
@@ -64,6 +64,10 @@ module DEBUGGER__
       to_s
     end
 
+    def duplicable?
+      false
+    end
+
     class << self
       include Color
 
@@ -155,6 +159,11 @@ module DEBUGGER__
       else
         # not actiavated
       end
+    end
+
+    def duplicable?
+      # only binding.bp or DEBUGGER__.console are duplicable
+      @oneshot
     end
 
     NearestISeq = Struct.new(:iseq, :line, :events)
--- a/test/debug/bp_test.rb
+++ b/test/debug/bp_test.rb
@@ -50,5 +50,13 @@ module DEBUGGER__
         type 'q!'
       end
     end
+
+    def test_debugger_doesnt_complain_about_duplicated_breakpoint
+      debug_code(program) do
+        type 'continue'
+        assert_no_line_text(/duplicated breakpoint:/)
+        type 'q!'
+      end
+    end
   end
 end
--- a/test/debug/break_test.rb
+++ b/test/debug/break_test.rb
@@ -73,6 +73,17 @@ module DEBUGGER__
       end
     end
 
+    def test_debugger_rejects_duplicated_method_breakpoints
+      debug_code(program) do
+        type 'break Foo::Baz.c'
+        type 'break Foo::Baz.c'
+        assert_line_text(/duplicated breakpoint/)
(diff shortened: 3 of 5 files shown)


test/debug/break_test.rb
```diff
@@ -206,6 +206,18 @@ def test_stops_at_correct_place_when_breakpoint_set_in_empty_line
       end
     end
 
+    def test_conditional_breakpoint_stops_for_repeated_iterations
```
st0012, line 209: to make sure the change doesn't prevent the same breakpoint from stopping multiple times.

test/debug/bp_test.rb
```diff
@@ -50,5 +50,13 @@ def test_breakpoint_execute_command_argument_correctly
         type 'q!'
       end
     end
+
+    def test_debugger_doesnt_complain_about_duplicated_breakpoint
+      debug_code(program) do
+        type 'continue'
+        assert_no_line_text(/duplicated breakpoint:/)
```
st0012, line 57: currently, writing multiple `binding.bp` triggers this warning. but the PR also fixes this.

st0012:
@ko1 this is ready for review now 😄

lib/debug/breakpoint.rb
```diff
@@ -64,6 +64,11 @@ def description
       to_s
     end
 
+    def duplicable?
+      # only binding.bp or DEBUGGER__.console are duplicable
+      is_a?(LineBreakpoint) && oneshot
```
ko1, line 69: Could you use inheritance?

define false and define it in `LineBreakpoint`.

lib/debug/breakpoint.rb
```diff
@@ -77,7 +82,7 @@ class LineBreakpoint < Breakpoint
     LABEL = generate_label("Line")
     PENDING_LABEL = generate_label("Line (pending)")
 
-    attr_reader :path, :line, :iseq
+    attr_reader :path, :line, :iseq, :oneshot
```
ko1, line 85: and `oneshot` is not needed.

lib/debug/breakpoint.rb
st0012: sounds good. will do 👍

st0012:
I accidentally deleted the branch 😬

ko1:
Question: MethodBreakpoint isn't duplicable?

st0012:
@ko1 do you mean I missed method breakpoint's test case? I've added it.

ko1:
Ah, I misread this ticket. Only LineBreakpoint is *NOT* duplicable, I missed. I see.

PR #86, opened 2021-06-13
Improve assert methods

ono-max:
- Override assert_block method to call create_message only when test fails
- use assert_block to output clearer fail logs

### DIFF
--- a/test/support/utils.rb
+++ b/test/support/utils.rb
@@ -10,6 +10,12 @@ module DEBUGGER__
       @queue.push(command)
     end
 
+    def assert_block(message)
+      _wrap_assertion do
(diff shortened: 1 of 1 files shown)


test/support/utils.rb
```diff
@@ -4,21 +4,35 @@
 require 'timeout'
 require 'json'
 
+#
+# assert methods based on test-unit project
+#
+# The License is https://github.com/test-unit/test-unit/blob/master/BSDL
+#
+
```
ono-max, line 12: I'm not sure if this is enough for requirements in License of test-unit.

ko1:
please consider the way without redefine the existing assert methods.

PR #88, opened 2021-06-14
Improve assert methods [2nd pattern]

ono-max:
This is an another approach to improve assert methods.

- use assert_block to output clearer fail logs
- override "to_s" method in "assert_block" method
  - We don't have to modify `assert_block` in `test-unit`.

### DIFF
--- a/test/support/assertions.rb
+++ b/test/support/assertions.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+module DEBUGGER__
+  module AssertionHelpers
+    def assert_line_num(expected)
+      @queue.push(Proc.new {
+        msg = "Expected line number to be #{expected}, but was #{@internal_info['line']}\n"
+        assert_block(MessageCreationController.new { create_message(msg) }) { expected == @internal_info['line'] }
+      })
+    end
+
+    def assert_line_text(expected)
+      @queue.push(Proc.new {
+        result = @last_backlog[2..].join
+        expected = Regexp.escape(expected) if expected.is_a?(String)
+        msg = "Expected to include `#{expected}` in\n(\n#{result})\n"
+        assert_block(MessageCreationController.new { create_message(msg) }) { result.match? expected }
+      })
+    end
+  end
+end
+
+class MessageCreationController
+  def initialize &block
+    @msg = nil
+    @create_msg = block
+  end
+
+  def to_s
+    return @msg if @msg
+
+    @msg = @create_msg.call
+  end
+end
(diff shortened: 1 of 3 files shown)


test/support/assertions.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+#
+# Copied `assert_block` method from test-unit project
+#
+# The License is https://github.com/test-unit/test-unit/blob/master/BSDL
+#
```
ono-max, line 7: I'm not sure if this is enough for requirements in License of test-unit.
I'd like some advices

ko1:
Do you need to redefine methods?

```ruby
require 'test/unit'

class MessageBuilder
  def initialize &b
    @b = b
    @msg = nil
  end

  def to_s
    return @msg if @msg
    @msg = @b.call
  end
end

class TestFoo < Test::Unit::TestCase
  def time_consuming_message
    sleep 3
    'msg'
  end

  def test_foo
    assert_equal 2, 1+1, MessageBuilder.new{ time_consuming_message }
    assert_equal 3, 1+1, MessageBuilder.new{ time_consuming_message }
  end
end
```

test/support/assertions.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module Test
+  module Unit
+    module Assertions
```
st0012, line 5: From my experience, we normally don't add project-specific helpers to a library's modules. We'd do this instead:

```ruby
module DEBUGGER__
  module AssertionHelpers
    include Test::Unit::Assertions # this may not be necessary

    def assert_line_num(expected)
      # .....
    end

# in test_case.rb

module DEBUGGER__
  class TestCase < Test::Unit::TestCase
    include TestUtils
    include AssertionHelpers

```

ono-max: @st0012 
Thanks for giving me information, but I'd like discuss more. Project-specific helpers are added to a library's modules in some ruby projects. What do you think?

https://github.com/ruby/irb/blob/master/test/lib/test/unit/core_assertions.rb
https://github.com/ruby/ruby/blob/master/tool/lib/test/unit/core_assertions.rb

ko1: If we run tests of different projects in a process, it is a bit safer to isolate the definitions.

ono-max: Ok, that way sounds better. I changed it.

test/support/assertions.rb
```diff
@@ -0,0 +1,31 @@
...
+
+class MessageCreationController
+  def initialize &block
+    @create_msg = block
+  end
+
+  def to_s
+    @create_msg.call
+  end
+end
```
ono-max, line 31: Should I include this class in `module DEBUGGER__`?

PR #95, opened 2021-06-16
Add delete tests

st0012:


test/support/utils.rb
```diff
@@ -68,7 +68,7 @@ def create_pseudo_terminal(boot_options: "-r debug/run")
               when /INTERNAL_INFO:\s(.*)/
                 @internal_info = JSON.parse(Regexp.last_match(1))
                 next # INTERNAL_INFO shouldn't be pushed into @backlog and @last_backlog
-              when %r{Really quit\? \[Y/n\]}
+              when %r{\[(y|Y)/(n|N)\]}
```
ko1, line 71: it can introduce mistake, so I'll fix it later.

PR #100, opened 2021-06-17
Rename MessageCreationController

ono-max:
`MessageCreationController` is not a clear name, @ko1 said. I did rename `MessageCreationController` to `FailureMessage`.

### DIFF
--- a/test/support/assertions.rb
+++ b/test/support/assertions.rb
@@ -5,7 +5,7 @@ module DEBUGGER__
     def assert_line_num(expected)
       @queue.push(Proc.new {
         msg = "Expected line number to be #{expected}, but was #{@internal_info['line']}\n"
-        assert_block(MessageCreationController.new { create_message(msg) }) { expected == @internal_info['line'] }
+        assert_block(FailureMessage.new { create_message(msg) }) { expected == @internal_info['line'] }
       })
     end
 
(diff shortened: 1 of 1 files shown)


test/support/assertions.rb
```diff
@@ -14,13 +14,13 @@ def assert_line_text(expected)
...
         msg = "Expected to include `#{expected}` in\n(\n#{result})\n"
-        assert_block(MessageCreationController.new { create_message(msg) }) { result.match? expected }
+        assert_block(ComposeMessageWhenFailed.new { create_message(msg) }) { result.match? expected }
       })
     end
   end
 end
 
-class MessageCreationController
+class ComposeMessageWhenFailed
```
st0012, line 23: imo class name should be a "thing" instead of an "action". for example: `AssertionMessageComposer`.

ono-max: Oh, I see... 🤔 Thank you!

st0012: Let me put it this way: if `ComposeMessageWhenFailed` can be performed directly, like `ComposeMessageWhenFailed.perform`, I think it's a proper name. Many Rails users do name their job classes this way because the API is `SomeJob.perform_later`.

But in this case, we need to initialize an instance of the class, naming it after an action (verb) is semantically wrong. Because a verb (compose), can not have instances.

(I hope you can understand it. I'm not a native speaker so it's hard for me to explain this concept.)

ono-max: Yeah, maybe It does make sense to me. If this class is used as a class method, the name of class can be verb, but if this class is used as an instance, it should be not verb. Did I get that right?

st0012: @ono-max exactly 😄

PR #101, opened 2021-06-17
Add trace tests

st0012:
This PR also adds a new `assert_no_line_text` test assertion helper.

### DIFF
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -499,7 +499,6 @@ module DEBUGGER__
         case arg
         when 'on'
           dir = __dir__
-          i = 0
           @tracer ||= TracePoint.new(:call, :return, :b_call, :b_return, :line, :class, :end){|tp|
             next if File.dirname(tp.path) == dir
             next if tp.path == '<internal:trace_point>'
--- a/test/support/assertions.rb
+++ b/test/support/assertions.rb
@@ -11,11 +11,32 @@ module DEBUGGER__
 
     def assert_line_text(expected)
       @queue.push(Proc.new {
-        result = @last_backlog[1..].join
+        result = collect_recent_backlog
         expected = Regexp.escape(expected) if expected.is_a?(String)
         msg = "Expected to include `#{expected}` in\n(\n#{result})\n"
-        assert_block(MessageCreationController.new { create_message(msg) }) { result.match? expected }
+
+        assert_block(MessageCreationController.new { create_message(msg) }) do
+          result.match? expected
+        end
       })
+    end
+
+    def assert_no_line_text(expected)
+      @queue.push(Proc.new {
+        result = collect_recent_backlog
+        expected = Regexp.escape(expected) if expected.is_a?(String)
+        msg = "Expected not to include `#{expected}` in\n(\n#{result})\n"
+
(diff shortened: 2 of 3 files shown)


lib/debug/session.rb
```diff
@@ -239,6 +239,7 @@ def wait_command
       #   * Finish debugger (with the debuggee process on non-remote debugging).
       when 'q', 'quit'
         if ask 'Really quit?'
+          disable_tracer
```
ko1, line 242: why does it needed?

st0012: because debugger would hang when tracer is on. you can reproduce the bug with

```
$ exe/rdbg -e  'trace on;; quit!' target.rb
```

ko1: Thank you. I'll check it.

ko1: 9d8f49389c24680ea53370c2d9210d1485b9ee5c may fix this issue.

ko1: for checking `at_exit` or similar feature, trace will be useful.

st0012: @ko1 cool thanks. I'll drop that commit.

st0012:
@ko1 I happen to found another bug caused by https://github.com/ruby/debug/commit/e4365550b48ca715ab1a1552afa3c4af5338f9b4 and fixed it in https://github.com/ruby/debug/pull/101/commits/9b68b0e0dc91d236141c7b9ad60371cb2384ac6b

st0012:
@ko1 ah sorry I think it's because I forgot to compile the extension 🤦

test/support/assertions.rb
```diff
@@ -9,14 +9,26 @@ def assert_line_num(expected)
...
-        assert_block(MessageCreationController.new { create_message(msg) }) { result.match? expected }
+        msg = msg.sub("to", "not to") if inverse
+
+        assert_block(MessageCreationController.new { create_message(msg) }) do
+          if inverse
+            !result.match? expected
+          else
+            result.match? expected
+          end
+        end
```
ono-max, line 25: Can we separate `assert_block` method and `assert_no_line_text`?

There are 2 reason.
1. Time complexity
`assert_block` method will be executed more than `assert_no_line_text`. Every time `assert_block` method are executed, it have to execute conditional branch two times. This process is a bit CPU-intensive because this method is called a lot.

2. Visual complexity
It's a little hard to understand the codes. test-unit also separates `assert_equal` and `assert_not_equal`
https://github.com/test-unit/test-unit/blob/master/lib/test/unit/assertions.rb#L224
https://github.com/test-unit/test-unit/blob/master/lib/test/unit/assertions.rb#L739

ono-max (changes requested):

test/support/assertions.rb
st0012: > Time complexity

evaluating a boolean in if condition is extremely cheap. `debug_print` evals a boolean value a lot more frequently than the `assert_block`. so tbh I don't think this is a sufficient reason for change.

> Visual complexity

ok I'll change it.

ono-max (approved):
It looks good to me!

test/support/assertions.rb
ono-max: Thanks.

PR #102, opened 2021-06-17
Unify watch & method breakpoint handling

st0012:
Currently, `Session` handles breakpoints in 3 different ways:

- `watch` breakpoints - enqueued as an `eval` thread client event. 
- `method` breakpoints - enqueued as a `breakpoint` thread client event.
- other breakpoints - initialized and registered directly without being enqueued.

This PR makes `Session` handle watch breakpoints similar to method breakpoints as:

- `watch`/`method` breakpoints - enqueued as a `breakpoint` thread client event.
- other breakpoints - initialized and registered directly without being enqueued.

This refactor will make future changes on breakpoints easier (for example, prevent duplicated breakpoints).

**It doesn't change any output or functionality**

**Before**

master

**After**

unify-breakpoint-registration

### DIFF
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -405,7 +405,8 @@ module DEBUGGER__
           puts e.message
           ::DEBUGGER__::METHOD_ADDED_TRACKER.enable
         end
-        event! :result, :method_breakpoint, bp
+
+        bp
       else
         raise "unknown breakpoint: #{args}"
       end
@@ -578,11 +579,10 @@ module DEBUGGER__
           event! :result, nil
 
         when :breakpoint
-          add_breakpoint args
-
+          bp = add_breakpoint args
+          event! :result, :method_breakpoint, bp
         when :dap
           process_dap args
-
         else
           raise [cmd, *args].inspect
         end
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -407,6 +407,9 @@ module DEBUGGER__
         end
 
         bp
+      when :watch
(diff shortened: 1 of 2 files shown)


lib/debug/thread_client.rb
```diff
@@ -576,13 +563,31 @@ def wait_next_action
...
+            result = frame_eval(ivar)
 
+            if @success_last_eval
+              object =
+                if b = current_frame.binding
+                  b.receiver
+                else
+                  current_frame.self
+                end
+              puts "#{object} #{ivar} = #{result}"
```
st0012, line 582: I think this can go now because it's the same content as the breakpoint's string representation.

```
#<Student:0x00007fa8f99465c0> @name = John # printed by this line 
#0  BP - Watch  #<Student:0x00007fa8f99465c0> @name = John # printed by show_bps
```

lib/debug/thread_client.rb
```diff
@@ -394,7 +394,7 @@ def show_object_info expr
       end
     end
 
-    def add_breakpoint args
+    def initialize_breakpoint args
```
st0012, line 397: I was confused by this name and it doesn't actually "add" a breakpoint to anything. so I think it's good to rename it.

ko1: `make`?

st0012: 👍 updated

lib/debug/thread_client.rb
st0012: @ko1 I've added the last commit to remove it.

PR #104, opened 2021-06-18
Fix trace tests

st0012:
`Regexp.union` doesn't create `AND` combination but `OR` combination, which makes the tests inaccurate.


ko1:
BTW `assert_line_text` can accept Array of regexp/string.

st0012:
@ko1 ok I'll make a PR to support it 👍 
(currently the argument will be passed into `match?` and cause error)

Issue #65, opened 2021-06-04
Allow using binding.bp to start a debug session


st0012:
As stated in the readme, the correct way to start a debug session in the middle of a program is:

```ruby
require "debug"
DEBUGGER__.console
```

And then the user can freely use `binding.bp` to insert new breakpoints.

This is a fine workflow, but I think we can make it more simpler. Since we already have the `binding.bp` method, which just looks like `byebug` and `binding.pry`, why don't we make it work like them?

```ruby
require "pry"
binding.pry

require "byebug"
byebug

require "debug"
binding.bp
```

Then the users will have one less thing to learn 🙂 

(I have to admit that it took me a while to remember to call `DEBUGGER__.console`, which I still forget from time to time 😂)

ko1:
Yes, it is planned idea.

The drawback is, explicit debug start will track the program execution:

* Thread creation
* Code loading including `eval` (store loaded Ruby code to show)
* And more (not yet)

In many cases it is not a problem.

-----
Terminology:

* Load debug functionality: Loading the debug classes/modules/methods
  * `require 'debug'`
  * `require 'debug/session'`
* Start debug: load debug functionality and start tracking the information for debugging. Also load `.rdbgrc` for configuration.
  * `require 'debug/run'`
  * `DEBUGGER__.console`
* Stop in a debugger context:
  * breakpoint
  * `binding.bp`
  * Ctrl-C on (remote) console

----

Maybe this proposal is, if `binding.bp` is called without start debugging, start debugging and stop at this location.

ko1:
Another possibility is, `require 'debug'` start debgging but don't stop the program (== `RUBY_DEBUG_NONSTOP` mode).
The drawback is, if user doesn't want to collect information (don't want to consume any overhead even if `require 'debug' is called), the interpreter starts by adding `gem 'debug'` in `Gemfile`.

ko1:
In a future, like `binding.irb`, I want to introduce it as standard. But there is an issue describe above.

st0012:
> Maybe this proposal is, if binding.bp is called without start debugging, start debugging and stop at this location.

That's what I have in mind too. Should I open a PR for it?

ko1:
I showed some points, and we need to choose one.

st0012:
Ah sorry I didn't stress that clearly: I took that line as 

`if binding.bp is called without start debugging, start "debugging session" and stop at this location.`, which is like

```ruby
# move this out from Session.initialize_session
[...]

PR #110, opened 2021-06-20
Simplify debugger starting flow

st0012:
This implements all the behaviors described in https://github.com/ruby/debug/issues/65#issuecomment-864499744

### DIFF
--- a/lib/debug/console.rb
+++ b/lib/debug/console.rb
@@ -93,14 +93,29 @@ module DEBUGGER__
     end
   end
 
-  def self.console **kw
-    set_config(kw)
-
-    set_starter_breakpoint
-
+  def self.set_prev_handler
     @prev_handler = trap(:SIGINT){
       ThreadClient.current.on_trap :SIGINT
     }
+  end
+
+  def self.stop_at_require
+    set_config({})
+    set_starter_breakpoint
+    set_prev_handler
+  end
+
+  def self.console command: nil, **kw
+    set_config(kw)
+
+    if command
+      cmds = command.split(";;")
+      SESSION.add_initial_commands cmds
+    end
+
+    ::DEBUGGER__.add_line_breakpoint __FILE__, __LINE__ + 1, oneshot: true
+
+    set_prev_handler
   end
 
   initialize_session(UI_Console.new)
--- a/lib/debug/console.rb
+++ b/lib/debug/console.rb
@@ -92,5 +92,16 @@ module DEBUGGER__
       trap(:INT, prev_handler)
     end
   end
-end
 
+  def self.console **kw
+    set_config(kw)
+
+    set_starter_breakpoint
+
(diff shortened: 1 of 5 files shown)


test/debug/session_test.rb
```diff
@@ -18,15 +18,15 @@ def program
 
     def test_session_starts_manually
       debug_code(program, boot_options: "") do
-        assert_line_num(5)
+        assert_line_num(4)
```
st0012, line 21: This change actually makes `DEBUGGER__.console` stop more accurately. It used to stop at the next line but it now stops at the same line.

lib/debug/session.rb
```diff
@@ -98,6 +98,11 @@ def initialize ui
       @tp_thread_begin.enable
     end
 
+    def set_ui(ui)
+      @ui = ui
+      @management_threads << @ui.reader_thread if @ui.respond_to? :reader_thread
```
st0012, line 103: Because:
- only remote ui needs an extra thread for reader
- the default ui is console ui

we don't need to consider the `remote -> console` transition.

lib/debug/console.rb
```diff
@@ -92,5 +92,31 @@ def setup_interrupt
...
+    }
+  end
+
+  def self.stop_at_require
+    set_config({})
+    set_starter_breakpoint
+    set_prev_handler
+  end
+
+  def self.console command: nil, **kw
```
st0012, line 108: it's not needed for the remote mode, so it's better to define it here.

lib/debug/session.rb
```diff
@@ -1038,53 +1043,44 @@ def self.open host: nil, port: ::DEBUGGER__::CONFIG[:port], sock_path: nil, sock
...
   end
 
   def self.open_tcp host: nil, port:, **kw
     set_config(kw)
     require_relative 'server'
-    initialize_session UI_TcpServer.new(host: host, port: port)
+    ui = UI_TcpServer.new(host: host, port: port)
+
+    if defined?(SESSION)
+      DEBUGGER__::SESSION.set_ui(ui)
```
st0012, line 1056: in scenarios like 

```ruby
require "debug"
a = 1
b = 2
DEBUGGER__.open
c = 3
d = 4
```

the debugger will first be initialized with `UI_Console`. and instead of initializing another session with a different ui, we can just assign it.

lib/debug/session.rb
```diff
@@ -1038,53 +1043,44 @@ def self.open host: nil, port: ::DEBUGGER__::CONFIG[:port], sock_path: nil, sock
...
+    else
+      initialize_session(ui)
```
st0012, line 1058: this is for the `-r debug/open` scenario.

ko1:
Thank you I will use this as a reference. But I want to reconsider the boot process from scratch.

PR #103, opened 2021-06-18
Add list tests

st0012:
This needs https://github.com/ruby/debug/pull/112 to work.

PR #112, opened 2021-06-22
Support multiple assertions after one command

ono-max:
@st0012 pointed out the problem that test framework doesn't support multiple assert methods after one command.
https://github.com/ruby/debug/pull/103#discussion_r656355053
This PR will fix it.


st0012 (approved):
what a brilliant fix 👍

PR #114, opened 2021-06-23
Remove duplicated evaluation

st0012:
For `[:eval, :call]` events (`eval` and `irb` commands), `ThreadClient` evaluates the source twice:

https://github.com/ruby/debug/blob/fbe17413d5e7c33a51502c2ca1763ed3bcae4ed0/lib/debug/thread_client.rb#L480-L484
https://github.com/ruby/debug/blob/fbe17413d5e7c33a51502c2ca1763ed3bcae4ed0/lib/debug/thread_client.rb#L494-L495

This is unnecessary and can cause some issues (like it'll trigger 2 `irb` sessions).

### DIFF
--- a/test/debug/eval_test.rb
+++ b/test/debug/eval_test.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+require_relative '../support/test_case'
+
+module DEBUGGER__
+  class EvalTest < TestCase
+    def program
+      <<~RUBY
+     1| a = "foo"
+     2| b = "bar"
+     3| c = 3
+     4| __END__
+      RUBY
+    end
+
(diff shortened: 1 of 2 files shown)


test/debug/eval_test.rb
```diff
@@ -0,0 +1,38 @@
...
+        type 'b 3'
+        type 'continue'
+        type 'e a.upcase!'
+        type 'p a'
+        assert_line_text(/"FOO"/)
+        type 'q!'
+      end
+    end
+
+    def test_eval_evaluates_computation_and_assignment
```
st0012, line 27: this one fails on master

ko1:
💯

PR #116, opened 2021-06-28
Add using with bundler in README

igaiga:
Write how to use with bundler in README.

As far as I know, "1.0.0.a" means "--pre" on Gemfile.
https://guides.rubygems.org/patterns/#pessimistic-version-constraint

I've tested the following codes.

Gemfile

```
# frozen_string_literal: true

source "https://rubygems.org"

git_source(:github) { |repo_name| "https://github.com/#{repo_name}" }

gem "debug", ">= 1.0.0.beta"
```

target.rb

```ruby
ENV['RUBY_DEBUG_NONSTOP'] = 'true'
require 'debug/run'

class Foo
  def foo
    string = "foo!!!"
    binding.bp # or binding.debug
    string
  end
end

p Foo.new.foo
```

```
$ rdbg -c bundle exec ruby target.rb
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,10 +29,14 @@ $ gem install debug --pre
 
 or specify `-Ipath/to/debug/lib` in `RUBYOPT` or each ruby command-line option, especially for debug this gem development.
 
-If you use Bunlder, write the following line to your Gemfile.
+If you use Bundler, write the following line to your Gemfile. And use rdbg command with -c option.
 
 ```
-gem "debug", ">= 1.0.0.a"
+gem "debug", ">= 1.0.0.beta"
+```
+
+```
+$ rdbg -c bundle exec ruby target.rb
 ```
 
 # How to use
--- a/README.md
+++ b/README.md
@@ -29,6 +29,12 @@ $ gem install debug --pre
 
 or specify `-Ipath/to/debug/lib` in `RUBYOPT` or each ruby command-line option, especially for debug this gem development.
 
+If you use Bunlder, write the following line to your Gemfile.
+
+```
+gem "debug", ">= 1.0.0.a"
+```
+
 # How to use
 
 ## Invoke with debugger
--- a/misc/README.md.erb
+++ b/misc/README.md.erb
@@ -29,6 +29,16 @@ $ gem install debug --pre
 
 or specify `-Ipath/to/debug/lib` in `RUBYOPT` or each ruby command-line option, especially for debug this gem development.
 
+If you use Bundler, write the following line to your Gemfile. And use rdbg command with -c option.
+
+```
+gem "debug", ">= 1.0.0.beta"
+```
+
+```
+$ rdbg -c bundle exec ruby target.rb
+```
+
 # How to use
 
 ## Invoke with debugger


README.md
```diff
@@ -29,6 +29,12 @@ $ gem install debug --pre
 
 or specify `-Ipath/to/debug/lib` in `RUBYOPT` or each ruby command-line option, especially for debug this gem development.
 
+If you use Bunlder, write the following line to your Gemfile.
+
+```
+gem "debug", ">= 1.0.0.a"
```
st0012, line 35: great suggestion 👍  
but I think `beta` would be easier to understand?
`gem "debug", ">= 1.0.0.beta"`

ko1:
Can I merge it or wait for the fix pointed by @st0012 (a -> beta)? I'm not sure how to specify the versions.

BTW,

```ruby
class Foo
  def foo
    string = "foo!!!"
    require 'debug/run'
    string
  end
end

p Foo.new.foo
```

is wrong example.

(1) modify the source code:

If you don't worry about modifying the source code, use the following.

```
ENV['RUBY_DEBUG_NONSTOP'] = 'true'
require 'debug/run'

class Foo
  def foo
    string = "foo!!!"
    binding.bp # or binding.debug
    string
  end
end

p Foo.new.foo
```

in this case, the debugger stops the execution at the `binding.bp`.

NOTE: This will be changed (https://github.com/ruby/debug/issues/65):

```
# Gemfile
gem 'debug'...

# target.rb
class Foo
  def foo
    string = "foo!!!"
    binding.bp # or binding.debug
    string
  end
end
```

simply it can stop at `binding.bp` line.

(2) Use `rdbg` command

```
$ rdbg -c bundle exec ruby target.rb
```

and it will stops at the beginning of target.rb. and setup breakpoint with "b 4" (4 for line number), and continue with "continue" command, the debugger stops the program at the line 4.

(3) Use VSCode extension

Try [VSCode rdbg Ruby Debugger - Visual Studio Marketplace](https://marketplace.visualstudio.com/items?itemName=KoichiSasada.vscode-rdbg) and setup breakpoint with IDE.

README.md
```diff
@@ -29,6 +29,12 @@ $ gem install debug --pre
 
 or specify `-Ipath/to/debug/lib` in `RUBYOPT` or each ruby command-line option, especially for debug this gem development.
 
+If you use Bunlder, write the following line to your Gemfile.
```
ko1, line 32: Could you modify `misc/README.md.erb` instead of it? and `rake` (without taskname) will re-make `README.md`.

adam12, line 32:
```suggestion
If you use Bundler, write the following line to your Gemfile.
```

README.md
igaiga: It looks good! I've fixed it.

igaiga:
Thank you all!

I've fixed followings.

- Change description of this PR to fix examples.
- Add `$ rdbg -c bundle exec ruby target.rb` command to README.
- Fixed spelling "bundler"
- Modify misc/readme.md.erb and run rake commad.

README.md
igaiga: Thanks!

igaiga: Fixed it.

ko1:
Thank you!

PR #123, opened 2021-07-03
Add display tests

st0012:


lib/debug/session.rb
```diff
@@ -478,9 +478,7 @@ def process_command line
         case arg
         when /(\d+)/
           if @displays[n = $1.to_i]
-            if ask "clear \##{n} #{@displays[n]}?"
```
st0012, line 481: I don't think confirmation should be required when deleting a specific display setting. This is unnecessary imo and inconsistent with the `delete` command.

PR #125, opened 2021-07-04
Allow assert_line_text helper to take array expectation

st0012:
As suggested in https://github.com/ruby/debug/pull/104#issuecomment-863797701

### DIFF
--- a/test/assertion_helper_test.rb
+++ b/test/assertion_helper_test.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+require_relative 'support/test_case'
+
+module DEBUGGER__
+  class AssertLineTextTest < TestCase
+    def program
+      <<~RUBY
+        a = 1
+      RUBY
+    end
+
+    def test_the_helper_takes_a_string_expectation_and_escape_it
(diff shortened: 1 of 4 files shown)


ono-max (approved):
Thanks!

ko1:
Can I confirm the spec?

https://github.com/ruby/debug/pull/125/files#diff-9b998a59b2de86476be86794ffe0d672575e0d38d8b6ce30dab32ee9062267a5R23

it seems `assert_line_text([/foo/, /bar/])` only match `/foobar/`. Is it your intention?

```
text = <<EOS
foo
bar
baz
EOS

assert_line_text([/foo/, /baz/])
```

I think the above example should match.

ko1:
Anyway, I merged it. Please make another PR for the further improvements if needed.

ono-max:
> I think the above example should match.

Can I create a PR for it?
I need the feature for test-builder.

Thank you.

st0012:
@ko1 I think your suggestion makes more sense 👍 
@ono-max feel free to open a PR for it. remember to also update the tests in https://github.com/ruby/debug/blob/master/test/debug/trace_test.rb though

PR #129, opened 2021-07-05
Improve test framework

ono-max:
Currently, `check_line_num` method will be executed after the test finished. The line numbers should be checked first because `test/assertion_helper_test.rb` and `test/test_utils_test.rb` will exit before `check_line_num` method is executed. This PR will fix it.

I also changed constant variable(`DEBUG_MODE`) to ENV variable(`ENV["RUBY_DEBUG_TEST_DEBUG_MODE"]`).

### DIFF
--- a/test/support/utils.rb
+++ b/test/support/utils.rb
@@ -16,6 +16,8 @@ module DEBUGGER__
 
     # This method will execute both local and remote mode by default.
     def debug_code(program, boot_options: '-r debug/run', remote: true, &block)
+      check_line_num!(program)
+
       @scenario = block
       write_temp_file(strip_line_num(program))
       inject_lib_to_load_path
@@ -29,8 +31,6 @@ module DEBUGGER__
         debug_on_unix_domain_socket
         debug_on_tcpip
       end
-
-      check_line_num!(program)
     end
 
     DEBUG_MODE = false
--- a/test/support/utils.rb
+++ b/test/support/utils.rb
@@ -33,11 +33,10 @@ module DEBUGGER__
       end
     end
 
-    DEBUG_MODE = false
     ASK_CMD = %w[quit delete kill undisplay]
 
     def debug_print msg
-      print msg if DEBUG_MODE
+      print msg if ENV['RUBY_DEBUG_TEST_DEBUG_MODE']
     end
 
     RUBY = RbConfig.ruby
--- a/test/assertion_helper_test.rb
+++ b/test/assertion_helper_test.rb
@@ -6,7 +6,7 @@ module DEBUGGER__
   class AssertLineTextTest < TestCase
     def program
       <<~RUBY
-        a = 1
+        1| a = 1
       RUBY
     end
 
--- a/test/support/test_case.rb
+++ b/test/support/test_case.rb
@@ -16,7 +16,7 @@ module DEBUGGER__
     end
 
     def remove_temp_file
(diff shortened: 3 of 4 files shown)


test/support/utils.rb
```diff
@@ -29,8 +31,6 @@ def debug_code(program, boot_options: '-r debug/run', remote: true, &block)
         debug_on_unix_domain_socket
         debug_on_tcpip
       end
-
-      check_line_num!(program)
     end
 
     DEBUG_MODE = false
```
st0012, line 36: I'd love to see it become a env var so I can always enable it locally. wdyt?

ono-max: Could you explain me more? I just moved `check_line_num` up. It is always enable, right?

ono-max: @st0012 
Could you explain me more? I just moved `check_line_num` up. It is always enable, right?

st0012: I mean can we make this:

```ruby
DEBUG_MODE = ENV["RUBY_DEBUG_TEST_DEBUG_MODE"]
```

so I don't need to change the code to enabled it. like what you did and forgot to change it back (I've done this a lot too 😂 ).

ono-max: Oh, ok. I thought you were talking about `check_line_num`. I'll change `DEBUG_MODE`, too.

ono-max: fccf992

test/support/utils.rb
```diff
@@ -29,15 +31,13 @@ def debug_code(program, boot_options: '-r debug/run', remote: true, &block)
...
-      check_line_num!(program)
     end
 
-    DEBUG_MODE = false
+    ENV['RUBY_DEBUG_TEST_DEBUG_MODE'] ||= 'false'
     ASK_CMD = %w[quit delete kill undisplay]
 
     def debug_print msg
-      print msg if DEBUG_MODE
+      print msg if ENV['RUBY_DEBUG_TEST_DEBUG_MODE']
```
st0012, line 40: shouldn't this be 

```ruby
print msg if ENV['RUBY_DEBUG_TEST_DEBUG_MODE'] != "false"
```
because it's a string instead of boolean?

ono-max: Thanks! That was my mistake.

test/support/utils.rb
```diff
@@ -29,15 +31,13 @@ def debug_code(program, boot_options: '-r debug/run', remote: true, &block)
...
-      check_line_num!(program)
     end
 
-    DEBUG_MODE = false
+    ENV['RUBY_DEBUG_TEST_DEBUG_MODE'] ||= 'false'
     ASK_CMD = %w[quit delete kill undisplay]
 
     def debug_print msg
-      print msg if DEBUG_MODE
+      print msg unless ENV['RUBY_DEBUG_TEST_DEBUG_MODE'] == 'false'
```
st0012, line 40: thanks 👍

ko1: `if ENV['RUBY_DEBUG_TEST_DEBUG_MODE']` ?

ono-max: I changed it.

ono-max: https://github.com/ruby/debug/pull/129/files#diff-ee14bf65f14927f7778a246f90c5cec0c581be3e2658c2a8a3726c59db5be3f9R39

PR #133, opened 2021-07-06
Use setup_remote_debuggee in manual_debug_code too

st0012:


test/support/utils.rb
```diff
@@ -162,25 +156,28 @@ def run_test_scenario(cmd, repl_prompt)
...
 
-      socket_path = DEBUGGER__.create_unix_domain_socket_name
-      inject_lib_to_load_path
-      ENV["RUBY_DEBUG_SOCK_PATH"] = socket_path
-      pid = spawn("#{RUBY} -r debug/open #{temp_file_path}")
-
-      while !File.exist?(socket_path)
-        sleep 0.1
-      end
+      setup_remote_debuggee("#{RDBG_EXECUTABLE} -O -- #{temp_file_path}")
```
ko1, line 176: I think `socket_path` should be remaind.
`debug_on_unix_domain_socket` should also use it.

Actually, the tests fail when there are other remote debug sessions (and I got this failure :p)

test/support/utils.rb
```diff
@@ -162,25 +156,28 @@ def run_test_scenario(cmd, repl_prompt)
...
+        @remote_debuggee_pid = nil
+      end
+    end
+
     # use this to start a debug session with the test program
     def manual_debug_code(program)
       print("[Starting a Debug Session with @#{caller.first}]\n")
       write_temp_file(strip_line_num(program))
 
       require_relative "../../lib/debug/client"
```
ko1, line 174: is it needed yet?

st0012:
Close because of #133

PR #134, opened 2021-07-06
Improve remote (socket) test setup

st0012:


test/support/utils.rb
```diff
@@ -3,6 +3,7 @@
 require 'pty'
 require 'timeout'
 require 'json'
+require_relative "../../lib/debug/client"
```
st0012, line 6: This is required for:
- `DEBUGGER__.create_unix_domain_socket_name`
- `DEBUGGER__::Client`

ono-max (approved):

test/support/utils.rb
```diff
@@ -162,25 +151,40 @@ def run_test_scenario(cmd, repl_prompt)
...
         sleep 0.1
       end
 
       DEBUGGER__::Client.new([socket_path]).connect
     ensure
-      Process.kill('QUIT', pid)
+      kill_remote_debuggee
+    end
+
+    def setup_socket_remote_debuggee
```
ko1, line 179: `socket_remote_debuggee` is ambiguous (TCP/IP also uses socket).

how about `setup_unix_doman_socket_remote_debuggee`?

st0012: 👍 updated

PR #111, opened 2021-06-20
test builder

ono-max:
https://user-images.githubusercontent.com/59436572/122718083-74730b00-d2a7-11eb-9db2-c50a117175ce.mov

### What's this?

This is a test generator to make it easier to write tests.

## Generate Tests
There is a test generator in `debug.rb` project to make it easier to write tests.
### Quickstart
This section shows you how to create test file by test generator. For more advanced informations on creating tests, please take a look at [gentest options](#gentest-options). (You can also check by `$bin/gentest -h`)
#### 1. Create a target file for debuggee.
Let's say, we created `target.rb` which is located in top level directory of debugger.
```ruby
module Foo
  class Bar
    def self.a
      "hello"
    end
  end
  Bar.a
  bar = Bar.new
end
```
#### 2. Run `gentest` as shown in the example below.
```shell
$ bin/gentest target.rb
```
#### 3. Debugger will be executed. You can type any debug commands.
```shell
$ bin/gentest target.rb
[1, 9] in ~/workspace/debug/target.rb
=>    1| module Foo
      2|   class Bar
      3|     def self.a
      4|       "hello"
      5|     end
      6|   end
      7|   Bar.a
      8|   bar = Bar.new
      9| end
=>#0	<main> at ~/workspace/debug/target.rb:1
INTERNAL_INFO: {"location":"~/workspace/debug/target.rb:1","line":1}

(rdbg)s
 s
[1, 9] in ~/workspace/debug/target.rb
      1| module Foo
=>    2|   class Bar
      3|     def self.a
      4|       "hello"
      5|     end
      6|   end
      7|   Bar.a
      8|   bar = Bar.new
      9| end
=>#0	<module:Foo> at ~/workspace/debug/target.rb:2
  #1	<main> at ~/workspace/debug/target.rb:1
INTERNAL_INFO: {"location":"~/workspace/debug/target.rb:2","line":2}

(rdbg)n
 n
[1, 9] in ~/workspace/debug/target.rb
      1| module Foo
      2|   class Bar
=>    3|     def self.a
      4|       "hello"
      5|     end
      6|   end
      7|   Bar.a
      8|   bar = Bar.new
      9| end
=>#0	<class:Bar> at ~/workspace/debug/target.rb:3
  #1	<module:Foo> at ~/workspace/debug/target.rb:2
  #2	<main> at ~/workspace/debug/target.rb:1
INTERNAL_INFO: {"location":"~/workspace/debug/target.rb:3","line":3}

(rdbg)b 7
[...]

PR #135, opened 2021-07-06
remove manual_debug_code method

ono-max:
We don't have to take a few step in #75 because we can use [test-builder](https://github.com/ruby/debug/pull/111)


ko1:
It is different feature, to debug existing test code.

st0012:
As @ko1 said, I often use it to test things with existing test cases too. Please keep it 🙂

PR #138, opened 2021-07-07
Support filtering in backtrace command

### LINKED ISSUE #137: Support filtering frames in backtrace command
Current `backtrace` always prints all frames, which could be a huge number in Rails applications (in my project it could be 170 frames). Some simple filtering would be helpful:

1. `backtrace [num]` - only prints the first num frames.
2. `backtrace /controller/` - only prints frames with paths that match `controller`

st0012:
Closes #137

### DIFF
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -432,6 +432,8 @@ module DEBUGGER__
         case arg
         when /\A(\d+)\z/
           @tc << [:show, :backtrace, arg.to_i]
+        when /\A\/.*\/\z/
+          @tc << [:show, :backtrace, eval(arg)]
         else
           @tc << [:show, :backtrace]
         end
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -434,6 +434,9 @@ module DEBUGGER__
           @tc << [:show, :backtrace, arg.to_i]
         when /\A\/.*\/\z/
           @tc << [:show, :backtrace, eval(arg)]
+        when /\A(\d+)\s+(\/.*\/)\z/
+          max, pattern = $1, $2
+          @tc << [:show, :backtrace, max.to_i, eval(pattern)]
         else
           @tc << [:show, :backtrace]
         end
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -437,14 +437,15 @@ module DEBUGGER__
       when 'bt', 'backtrace'
         case arg
         when /\A(\d+)\z/
-          @tc << [:show, :backtrace, arg.to_i]
-        when /\A\/.*\/\z/
-          @tc << [:show, :backtrace, eval(arg)]
-        when /\A(\d+)\s+(\/.*\/)\z/
+          @tc << [:show, :backtrace, arg.to_i, nil]
+        when /\A\/(.*)\/\z/
+          pattern = $1
+          @tc << [:show, :backtrace, nil, Regexp.compile(pattern)]
+        when /\A(\d+)\s+\/(.*)\/\z/
           max, pattern = $1, $2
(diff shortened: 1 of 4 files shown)


lib/debug/session.rb
```diff
@@ -428,8 +428,24 @@ def process_command line
 
       # * `bt` or `backtrace`
       #   * Show backtrace (frame) information.
+      # * `bt <num>`
```
ko1, line 431: `bt <num>` or `backtrace <num>` ?

lib/debug/session.rb
```diff
@@ -428,8 +428,24 @@ def process_command line
...
+      #   * Only shows frames with location info that matches `/location_regexp/`.
+      # * `bt <num> /location_regexp/`
+      #   * Only shows first `<num>` frames with location info that matches `/location_regexp/`.
       when 'bt', 'backtrace'
-        @tc << [:show, :backtrace]
+        case arg
+        when /\A(\d+)\z/
+          @tc << [:show, :backtrace, arg.to_i]
+        when /\A\/.*\/\z/
+          @tc << [:show, :backtrace, eval(arg)]
```
ko1, line 442: do not use `eval`, but use `Regexp.compile`.

ko1: The message format should be `[:show, :backtrace, num, filter_regexp]`.
Do not mix the types.

ko1: and simply `nil` represents "not specified".

lib/debug/thread_client.rb
```diff
@@ -359,21 +359,27 @@ def frame_eval src, re_raise: false
...
+            @target_frames.select { |f| f.location_str.match?(location_pattern) }
+          else
+            @target_frames
+          end
+
+        size = frames.size
         max.times{|i|
           break if i >= size
-          puts frame_str(i)
+          puts frame_str(i, frames: frames)
```
ko1, line 382: Before filter

```
#0: foo
#1: bar
#2: boo
#3: baz
```

with filter /a/

```
#0: bar
#1: baz
```

But it is another option:

```
#1: bar
#3: baz
````

(remain original index).

Pros:
* We can recognize how deep frames.
Cons:
* Need to input huge number like 123 even if only 4 frames are shown.

Comment (ko1) [reply lib/debug/thread_client.rb:382]: I prefer remaining original index. But not sure how users feel...

Comment (st0012) [reply lib/debug/thread_client.rb:382]: I think it provides better information to users 👍 
But this requires assigning the frame index to `FrameInfo`, which is a bit outside the scope of this PR. Can I open another one for this?

Comment (ko1) [reply lib/debug/thread_client.rb:382]: You don't need to store the index into FrameInfo, but you need to collect with index like,

```ruby
fs = []
frames.each_with_index([]){|f, i|
  r << [i, f] if f.match(pat)
}```

Comment (ko1) [reply lib/debug/thread_client.rb:382]: ah, and we can choose this frame with `frame` command with original index.

Comment (st0012) [reply lib/debug/thread_client.rb:382]: I see 🤔 I'll open another PR for this after this one is merged.

File: lib/debug/thread_client.rb [hunk 4]
```diff
@@ -549,7 +555,19 @@ def wait_next_action
 
           case type
           when :backtrace
-            show_frames
+            max_lines, pattern = args
```
ko1, line 558: This code can make more shorter. I'll fix it.

ko1:
c9d9f58910d0f6eb2f641996730bbe12a99e57a9 fixed the index issue.
Also support to filter by method name.
Do you have any problem about that?

st0012:
of course I don't have a problem about that 😅 
I think it should be helpful too 👍

PR #150, opened 2021-07-09
Check if there are any errors in @last_backlog

ono-max:
### Example target file
```ruby
# frozen_string_literal: true

require_relative '../support/test_case'

module DEBUGGER__
  class BindingBPWithCommandTest < TestCase
    def program
      <<~RUBY
      1| at_exit{
      2|   binding.bp
      3|   eval('a = 1')
      4|   Thread.new{}
      5| }
      6|
      7| binding.bp(command: 'q!')
      RUBY
    end

    def test_foo
      debug_code(program) do
        type 's'
        type 'n'
      end
    end
  end
end

```

### Before
Current test framework doesn't check if there are any errors in @last_backlog after `q!`.
```shell
$ RUBY_DEBUG_TEST_NO_REMOTE=1 ruby test/debug/foo_test.rb
Loaded suite test/debug/foo_test
Started
.
Finished in 0.560616 seconds.
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
1 tests, 1 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
1.78 tests/s, 1.78 assertions/s
```
### After
```shell
$ RUBY_DEBUG_TEST_NO_REMOTE=1 ruby test/debug/foo_test.rb
Loaded suite test/debug/foo_test
Started
E
================================================================================================================================================================================================================================================================================
Error: test_foo(DEBUGGER__::BindingBPWithCommandTest):
  RuntimeError: Debugger terminated because of: Traceback (most recent call last):
  	4: from /var/folders/kv/w1k6nh1x5fl7vx47b2pd005w0000gn/T/debugger20210709-56507-1wv0lsb.rb:2:in `block in <main>'
  	3: from /Users/naotto/workspace/debug/lib/debug/thread_client.rb:207:in `block in step_tp'
  	2: from /Users/naotto/workspace/debug/lib/debug/thread_client.rb:182:in `on_suspend'
  	1: from /Users/naotto/workspace/debug/lib/debug/thread_client.rb:463:in `wait_next_action'
  /Users/naotto/workspace/debug/lib/debug/session.rb:1094:in `check_forked': DEBUGGER: stop at forked process is not supported yet. (RuntimeError)
   on LOCAL mode
  [DEBUGGER SESSION LOG]
[...]

PR #152, opened 2021-07-09
Fix debugger exception detection

st0012:
1. The exception detection introduced in https://github.com/ruby/debug/pull/150 doesn't apply to GNU/Linux platforms. A test fails after the detection is added but it only presents on some platforms (like macOS)

```
✦ ❯ ruby test/debug/catch_test.rb -n /reject/
Loaded suite test/debug/catch_test
Started
E
=========================================================================================================================================================================================================================
Error: test_debugger_rejects_duplicated_catch_bp(DEBUGGER__::BasicCatchTest):
Traceback (most recent call last):d because of:
        1: from /var/folders/yg/hnbymwxd5pn7v94_clc59y6r0000gn/T/debugger20210709-94267-3vfzfn.rb:4:in `<main>'
  /var/folders/yg/hnbymwxd5pn7v94_clc59y6r0000gn/T/debugger20210709-94267-3vfzfn.rb:4:in `/': divided by 0 (ZeroDivisionError)
```
2. The detection shouldn't use `Traceback` for detection because it'll include the exceptions raised by the program itself.

So this PR fixes these 2 issues and introduce better formatted messages for debugger exceptions.

### DIFF
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -634,7 +634,7 @@ module DEBUGGER__
     rescue SystemExit
       raise
     rescue Exception => e
-      pp [__FILE__, __LINE__, e, e.backtrace]
+      pp ["DEBUGGER Exception: #{__FILE__}:#{__LINE__}", e, e.backtrace]
       raise
     ensure
       set_mode nil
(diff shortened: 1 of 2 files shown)


ko1:
> The detection shouldn't use Traceback for detection because it'll include the exceptions raised by the program itself.

Exception can be thrown at other locations. I'll try to fix it later.

PR #155, opened 2021-07-09
Add `binding.bp` to README

adam12:
Document use of `binding.bp` to setup breakpoints inside application,
without immediately invoking the debugger console.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,12 +83,31 @@ require 'debug/run' # start the debug console
 # ... rest of program ...
 ```
 
+
 ```
 $ ruby target.rb
 ```
 
 When you run the program with the debug console, you will see the debug console prompt `(rdbg)`.
 The debuggee program (`target.rb`) is suspended at the beginning of `target.rb`.
+
+
+Alternatively, start the debugger at a specific location in your program using `binding.bp`.
+
+```ruby
+# target.rb
+require 'debug' # start the debugger
+
+# ... program ...
+
+binding.bp # setup a breakpoint at this line
+
+# ... rest of program ...
+```
+
+```
+$ ruby target.rb
+```
 
 You can type any debugger's command described bellow. "c" or "continue" resume the debuggee program.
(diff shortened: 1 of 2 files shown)


st0012:
I actually think this should be the first entry as it's how I used to use `pry`/`byebug` and now use this tool. I imagine many people (maybe most Rails devs?) would use it this way too.

ko1:
I agree that many Rubyists familiar with `binding.pry` culture.

However, honestly speaking, I don't think current `binding.pry` culture is good solution.
So I don't want to make it first.

Editors or IDEs should support to specify the breakpoints.
Current debug.rb doesn't support them, so we need to improve.

ko1:
Anyway, current doc is not smart, so I'll rewrite it.

ko1:
Another concern is, if we push `binding.bp` like `binding.pry`. people will claim that the debugger console doesn't have useful features like `pry` :p

st0012:
> Another concern is, if we push `binding.bp` like `binding.pry`. people will claim that the debugger console doesn't have useful features like `pry` :p

I understand the concern but I'm not too worry about it.

Having `binding.bp` usage obvious is to let a big portion of Rubyists buy the ideal quickly. Once they enter the debug session they'll quickly find it's not `pry` at all. Then they'll come back to see what it really is. (This is what I experienced when I first tried this project)

So I'd layout the readme like this:

1. Installation 
1. Local usages (put `binding.bp` first)
1. List all commands
1. Remote usages
1. Initial Scripts
1. Env vars


The goal is to make developers try it first in the most easy and familiar way, and learn the basic usage of this tool. I believe this is very important because I didn't start using it until [direct binding.bp](https://github.com/ruby/debug/issues/65) is implemented this week. Even though I've been contributing this project for 2 months.

Having a familiar starting point is important for people to adopt a tool. But once they give it a try and have an impression on it, we can gradually change the culture by introducing better IDE integrations, better designed APIs....etc.

PR #157, opened 2021-07-09
Frame Filter POC

st0012:
To register a frame filter:

```ruby
# so extension gems don't need to require the entire debugger to register the filter
require "debug/frame_filter"

DEBUGGER__.register_inclusion_pattern(/#{Rails.root}\/(app|config|lib|test)/)
```

The filters are applied to:

- Control flow commands. For example, pressing `step` will skip frames with `library` in its path.
- `backtrace` command. So frames with `library` in its path will not be displayed in the output. But the displayed frames will keep the original index.
    - I'm thinking adding `backtrace all` to show all frames.

The filters are not applied to:

- `frame`, `up` and `down` commands. I figure precise frame operations should still be available.

#### Scripts

```ruby
# library.rb

def wrapper
  yield
end

def foo
  10
end
```

```ruby
# target.rb
DEBUGGER__.register_exclusion_pattern(/library/)

require_relative "library"

def bar
  20
end

result = wrapper do
  foo + bar
end

puts(result)

```

Command: `$ exe/rdbg -e "b 8 ;; c" target.rb`

**Without Filtering**

You can see the step enters into methods defined in `library.rb`

截圖 2021-07-11 下午4 18 04

**With Filtering**

Step now ignores methods from `library.rb`. And frames from `library.rb` are also hidden without changing the original index.

截圖 2021-07-11 下午4 18 49

### Rails Example

(It's a demo app so the code is very simple with not many application frames)

Filtering Script:

```ruby
# defining patterns seems to be simpler than extracting and modifying them from Rails' BacktraceCleaner
DEBUGGER__.register_inclusion_pattern(/#{Rails.root}\/(app|config|lib|test|spec)/)
```

**Without Filtering**

The debugger steps right into Rails internal:

截圖 2021-07-09 下午11 30 16

**With Filtering**

Stays inside application code:

截圖 2021-07-11 下午4 21 19


lib/debug/frame_filter.rb
```diff
@@ -0,0 +1,10 @@
+module DEBUGGER__
```
st0012, line 1: This could be required alone so extensions don't need to require the debugger and activate the debugger unnecessarily.

lib/debug/frame_filter.rb
```diff
@@ -0,0 +1,10 @@
...
+  def self.register_frame_filter(&block)
+    @frame_filters ||= []
```
st0012, line 3: I imagine there could be multiple debugger extensions in the future, so having multiple filters would be possible.

ko1:
This spec is not ractor-safe so I can't accept this spec.

Also it should be discussed that we need to allow

> # so extension gems don't need to require the entire debugger to register the filter

st0012:
> Also it should be discussed that we need to allow

I don't have strong opinion on this. But I thought it's something that you'd prefer? So do you think it's better to activate the debugger before extensions adding anything to it?

> This spec is not ractor-safe so I can't accept this spec.

Can you let me know how to make it ractor-safe? Is it because it uses a singleton class' instance variable?

st0012:
Would it be safer if we store the registered filters in `DEBUGGER__::CONFIG`?

ko1:
Registering the data is safe on ractors, but Proc is not safe on multi-ractors.
Registering patterns are not enough?

like:

```
() set bt-exclude /gem_path/
() set bt-include  /specific_gem_name/
```

st0012:
I see. Inserting patterns can probably work. I just need to find a way to extract them from Rails' backtrace cleaner 🤔 
From your example I assume you want it to be a config option and can be set/update with the `set` command? If that's the case, are you ok with supporting an array of values? That'll require adding `add` and `remove` action to the `set` command because ideally we should support multiple patterns.

**Not directly related**

I think the `set` command could be renamed as `config` with a bit different format so it can be more flexible:

```
config # list all configs, same as `set`
config key # same as `set key`
config set key value # same as `set key=value`
config unset key  

# for multi-value configs
config add key value 
config remove key index
```

ko1:
same as `git config`?

st0012:
@ko1 yes, pretty similar to git and many CLI tools' config syntax.

ko1:
sorry, I don't know other tools (because I'm not sure recent tools). Could you list some?

ko1:
>  I just need to find a way to extract them from Rails' backtrace cleaner 🤔

It is good example. At least Rails is not a ractor-safe, so it is convenient to collaborate this feature. hmm.

st0012:
> sorry, I don't know other tools (because I'm not sure recent tools). Could you list some?

GitHub's CLI `gh` 

```
❯ gh config -h
Display or change configuration settings for gh.

Current respected settings:
- git_protocol: the protocol to use for git clone and push operations (default: "https")
- editor: the text editor program to use for authoring text
- prompt: toggle interactive prompting in the terminal (default: "enabled")
- pager: the terminal pager program to send standard output to


USAGE
  gh config <command> [flags]

CORE COMMANDS
  get:        Print the value of a given configuration key
  set:        Update configuration with a value for the given key

INHERITED FLAGS
  --help   Show help for command

LEARN MORE
  Use 'gh <command> <subcommand> --help' for more information about a command.
  Read the manual at https://cli.github.com/manual
```

ko1:
Allow to pattern to write a method call like `:send:foo` is one idea.

ko1:
my misunderstand. `git config` has different style.

```
git config foo # get foo
git config foo bar # set foo as bar
git config --unset foo
```

st0012:
> Allow to write a method call in patterns like `:send:foo` is one idea.

I don't get this. can you explain more?

st0012:
@ko1 I've updated the implementation to use patterns instead of procs.

ko1:
I added: 6b525f7f1303c4404c70598831522e9b52b415eb

* set -> config
* skip_path configuration

st0012:
I think frame filtering can apply to CatchBreakpoint as well. It's not uncommon to see Rails using Exception for flow control (e.g. using `NameError` in `ActiveSupport`). So `catch Exception` would frequently enter the Rails internal too.

ko1:
> I think frame filtering can apply to CatchBreakpoint as well. It's not uncommon to see Rails using Exception for flow control (e.g. using NameError in ActiveSupport). So catch Exception would frequently enter the Rails internal too.

Sorry, I can't catch. What will be filtered? Ignoring exception when it is raised at skip_path?

st0012:
@ko1 yeah.
but of course exceptions are different than backtrace and normal frames, directly ignoring them could cause more confusion (as people don't always aware settings of the tools they use).

ko1:
For this purpose, we need to introduce different configuration, 

skip_frame_path
skip_raise_path

for example. anyway, current skip_path is ambiguous so I rename it to skip_frame_path.

st0012:
@ko1 the current `skip_path` doesn't skip on frame control commands. is that intentional?

PR #162, opened 2021-07-11
color_pp's default width shouldn't be nil

st0012:
Reproduction Script:

```ruby
class Foo
  def initialize
    @string = "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
  end
end

f = Foo.new

binding.bp
```

```
$ exe/rdbg -e 'c ;; info ;; q!' target.rb
```

**Before**

截圖 2021-07-11 下午5 03 38

**After**

截圖 2021-07-11 下午5 03 52

This only happens with coloring enabled, so it's not catchable by tests.

### DIFF
--- a/lib/debug/color.rb
+++ b/lib/debug/color.rb
@@ -24,7 +24,7 @@ module DEBUGGER__
     end
 
     if defined? IRB::ColorPrinter.pp
-      def color_pp obj, width = nil
+      def color_pp obj, width = SESSION.width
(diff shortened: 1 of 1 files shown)


st0012:
@ko1 this is a release blocker.

PR #163, opened 2021-07-12
Enable color display in tests

ono-max:


ono-max:
I will test color feature in other way.

Issue #29, opened 2021-05-25
Introduce something similar to Pry's ls command


st0012:
The `ls` command is one of my favorites when using `Pry`. It gives users a quick overview on the objects with information like:

- Methods & their defined classes
- Constants
- Instance variables
- Locals

<img width="80%" alt="pry ls" src="https://user-images.githubusercontent.com/5079556/119435613-0f320580-bd4d-11eb-935a-1e3904f6273e.png">

### Usage

```
[4] pry(main)> ls -h
Usage: ls [-m|-M|-p|-pM] [-q|-v] [-c|-i] [Object]
       ls [-g] [-l]

ls shows you which methods, constants and variables are accessible to Pry. By
default it shows you the local variables defined in the current shell, and any
public methods or instance variables defined on the current object.

# .....

Also check out `find-method` command (run `help find-method`).

    -m, --methods               Show public methods defined on the Object
    -M, --instance-methods      Show public methods defined in a Module or Class
    -p, --ppp                   Show public, protected (in yellow) and private (in green) methods
    -q, --quiet                 Show only methods defined on object.singleton_class and object.class
    -v, --verbose               Show methods and constants on all super-classes (ignores Pry.config.ls.ceiling)
    -g, --globals               Show global variables, including those builtin to Ruby (in cyan)
    -l, --locals                Show hash of local vars, sorted by descending size
    -c, --constants             Show constants, highlighting classes (in blue), and exceptions (in purple).
                                Constants that are pending autoload? are also shown (in yellow)
    -i, --ivars                 Show instance variables (in blue) and class variables (in bright blue)
    -G, --grep                  Filter output by regular expression
    -d, --dconstants            Show deprecated constants
    -h, --help                  Show this message.
```

#### Add a new command

If we go for this, we should come up with a new name though. Because this project already has the `l[ist]` command for listing source code, naming this command `ls` could be confusing.

#### Enhance the `info` command

The `info` command is designed to show information about the current frame. So it doesn't exactly match what `Pry`'s `ls` does. But if we can add the current functionality to a `info f[rame]` subcommand, `info` can be more generic and perform something like the `ls` does.

hsbt:
FYI: It's already implemented on IRB at https://github.com/ruby/irb/commit/19b6c2060405583f9ac64db1d87f3bedfbe053f5

st0012:
@hsbt wow that looks awesome. I guess the question now becomes: is there a way we can use it from the debug session without entering `irb` 🤔

[...]

PR #172, opened 2021-07-13
Support ls command from irb

st0012:
Implements #29 

截圖 2021-07-13 下午11 15 17

Some notes:

1. `ls` command can't disable coloring because `irb`'s implementation seems to ignore the config. I'll open an issue in the irb repo later.
截圖 2021-07-13 下午11 17 52

2. The current implementation should be able to support other irb commands (like `show_source`) with very few changes. But let's do that in another PR.

### DIFF
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -482,6 +482,10 @@ module DEBUGGER__
         end
 
       when 'ls'
+        if @ui.remote?
+          @ui.puts "not supported on the remote console."
+          return :retry
+        end
         @tc << [:irb, :ls, arg]
 
       # * `edit`
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -481,6 +481,9 @@ module DEBUGGER__
           return :retry
         end
 
+      when 'ls'
+        @tc << [:irb, :ls, arg]
+
       # * `edit`
       #   * Open the current file on the editor (use `EDITOR` environment variable).
       #   * Note that edited file will not be reloaded.
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -481,6 +481,10 @@ module DEBUGGER__
           return :retry
         end
 
+      # * `ls`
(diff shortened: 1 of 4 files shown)


lib/debug/thread_client.rb
```diff
@@ -15,6 +16,7 @@ def self.current
       end
     end
 
+    extend IRB::ExtendCommandBundle
```
st0012, line 19: This module injects irb commands as methods. Pretty handy 👍 
https://github.com/ruby/irb/blob/master/lib/irb/extend-command.rb

lib/debug/session.rb
```diff
@@ -481,6 +481,17 @@ def process_command line
           return :retry
         end
 
+      # * `ls`
+      #   * Executes `irb`'s `ls` command with current binding.
+      # * `ls <obj>`
+      #   * Executes `irb`'s `ls` command with the given obj.
+      when 'ls'
+        if @ui.remote?
+          @ui.puts "not supported on the remote console."
```
st0012, line 490: For some reason it can't run on remote console either. I think we can sort it out later.

lib/debug/thread_client.rb
```diff
@@ -658,5 +680,16 @@ def to_s
...
+
+    def init_irb_context
+      IRB.init_config(nil)
+      workspace = IRB::WorkSpace.new(current_frame.binding)
+      irb = IRB::Irb.new(workspace, nil)
+      IRB.conf[:MAIN_CONTEXT] = irb.context
+    end
+
+    def reset_irb_context
+      IRB.conf[:MAIN_CONTEXT] = nil
```
st0012, line 692: I'm not sure if this is needed, but just in case.

ko1:
Thanks, but I can't accept this PR because it introduced IRB's feature and I can't manage them.
`ThreadClient` should be enough small for maintenance and performance.

Can we copy the implementation from irb?

st0012:
ok.

st0012:
@ko1 another implementation: #173

PR #178, opened 2021-07-15
Remove manual testing section from contributing guideline

st0012:
Thanks to @ono-max's great work, we now have a convenient test framework to write and run tests easily 🙂 
Because of this, the manual testing has become obsolete and we can let it go now.


ko1:
`-e` technique is useful to make a feature (I'm using many times), so I think it is not an obsolete contents.

st0012:
Shouldn’t we mention that in the readme (I thought it’s already there)? Putting it in contributing guide is not very visible to most users.

ko1:
Your example is very descriptive and easy to understand how to use this feature on new feature development. README.md only shows a few lines (and it shouldn't show long examples).

st0012:
I see. I'll just leave it there then 🙂

PR #180, opened 2021-07-15
Implementation of Multi-thread in test framework

ono-max:
These are example of test(debug) time for Github workflow.

### Before
Finished in 379.300202619 seconds.

https://github.com/ruby/debug/runs/3179106902

### After()
Finished in 199.289802343 seconds.

https://github.com/ruby/debug/runs/3179484608

#### - I removed all instance variables from `utils.rb` and `assertions.rb` for thread-safe.
  - The all instance variables are stored as `Struct` class.
  - Queue instance is created for each threads.
#### - `test-unit` doesn't support multi thread (FYI, https://github.com/test-unit/test-unit/issues/204)
  - That's why I don't use `test-unit`'s assert method under multi threads. As the alternative plan, I implemented as follows.
    - If a test fail, the current thread will do nonlocal exits.
    -  If a test succeeds, assert method doesn't do nothing and `assert true` will be run.
  - I am concerned that the number of assert's is not accurate.

### DIFF
--- a/test/support/utils.rb
+++ b/test/support/utils.rb
@@ -147,7 +147,7 @@ module DEBUGGER__
           check_error(/DEBUGGEE Exception/)
           assert_empty_queue(exception: e)
         rescue Timeout::Error => e
-          assert false, create_message("TIMEOUT ERROR (#{TIMEOUT_SEC} sec)")
+          assert_block(create_message("TIMEOUT ERROR (#{TIMEOUT_SEC} sec)")) { false }
         ensure
           kill_remote_debuggee
           # kill debug console process
@@ -163,7 +163,7 @@ module DEBUGGER__
 
     def check_error(error)
       if error_index = @last_backlog.index { |l| l.match?(error) }
-        raise create_message("Debugger terminated because of: #{@last_backlog[error_index..-1].join}")
+        assert_block(create_message "Debugger terminated because of: #{@last_backlog[error_index..-1].join}" ) { false }
       end
     end
 
@@ -215,7 +215,7 @@ module DEBUGGER__
         message += "\nAssociated exception: #{exception.class} - #{exception.message}" +
                    exception.backtrace.map{|l| "  #{l}\n"}.join
       end
-      assert_empty @queue, FailureMessage.new { create_message message }
+      assert_block(FailureMessage.new { create_message message }) { @queue.empty? }
     end
 
     def strip_line_num(str)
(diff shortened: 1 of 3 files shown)


test/support/utils.rb
```diff
@@ -32,26 +32,59 @@ def debuggee_backlog
...
+        end
+      else
+        debug_on_local boot_options, TestInfo.new(new_queue)
       end
+      assert true
     end
 
-    ASK_CMD = %w[quit q delete del kill undisplay]
+    def new_thread &block
+      Thread.new do
```
ko1, line 67: If thread terminates with an exception, `t.join` on line 55 will raise an exception.
It should rescue all exceptions and return it as a failure reason.

test/support/utils.rb
```diff
@@ -32,26 +32,59 @@ def debuggee_backlog
...
+        Thread.current[:is_subthread] = true
+        catch(:fail) do
+          block.call
+        end
+      end
+    end
+
+    def multithreading?
+      Thread.current[:is_subthread]
```
ko1, line 76: `Thread` is general terminology and this case, you may want to know it is a worker thread for debugger testing. You should use more specific term.

test/support/utils.rb
```diff
@@ -32,26 +32,59 @@ def debuggee_backlog
...
+    end
+
+    def new_queue
```
ko1, line 79: It is better to use more specific name.

test/support/utils.rb
ono-max: I changed it to `multithreaded_test?`.

test/support/utils.rb
ono-max: https://github.com/ruby/debug/pull/180/files#diff-ee14bf65f14927f7778a246f90c5cec0c581be3e2658c2a8a3726c59db5be3f9R61-R64

test/support/utils.rb
ono-max: https://github.com/ruby/debug/pull/180/files#diff-ee14bf65f14927f7778a246f90c5cec0c581be3e2658c2a8a3726c59db5be3f9R80

test/support/utils.rb
ono-max: https://github.com/ruby/debug/pull/180/files#diff-ee14bf65f14927f7778a246f90c5cec0c581be3e2658c2a8a3726c59db5be3f9R84

ono-max: https://github.com/ruby/debug/pull/180/files#diff-ee14bf65f14927f7778a246f90c5cec0c581be3e2658c2a8a3726c59db5be3f9R71

ko1:
🎉

PR #183, opened 2021-07-18
Add tests for config options

st0012:
- `skip_path`
- `show_src_lines`
- `show_frames`

### DIFF
--- a/test/debug/config_test.rb
+++ b/test/debug/config_test.rb
@@ -67,4 +67,78 @@ module DEBUGGER__
       end
     end
   end
+
+  class ShowSrcLinesTest < TestCase
+    def program
+      <<~RUBY
+      1| p 1
+      2| p 2
+      3| p 3
+      4| p 4
+      5| p 5
+      6| p 6
+      7| p 7
+      8| p 8
+      9| p 9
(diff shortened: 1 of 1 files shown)


test/debug/config_test.rb
```diff
@@ -67,4 +67,142 @@ def test_config_show_frames_set
...
+        end
+        def lib_m2
+          2
+        end
+      RUBY
+    end
+
+    def program(lib_file)
+      <<~RUBY
+     1| DEBUGGER__::CONFIG[:skip_path] = [/library/]
```
st0012, line 159: I imagine this config will more likely to be set with code instead of command.

ko1: Do not modify `CONFIG` hash because they will be frozen for Ractor support. Use `set_config skip_path: [...]`.

ko1: Sorry I will prepare `CONFIG#[]=(key, val)` to change the configuration, so no need to change.

st0012: Thanks for the update. I think this PR is ready for merge if no other changes are needed?

PR #184, opened 2021-07-18
The skip_path config should also be considered in flow control commands

st0012:


ko1:
Now, `skip_path` is designed only for `bt` command so that description says it's "UI" category.
I understand the motivation, if we skip on `step` command, should it be a `Control` category?

st0012:
@ko1 I agree with that 👍 I've updated its category and added test case for the change.

ko1:
`skip_nosrc` is already `CONTROL` option. wow.

st0012:
oh I didn't notice that, good catch 👍 I've updated it too.

PR #192, opened 2021-07-21
new config `keep_alloc_site` to keep allocation site

ko1:
```
(rdbg) config set trace_alloc true
trace_alloc = true             # CONTROL: Keep allocation site and p, pp shows it (default: false)
(rdbg) n
[1, 4] in target.rb
      1| a = Object.new
=>    2| p a
      3|
      4| __END__
=>#0    <main> at target.rb:2
(rdbg) p a
=> #<Object:0x00005564a2b4ebd0>
allocated at target.rb:1
```

### DIFF
--- a/lib/debug/config.rb
+++ b/lib/debug/config.rb
@@ -12,8 +12,9 @@ module DEBUGGER__
     no_reline:      ['RUBY_DEBUG_NO_RELINE',      "UI: Do not use Reline library (default: false)",                 :bool],
 
     # control setting
-    skip_path:      ['RUBY_DEBUG_SKIP_PATH',      "CONTROL: Skip showing/entering frames for given paths (default: [])",          :path],
+    skip_path:      ['RUBY_DEBUG_SKIP_PATH',      "CONTROL: Skip showing/entering frames for given paths (default: [])", :path],
     skip_nosrc:     ['RUBY_DEBUG_SKIP_NOSRC',     "CONTROL: Skip on no source code lines (default: false)",              :bool],
+    keep_alloc_site:['RUBY_DEBUG_KEEP_ALLOC_SITE',"CONTROL: Keep allocation site and p, pp shows it (default: false)",   :bool],
 
     # boot setting
     nonstop:        ['RUBY_DEBUG_NONSTOP',     "BOOT: Nonstop mode",                                                :bool],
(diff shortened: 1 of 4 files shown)


ko1:
Discussion is, should it be a `trace` sub command or configuration?

st0012:
I prefer a `trace` sub command because that's the first place I'll look for tracing related features.

ko1:
Yes, but it doesn't print lines at allocation.
For example, line/call/passing events are printed (or recorded in a memory or a file) at each event.
By analogy of these examples, `trace allocation` will print each allocations (it can be useful on specific class, anyway).

But this feature only affects on `p` or `pp` command.

ko1:
This is why I chose configuration.

ko1:
`keep_alloc_site` is another name option.

st0012:
The original `trace on|off` doesn't print things itself either (I mean when typing the command). 
I think it's fine to let `trace allocate` (the sub command name I'd imagine) only enable functionalities instead of printing things, as long as other sub commands behave the same way too.

(btw maybe `track` (implies following) is better than `trace` (implies backward) in this case?)

st0012:
Another argument could be: configurations should be more static than command toggling. And in this case, tracing allocation is more likely to be enabled/disabled differently across debugging sessions instead of consistently. For example,

- Debug attempt 1 - not enabling allocation tracing.
- Debug attempt 2 - enable allocation tracing.
- Debug attempt 3 - not enabling allocation tracing.

So I think it's not a "configuration" like `skip_path` that should be the same across multiple sessions. It's like "do I need this enabled now?", which is more suitable with commands.

ko1:
> The original trace on|off doesn't print things itself either (I mean when typing the command).
> I think it's fine to let trace allocate (the sub command name I'd imagine) only enable functionalities instead of printing things, as long as other sub commands behave the same way too.

I couldn't understand this opinion. Original `trace on` will prints line events.

st0012:
I meant technically `trace on` doesn't print the information. It activates a `TracePoint` that prints things when the program proceeds. So to me the `trace` command is more like a `switch` to enable certain mechanisms.

In this case, allocation tracing is a mechanism to be activated too. At the moment, I don't see any config option that also performs this kind of activation. That's why I think it fits the same `trace` pattern better than a config option.

ko1:
I decided to remain it as configuration.

* `trace` terminology is known as "trace the behavior and print/save the history", because `tracer` library of Ruby, or `strace` command. Proposed feature of this PR doesn't care about history. I think `trace` command doesn't mean `TracePoint`.
* We can enable this feature by an environment variable, if it is an option in `config`.
* I'm checking gdb's `set` command (similar to `config`) and it has `set record` which enable the feature.
* Alternative command would be `enable` command (`enable save-allocation-site` for example) but now it doesn't have other sub-command candidates.

ko1:
I can introduce another command if there are this kind of features.

PR #197, opened 2021-07-25
Add skip_raise_path config option and rename skip_path to skip_frame_path

st0012:
As suggested in https://github.com/ruby/debug/pull/157#issuecomment-878383710

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,7 +360,7 @@ config set no_color true
   * `RUBY_DEBUG_NO_RELINE` (`no_reline`): Do not use Reline library (default: false)
 
 * CONTROL
-  * `RUBY_DEBUG_SKIP_PATH` (`skip_path`): Skip showing/entering frames for given paths (default: [])
+  * `RUBY_DEBUG_SKIP_FRAME_PATH` (`skip_frame_path`): Skip showing/entering frames for given paths (default: [])
   * `RUBY_DEBUG_SKIP_RAISE_PATH` (`skip_raise_path`): Skip catching exceptions raised from given paths (default: [])
   * `RUBY_DEBUG_SKIP_NOSRC` (`skip_nosrc`): Skip on no source code lines (default: false)
   * `RUBY_DEBUG_KEEP_ALLOC_SITE` (`keep_alloc_site`): Keep allocation site and p, pp shows it (default: false)
--- a/README.md
+++ b/README.md
@@ -361,6 +361,7 @@ config set no_color true
 
 * CONTROL
   * `RUBY_DEBUG_SKIP_PATH` (`skip_path`): Skip showing/entering frames for given paths (default: [])
+  * `RUBY_DEBUG_SKIP_RAISE_PATH` (`skip_raise_path`): Skip catching exceptions raised from given paths (default: [])
   * `RUBY_DEBUG_SKIP_NOSRC` (`skip_nosrc`): Skip on no source code lines (default: false)
   * `RUBY_DEBUG_KEEP_ALLOC_SITE` (`keep_alloc_site`): Keep allocation site and p, pp shows it (default: false)
   * `RUBY_DEBUG_POSTMORTEM` (`postmortem`): Enable postmortem debug (default: false)
--- a/lib/debug/breakpoint.rb
+++ b/lib/debug/breakpoint.rb
@@ -247,7 +247,9 @@ module DEBUGGER__
     def setup
       @tp = TracePoint.new(:raise){|tp|
         exc = tp.raised_exception
+
         next if SystemExit === exc
+        next if CONFIG[:skip_raise_path] && CONFIG[:skip_raise_path].any? { |p| tp.path.match?(p) }
         should_suspend = false
 
         exc.class.ancestors.each{|cls|
--- a/lib/debug/config.rb
+++ b/lib/debug/config.rb
@@ -12,7 +12,7 @@ module DEBUGGER__
     no_reline:      ['RUBY_DEBUG_NO_RELINE',      "UI: Do not use Reline library (default: false)",                 :bool],
 
     # control setting
-    skip_path:       ['RUBY_DEBUG_SKIP_PATH',       "CONTROL: Skip showing/entering frames for given paths (default: [])",     :path],
+    skip_frame_path: ['RUBY_DEBUG_SKIP_FRAME_PATH', "CONTROL: Skip showing/entering frames for given paths (default: [])",     :path],
     skip_raise_path: ['RUBY_DEBUG_SKIP_RAISE_PATH', "CONTROL: Skip catching exceptions raised from given paths (default: [])", :path],
     skip_nosrc:      ['RUBY_DEBUG_SKIP_NOSRC',      "CONTROL: Skip on no source code lines (default: false)",              :bool],
     keep_alloc_site: ['RUBY_DEBUG_KEEP_ALLOC_SITE',"CONTROL: Keep allocation site and p, pp shows it (default: false)",   :bool],
--- a/lib/debug/config.rb
+++ b/lib/debug/config.rb
@@ -12,10 +12,11 @@ module DEBUGGER__
     no_reline:      ['RUBY_DEBUG_NO_RELINE',      "UI: Do not use Reline library (default: false)",                 :bool],
 
     # control setting
-    skip_path:      ['RUBY_DEBUG_SKIP_PATH',      "CONTROL: Skip showing/entering frames for given paths (default: [])", :path],
-    skip_nosrc:     ['RUBY_DEBUG_SKIP_NOSRC',     "CONTROL: Skip on no source code lines (default: false)",              :bool],
-    keep_alloc_site:['RUBY_DEBUG_KEEP_ALLOC_SITE',"CONTROL: Keep allocation site and p, pp shows it (default: false)",   :bool],
-    postmortem:     ['RUBY_DEBUG_POSTMORTEM',     "CONTROL: Enable postmortem debug (default: false)",                   :bool],
+    skip_path:       ['RUBY_DEBUG_SKIP_PATH',       "CONTROL: Skip showing/entering frames for given paths (default: [])",     :path],
+    skip_raise_path: ['RUBY_DEBUG_SKIP_RAISE_PATH', "CONTROL: Skip catching exceptions raised from given paths (default: [])", :path],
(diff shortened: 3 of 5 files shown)


test/support/utils.rb
```diff
@@ -45,6 +43,8 @@ def debug_code(program, boot_options: '-r debug/start', remote: true, &block)
 
       debug_on_local boot_options
 
+      check_line_num!(program)
```
st0012, line 46: When writing a new test case, verifying the test case is valid is more important than adding the line numbers.
So the framework should wait for the test to pass before asking for line numbers.

st0012:
After this is merged, I'll add a section to readme for suggesting an example setup for Rails applications with `skip_frame_path` and `skip_raise_path`.

ko1:
After thinking about that, I'm confusing with these two configurations.

Maybe there are several types.

* 1. show frame
* 2. stop with step/next/finish
* 3. stop with catch
* 4. stop with watch <- maybe watch should not skip.
* 5. trace https://github.com/ruby/debug/pull/198
* others?

Now, 1 and 2 with `skip_path`. I commented before separation is needed, but is it needed?
Do you want to skip exception with path name?

At least, `skip_frame_path` affects 1 and 2 is not good idea if we introduce `skip_raise_path`.

st0012:
@ko1 yes, I want to skip exception with path name, for example: `/active_support`.

Here's the log when I catch `StandardError` in my app's controller:

```
#0  BP - Catch  "StandardError"
(rdbg) c
[589, 598] in ~/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/activesupport-6.0.4/lib/active_support/dependencies.rb
    589|         end
    590|       end
    591|
    592|       name_error = NameError.new("uninitialized constant #{qualified_name}", const_name)
    593|       name_error.set_backtrace(caller.reject { |l| l.starts_with? __FILE__ })
=>  594|       raise name_error
    595|     end
    596|
    597|     # Remove the constants that have been autoloaded, and those that have been
    598|     # marked for unloading. Before each constant is removed a callback is sent
=>#0    ActiveSupport::Dependencies#load_missing_constant(from_mod=Account(id: integer, name: string, count..., const_name=:Member) at ~/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/activesupport-6.0.4/lib/active_support/dependencies.rb:594
  #1    block in load_missing_constant at ~/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/bootsnap-1.4.7/lib/bootsnap/load_path_cache/core_ext/active_support.rb:61
  # and 198 frames (use `bt' command for all frames)

Stop by #0  BP - Catch  "StandardError"
(rdbg)
```

Rails apps in development usually lazily load classes by `ActiveSupport`. And because it uses exceptions as a flow control approach (which is a common usage in many libs), it constantly rescues and re-raises exceptions and trigger the breakpoint outside the application code.

ko1:
specifying `catch` exception class specification is not enough?

st0012:
I'm not sure if this is a general pattern, but my application has:

```ruby
class ApplicationController < ActionController::Base
    rescue_from ActiveRecord::RecordNotUnique, with: :render_error_response
    rescue_from ActiveRecord::RecordInvalid, with: :render_error_response
    rescue_from CustomExceptionA, with: :render_error_response
    rescue_from CustomExceptionB, with: :render_something_else
    # other rescue_from
end
```

Without the debugger, to debug an exception response I'll need to:

1. Make a request and inspect the exception info from the response.
2. Go to the file that raises the exception. (sometimes it's a library, which requires `bundle open gem_name`)
3. Add `binding.pry` before the exception.
4. Make another request.
5. Start debugging.

I'd hope the `catch` command to help me reduce it to:

1. Visit the controller action I'm debugging.
2. Add `binding.b(do: "catch StandardError")`
3. Make a request.
4. Start debugging.

Ideally, in the latter case I don't need to know much about exception nor its origin to place a breakpoint. Which I think it's a big advantage of using `catch`. But currently it's not viable if it constantly stops at Rails' internal exceptions.

ko1:
sorry what is `rescue_from`?

ko1:
In this case, you know where the exception from? (If no, you can't specify `skip_raise_path`)
For example, if the exceptions are from `app` dir, `catch Exception from: app/` helps?

ko1:
or gems can raises an exception, but you want to add ignore lists if it is internal exceptions you don't have interest incrementally?

st0012:
> sorry what is rescue_from?

it's a common helper in Rails controller: https://edgeapi.rubyonrails.org/classes/ActiveSupport/Rescuable/ClassMethods.html

in short: it rescues the specified exception raised from the controller's actions and call the given `with:` method. the `with:` method usually renders a universal error page as the response.

> In this case, you know where the exception from? (If no, you can't specify skip_raise_path)

I don't know where the exception is from, but I generally know where it's not from. That's why I want to specify a skip path.

let me demonstrate my usage this way: I'd put this in my application as the default debugger config:

```ruby
# config/initializers/debug.rb
# this will be evaluated when ever the Rails application is initialized
framework_patterns = [/active_support/, /action_pack/, ...] # all Rails components
DEBUGGER__::CONFIG[:skip_raise_path] = framework_patterns
DEBUGGER__::CONFIG[:skip_frame_path] = framework_patterns + [/lib/]
```

ko1:
> in short: it rescues the specified exception raised from the controller's actions and call the given with: method. the with: method usually renders a universal error page as the response.

Thanks!

> let me demonstrate my usage this way: I'd put this in my application as the default debugger config:

In this case, if your interest exception is from `action_pack`, you will miss the exception. Is it okay?

ah, 

> but I generally know where it's not from.

You wrote it. So you don't know which gem raises, but you know `action_pack` (for example) doesn't raise an error.

st0012:
> You wrote it. So you don't know which gem raises, but you know action_pack (for example) doesn't raise an error.

yes. in most cases (99%), the exceptions are not caused by Rails itself, so I'm happy to skip its exceptions in most debug sessions.

st0012:
@ko1 I imagine in some cases the separation would be helpful. But in general they should be similar to each other. So I think it's ok to do the separation later when we have a specific case.

st0012:
@ko1 should I merge the `skip_raise_path` config into `skip_path`?

ko1:
Sorry for late.
Now I think separation seems better. So I want to accept `skip_raise_path`.
Now, `skip_frame_path` is not good name, I think.

For (1) in https://github.com/ruby/debug/pull/197#issuecomment-887352661 `skip_frame_path` is better name, but not sure it is for (2). Remaining `skip_path` is one idea.

st0012:
@ko1 If we'll have more than one `skip_*_path`, I think ambiguous `skip_path` should be avoided.

> For (1) in #197 (comment) skip_frame_path is better name, but not sure it is for (2).

Personally, I understand that commands like `step` or `next` are frame-related operations. So using `skip_frame_path` for them is understandable to me.

I think users that know these commands should also be familiar with the concept of frame too. For example, `byebug` describes the [`step` command](https://github.com/deivid-rodriguez/byebug/blob/master/lib/byebug/commands/step.rb#L10-L11) like this:

```
# double quotes are added by me

Allows the user the continue execution until the next instruction, possibly
in a different "frame". Use step to step into method calls or blocks.
```

If that's the case, then `skip_frame_path` affecting the `step` command should be clear to them as well.

ko1:
"next" and "finish" is related to the frame, but "step" isn't. If it is related to "frame", most of operations are related to the frame.

If I name to (2), `skip_stop_path` or `skip_step_path`.
We can separate (1) and (2) from `skip_path`, but maybe (1) and (2) are used at once, so how about to name it `skip_path`? maybe (5) trace should be `skip_trace_path`, but maybe `skip_path` can skip tracing with same reason.

* (1) `skip_frame_path`
* (2) `skip_step_path`
* (3) `skip_raise_path`
* (4) N/A
* (5) `skip_trace_path`

but (1), (2), (5) will be used same time so the result:

* (1) `skip_path`
* (2) `skip_path`
* (3) `skip_raise_path`
* (4) N/A
* (5) `skip_path`

In other words, `skip_path` declares that they should be ignored files.
If we have a request to specify (1, 2, 5) respectively, we can introduce them.

st0012:
I'm ok with that. But I want to propose a slight change on some of the names:

1. show frame - `skip_frame_path`
2. stop with step/next/finish - `skip_control_path`
3. stop with catch - `skip_catch_path`
4. stop with watch - `skip_watch_path` (not introducing now)
5. trace - `skip_trace_path` (not introducing now)

Also, in the long-term, there could be a hierarchy for the new settings instead of a replacement:
By default, `skip_path` applies to all of them. But when a more specific config is introduced and set, it overrides `skip_path` on the functionality.

So for now, we can introduce `skip_path` and let it control 1~5. But if the user wants to configure `catch`'s behavior, he/she can also set `skip_catch_path`.

And to keep implementation simple, sub-configs like `skip_catch_path` don't inherit `skip_path`'s value.

ko1:
> By default, skip_path applies to all of them.

My assumption is, (1, 2, 5) is similar cases and (3) is special case (this is why I suggests to separate it).
This assumption can be wrong (people want to ignore with 3).

Maybe nobody want to ignore (4).

> But when a more specific config is introduced and set, it overrides skip_path on the functionality.

This logic is complicated for me.
To simplify, we can introduce separate for each operation (maybe people use ruby code for them...)
Hmm...

BTW "control" for (2) is ambiguous. In most of debugger, this operation is called "step execution".

ko1:
another idea is to allow specify skip type for each path like:
`/foo/bar:[frame]/baz/boo` but it's also complicated...

st0012:
> My assumption is, (1, 2, 5) is similar cases and (3) is special case (this is why I suggests to separate it).

My opinions are:

- 1 and 2 should be ignored together to avoid confusion.
- 3 would be special as you said.
- Ignoring 4 will be necessary for Rails apps. For example, I generally don't care about how Rails itself change my records' state, so I'd ignore all Rails components' path here. I assume the value would be similar or even the same as 1 and 2's.
- I won't ignore 5 with a global configuration for now. When I use my gem [object_tracer](https://github.com/st0012/object_tracer), which is similar to the tracer feature, I generally use it to learn the behavior that I'm not aware of. So I'd like to see every activity in the app. And if it gets noisy because of certain parts of the app, I use per-tracer option like `filter_by_paths` for filtering.

> BTW "control" for (2) is ambiguous. In most of debugger, this operation is called "step execution".

I see. Then `skip_step_path` sounds good to me 👍

ko1:
after consideration, I decide to use `skip_path` for skipping catch breakpoint (so I re-implemented #216).
When there are requests to separate them, revisit it.

PR #199, opened 2021-07-28
Document and reorganize ThreadClient

st0012:
- Similar methods are now closer together, for example `#to_s` is now next to `#inspect`.
- Command-specific methods are grouped by the command.
- Comments are added for different sections.

### DIFF
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -410,6 +410,7 @@ module DEBUGGER__
       }
     end
 
+    # this method is extracted to hide frame_eval's local variables from C method eval's binding
     def instance_eval_for_cmethod frame_self, src
       frame_self.instance_eval(src)
     end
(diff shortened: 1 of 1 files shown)


ko1:
Maybe it is not an command (confusing with debug command), but okay (I'll change when I find another term)

st0012:
I took the `cmd` term from 

https://github.com/ruby/debug/blob/5e7cba4eead1265f3bbbf0a5d45e25d67b150121/lib/debug/thread_client.rb#L580

It's hard to use find a proper term because the `cmd` can be either verb (`show` or `eval`) or noun (`frame` or `breakpoint`). If we can convert nouns to verbs (`frame -> switch_frame`) we can name it `action`? 

Or we can rename `cmd`  as `category`, which I think fits here but is a little too generalized.

ko1:
ah, I see. sorry. It means message between thread and debugger (session) and message contains a command.

st0012:
So there are:
- User commands
- Session to Thread commands

Perhaps we can name the latter as Session to Thread `instructions`?

st0012:
I added https://github.com/ruby/debug/pull/200 as a demonstration.

PR #200, opened 2021-07-28
Rename ThreadClient's cmd as instruction

st0012:
Currently, both

- User commands
- Session to thread operations

uses the term `cmd`, which could cause confusion when reading the code.

By renaming "session to thread operation" as "instruction", we can avoid the confusion.


st0012:
@ko1 yeah I understand that confusion so I opened this as a draft to demo the change with code. you can close it anytime if you think the name is improper 🙂

PR #204, opened 2021-07-29
Add tests for 82ac03e

st0012:


st0012:
@ko1 I think your implementation works great. thanks 👍

PR #209, opened 2021-07-30
Fix anonymous rest argument tracing

st0012:
Methods with `...` parameter doesn't read any argument, so the tracer should just ignore it.

Currently the tracer doesn't handle this case:

```
❯ ruby -Ilib -r debug target.rb
DEBUGGER: Session start (pid: 54807)
[1, 7] in target.rb
=>    1| binding.b(do: "trace pass 1")
      2|
      3| def foo(...)
      4|   block.call
      5| end
      6|
      7| foo(1)
=>#0    <main> at target.rb:1
(rdbg:binding.break) trace pass 1
Enable PassTracer for 1 (enabled)
        5: from target.rb:7:in `<main>'
        4: from target.rb:4:in `foo'
        3: from /Users/st0012/projects/debug/lib/debug/tracer.rb:142:in `block in setup'
        2: from /Users/st0012/projects/debug/lib/debug/tracer.rb:142:in `each'
        1: from /Users/st0012/projects/debug/lib/debug/tracer.rb:153:in `block (2 levels) in setup'
/Users/st0012/projects/debug/lib/debug/tracer.rb:153:in `local_variable_get': wrong local variable name `*' for #<Binding:0x00007ffb7da840f0> (NameError)
```

This makes tracer unusable in Rails apps because it uses `...` in method delegation 

https://github.com/rails/rails/blob/7e8946c4da8b69e3b33a3f16cdfd8923c8756000/activesupport/lib/active_support/core_ext/module/delegation.rb#L198-L199

### DIFF
--- a/lib/debug/tracer.rb
+++ b/lib/debug/tracer.rb
@@ -174,7 +174,6 @@ module DEBUGGER__
               ary = b.local_variable_get(name)
               ary.each{|e|
                 if e.object_id == @obj_id
-                  m = "#{tp.defined_class}\##{tp.method_id}"
                   out tp, " `#{@obj_inspect}` is used as a parameter in `#{name}` of #{minfo(tp)}"
                 end
               }
--- a/lib/debug/tracer.rb
+++ b/lib/debug/tracer.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 
 module DEBUGGER__
   class Tracer
@@ -168,7 +169,8 @@ module DEBUGGER__
               end
 
             when :rest
-              next unless name
+              next unless name && name.to_s != "*"
+
               ary = b.local_variable_get(name)
               ary.each{|e|
                 if e.object_id == @obj_id
--- a/test/debug/trace_test.rb
+++ b/test/debug/trace_test.rb
@@ -104,17 +104,29 @@ module DEBUGGER__
 
   class TracePassTest < TestCase
     def program
-      <<~RUBY
-     1| def foo(...); end
-     2| def bar(a:); end
-     3| def baz(**kw); end
-     4|
-     5| foo(1)
-     6| bar(a: 2)
-     7| baz(b: 3)
-     8|
-     9| binding.b
-      RUBY
+      if RUBY_VERSION >= "2.7"
+        <<~RUBY
+       1| def foo(...); end
+       2| def bar(a:); end
+       3| def baz(**kw); end
+       4|
+       5| foo(1)
+       6| bar(a: 2)
+       7| baz(b: 3)
+       8|
+       9| binding.b
+        RUBY
+      else
+        <<~RUBY
+       1| def bar(a:); end
+       2| def baz(**kw); end
+       3|
+       4| bar(a: 2)
+       5| baz(b: 3)
+       6|
+       7| binding.b
+        RUBY
+      end
     end
 
     def test_not_tracing_anonymous_rest_argument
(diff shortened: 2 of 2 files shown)


test/support/utils.rb
```diff
@@ -38,8 +38,6 @@ def debuggee_backlog test_info
 
     # This method will execute both local and remote mode by default.
     def debug_code(program, boot_options: '-r debug/start', remote: true, &block)
-      check_line_num!(program)
```
ono-max, line 41: Why did you move this one?

ono-max (changes requested):

test/support/utils.rb
```diff
@@ -71,7 +69,8 @@ def debug_code(program, boot_options: '-r debug/start', remote: true, &block)
       else
         debug_on_local boot_options, TestInfo.new(dup_scenario)
       end
-      assert true
```
ono-max, line 74: I think this change will confuse users. If they see the following results, they don't know if test works well.

```shell
$ ruby test/debug/quit_test.rb
Loaded suite test/debug/quit_test
Started
...
Finished in 1.047331 seconds.
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
3 tests, 0 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
2.86 tests/s, 0.00 assertions/s
```

st0012: Ah I thought this is just a redundant line. But why aren't the previous assertion calls counted?

test/support/utils.rb
st0012: I think `test/assertion_helper_test.rb` and `test/test_utils_test.rb` are not the main use case for line numbers?
Anyway, I'm not against checking line numbers on them either. But currently I can't adjust my test script easily before they pass because of line numbers. And that's quite frustrating 🙁

test/support/utils.rb
ono-max: https://github.com/ruby/debug/pull/180#issue-690609876 should help you.

st0012: I see it makes sense now. Great work reducing the testing time btw 👍

st0012: I've removed this change.

test/support/utils.rb
st0012: I've removed this change as well. But can you find a way to let line checking happen after tests? Otherwise I need to make this change locally every time to I can finish my test script first.

ono-max: Thanks!
`test builder` is fastest way to solve the problem, but you don't wanna use it, right?

ono-max (approved):

test/support/utils.rb
st0012: IMO, test builder should not be a replacement of writing tests in the old fashion way 🙂 
Writing test script in a separate file, run cli commands, and go back to test file to finish the test is just too much switching. 

And even if I do that, I still can't easily modify the test script because it has line numbers. (I know in this case I should update the original file again, but that's confusing).

ko1:
could you rebase it?

st0012:
rebased.

lib/debug/tracer.rb
```diff
@@ -168,11 +169,11 @@ def setup
               end
 
             when :rest
-              next unless name
+              next unless name && name.to_s != "*"
```
ko1, line 172: FYI: You only need to use `:'*'`

st0012: ah, right 🤦 
I've opened https://github.com/ruby/debug/pull/212 for this.

PR #210, opened 2021-08-01
Make sure to enable `check_line_num` only when Github Workflow.

ono-max:
As https://github.com/ruby/debug/pull/209#discussion_r680291109, we need the option for disabling `check_line_num!` method. I added the environment variable in `utils.rb` and `ruby.yml` to make `check_line_num!` valid only when Github Workflow.

Of course, we can also enable it in local environment by adding environment variable such as
```shell
$ RUBY_DEBUG_TEST_CHECK_LINE_NUM=true rake test
```

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -32,3 +32,5 @@ jobs:
         bundler-cache: true # runs 'bundle install' and caches installed gems automatically
     - name: Run tests
       run: bundle exec rake
+      env:
+        RUBY_DEBUG_TEST_CHECK_LINE_NUM: true
--- a/test/support/utils.rb
+++ b/test/support/utils.rb
@@ -38,7 +38,7 @@ module DEBUGGER__
 
     # This method will execute both local and remote mode by default.
     def debug_code(program, boot_options: '-r debug/start', remote: true, &block)
-      check_line_num!(program)
+      check_line_num!(program) if ENV['RUBY_DEBUG_TEST_CHECK_LINE_NUM']
 
       write_temp_file(strip_line_num(program))
       @scenario = []


ono-max:
@st0012 

I hope this PR should solve your problem :)

st0012:
@ono-max thank you for the PR!
but this doesn't solve my problem. I still want the line number check to run because I rely on it to generate line-numbered program script. 
I'll open a PR to propose another change 🙂

ono-max:
@st0012 

> but this doesn't solve my problem. I still want the line number check to run because I rely on it to generate line-numbered program script.

It doesn't make sense to me. You can run `check_line_num!` method as follows.
```shell
$ RUBY_DEBUG_TEST_CHECK_LINE_NUM=true rake test
```

st0012:
I'd like to avoid running tests with another env var when possible. The behavior I'm looking for is simple:

1. Verify test result
2. If 1 passed, verify line numbers

With the correct verification order, it's done automatically by the test framework (like it used to do). 

But now it only does `1` locally and I have to perform `2` manually.

I added https://github.com/ruby/debug/pull/211 to demonstrate the change.

ono-max:
Thank you for giving me much detailed information. There are some stuff I would like to confirm with you.

- The reason why you don't wanna avoid env var is not automatic?(You'd like to check without anything)
- When users created a PR, they can check the log on Github Workflow and they can copy and paste line number from it.
  - Doesn't this solve your problem?

st0012:
> The reason why you don't wanna avoid env var is not automatic?(You'd like to check without anything)

Yes, it should be automatic because it "was" automatic.

> When users created a PR, they can check the log on Github Workflow and they can copy and paste line number from it.

Don't you think it's weird that users have to rely on CI to finish their test case? Especially for something that could've been done locally.

ono-max:
Ok, I close my PR.

PR #211, opened 2021-08-02
Only check line numbers after the test is finished

st0012:
Users should be able to freely change their test script without the interruption of the line numbers until the script is confirmed working.

Only then should the test framework check the script's line numbers with a suggested change (if it doesn't have them yet).

This change will disable line number check on the test framework's tests because most of those cases are designed to fail. But the number check wasn't design for those tests anyway.

### DIFF
--- a/test/assertion_helper_test.rb
+++ b/test/assertion_helper_test.rb
@@ -6,7 +6,7 @@ module DEBUGGER__
   class AssertLineTextTest < TestCase
     def program
       <<~RUBY
-        1| a = 1
+        a = 1
       RUBY
     end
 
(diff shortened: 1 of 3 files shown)


st0012:
@ono-max @ko1 can we merged this too? I think the small drawback of this change to bring back the "old" checking order is reasonable. 
It's bothering me for a while as I started proposing such change since https://github.com/ruby/debug/pull/197

ono-max:
Yes we can. I agree with your opinion.

ono-max (approved):

st0012:
@ono-max thank you 😄

PR #213, opened 2021-08-03
Truncate trace message

st0012:
Printing full inspect result would make the tracing unusable in Rails apps because there are many big objects. 

If we want to let user know about the trace's object, a truncated result should do the work too.

**Before**
截圖 2021-08-03 下午12 43 47

**After**

截圖 2021-08-03 下午12 44 07

### DIFF
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -438,7 +438,7 @@ module DEBUGGER__
       w = SESSION::width
 
       if mono_info.length >= w
-        info = mono_info[0 .. (w-4)] + '...'
+        info = truncate(mono_info, width: w)
       else
         valstr = colored_inspect(obj, width: 2 ** 30)
         valstr = inspected if valstr.lines.size > 1
@@ -446,6 +446,10 @@ module DEBUGGER__
       end
 
       puts info
+    end
+
+    def truncate(string, width:)
+      string[0 .. (width-4)] + '...'
     end
 
     ### cmd: show edit
@@ -771,7 +775,15 @@ module DEBUGGER__
             begin
               obj = frame_eval args.shift, re_raise: true
               opt = args.shift
-              event! :result, :trace_pass, obj.object_id, obj.inspect, opt
+              obj_inspect = obj.inspect
+
+              width = 50
+
+              if obj_inspect.length >= width
+                obj_inspect = truncate(obj_inspect, width: width)
+              end
+
+              event! :result, :trace_pass, obj.object_id, obj_inspect, opt
             rescue => e
               puts e.message
               event! :result, nil


lib/debug/tracer.rb
```diff
@@ -156,7 +156,7 @@ def setup
         next if skip?(tp)
 
         if tp.self.object_id == @obj_id
-          out tp, "`#{@obj_inspect}` is used as a receiver of #{minfo(tp)}"
+          out tp, "`#{@obj_inspect}` receives #{minfo(tp)}"
```
ko1, line 159: "obj receives C#foo" doesn't make sense.
"obj receives the message #foo" does make sense. But "C" is gone.

ko1: ideas:
"C#foo is called with obj as a receiver"
"C#foo is called with obj as a parameter"

st0012: > C#foo is called with obj as a receiver 

The subject of this sentence is `C#foo` instead of `obj`. But in the context of `trace pass`, I think the `obj` should be highlighted as the subject like the current messages do.

Also, I might be wrong here, but from my understanding `C#foo is called with obj as a receiver` is not grammatically correct either.

`C#foo is called with obj` implies `obj` is an argument passed to it, not the receiver of this call. An example found in `python`'s [unittest.mock document](https://docs.python.org/3/library/unittest.mock.html):

>  The function is called with the same arguments as the mock, and unless it returns DEFAULT

And `as a receiver` still refers to the subject of this sentence, which is the `C#foo`, not the `obj`. 

I also don't think the modified version "not making sense". In general, `being a receiver of something` == `receives something`. For example:

- I'm the receiver of a letter from my friend.
- I received a letter from my friend.

So I think this change still expresses the similar idea.

st0012: I also don't think we need to pursue 100% grammatical correctness in these messages. As long as it's not confusing with one another, keeping the message short should be prioritized.

ko1: another idea:

"obj.foo (C#foo)"
"obj.foo" (for singleton method)

st0012: how about

- `obj receives #foo (C#foo)` for `C#foo`
- `C receives .bar (C.bar)` for `C.bar`
- `obj receives .singleton_foo` for singleton method

ko1: "obj.foo (C#foo)" doesn't enough?

st0012: because we use `inspect`'s result, `obj` doesn't always look like an object. for example:
```
`#<ActiveRecord::Relation [#<Post id: 1, title: ...`
```

in such condition, that format isn't easy to read

```
`#<ActiveRecord::Relation [#<Post id: 1, title: ...`#where (ActiveRecord::Relation#where)`
```

so I think this is better

```
`#<ActiveRecord::Relation [#<Post id: 1, title: ...` receives #where (ActiveRecord::Relation#where)`
```

st0012: I've opened another PR https://github.com/ruby/debug/pull/214 for this conversation and dropped this change because it's not the goal of this PR 😅

PR #219, opened 2021-08-08
Update README

Juanmcuello:
- Fix typos.
- Remove trailing whitespaces.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,9 +43,9 @@ To use a debugger, roughly you will do the following steps:
 2. Run a program with the debugger.
 3. At the breakpoint, enter the debugger console.
 4. Use debug commands.
-    * Query the prgram status (e.g. `p lvar` to see the local variable `lvar`).
+    * Query the program status (e.g. `p lvar` to see the local variable `lvar`).
     * Control program flow (e.g. move to the another line with `step`, to the next line with `next`).
-    * Set another breakpoints (e.g. `catch Exception` to set the breakpoints when `Exception` is raiesd).
+    * Set another breakpoint (e.g. `catch Exception` to set a breakpoint when `Exception` is raised).
     * Change the configuration (e.g. `config set no_color true` to disable coloring).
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -649,7 +649,7 @@ def foo
 end
 ```
 
-On this case, you can see the result of `bar()` everytime when you stops there.
+On this case, you can see the result of `bar()` every time when you stops there.
```
olleolleolle, line 652:
```suggestion
On this case, you can see the result of `bar()` every time you stop there.
```

How about this?

Juanmcuello: Good catch, I missed that. Fixed.

misc/README.md.erb
```diff
@@ -411,7 +411,7 @@ Emacs support available.
 
 #### Start by method
 
-After loading `debug/session`, you can start debug session with the following methods. They are convinient if you want to specifies debug configrations in your program.
+After loading `debug/session`, you can start debug session with the following methods. They are convenient if you want to specifies debug configurations in your program.
```
olleolleolle, line 414:
```suggestion
After loading `debug/session`, you can start debug session with the following methods. They are convenient if you want to specify debug configurations in your program.
```

misc/README.md.erb
```diff
@@ -355,7 +355,7 @@ config set no_color true
 
 ### Initial scripts
 
-If there is `~/.rdbgrc`, the file is loaded as an initial scripts which contains debug commands) when the debug session is started. 
+If there is `~/.rdbgrc`, the file is loaded as an initial scripts which contains debug commands) when the debug session is started.
```
olleolleolle, line 358:
```suggestion
If there is `~/.rdbgrc`, the file is loaded as an initial script (which contains debug commands) when the debug session is started.
```

olleolleolle (approved):
This is a good upgrade of the README's readability.

Thanks, @Juanmcuello!

ko1:
Thank you!

PR #220, opened 2021-08-14
Add tests for 4aca49

st0012:


st0012:
The tests on TCP remote mode seems to become unstable lately. I see `TIMEOUT ERROR (10 sec) on TCP/IP mode` happens in different test cases with different Ruby versions.

ono-max:
Can you rebase this one? It seems that you are not latest version.

ono-max (approved):
Thanks 👍

st0012:
@ono-max one small suggestion: don't delete comments even if it's mistaken. it's confusing when seeing the email notification but can't find the comment. you can simply edit/add another comment to correct it 🙂

PR #221, opened 2021-08-14
Call tracer should ignore calls made from the library it uses

st0012:
I spotted that the call tracer ignores the debugger's methods but not the methods from related libraries, like `irb` for coloring. So it looks like this:

截圖 2021-08-14 下午4 52 51

This PR fixes the issue:

截圖 2021-08-14 下午4 53 05


st0012:
@ko1 I've rebased it on the master and it should be ready for review now.

PR #226, opened 2021-08-19
Fix NoMethodError when use trace raise with pattern

st0012:
I also improved tracer's test cases.

### DIFF
--- a/lib/debug/tracer.rb
+++ b/lib/debug/tracer.rb
@@ -58,12 +58,18 @@ module DEBUGGER__
       if tp.path.start_with?(__dir__) ||
          tp.path.start_with?('<internal:') ||
          ThreadClient.current.management? ||
-         (@pattern && !tp.path.match?(@pattern) && !tp.method_id.match?(@pattern)) ||
+         skip_with_pattern?(tp)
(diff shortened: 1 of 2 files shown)


lib/debug/tracer.rb
```diff
@@ -58,14 +58,20 @@ def skip? tp
       if tp.path.start_with?(__dir__) ||
          tp.path.start_with?('<internal:') ||
          ThreadClient.current.management? ||
-         (@pattern && !tp.path.match?(@pattern) && !tp.method_id.match?(@pattern)) ||
```
st0012, line 61: Exception event's `method_id` could be `nil` and thus cause `NoMethodError` when called with `match?`

st0012:
@ko1 I think `trace raise` should use the exception's name as filtering criteria. I can open another PR to add that if it makes sense to you?

lib/debug/tracer.rb
ko1: `method_id&.` is enough, and `||` is needed. I'll fix it later.

st0012: I missed `||` in refactor 🤦 
I've added https://github.com/ruby/debug/pull/228 for the fix.

PR #227, opened 2021-08-20
use reline

ko1:
reline 0.2.7 supports running on non-main thread so use it.
`rake test` fails so it is disabled on a test.


ko1:
image

Now Ruby expression is allowed directly.

st0012:
the `#command` and `#ruby` hints are really helpful 👍

PR #229, opened 2021-08-20
Simplify tracer messages 2

st0012:


ko1:
Could you write a commit message to understand what kind of changes, and it happy with examples.

PR #231, opened 2021-08-21
Filter ExceptionTracer with exception info

st0012:
It also changes the pattern displayed in the message to make it easier to understand:

**Before**

```
Enable ExceptionTracer (enabled) with pattern (?-mix:foo)
```

**After**

```
Enable ExceptionTracer (enabled) with pattern /foo/
```

### DIFF
--- a/lib/debug/tracer.rb
+++ b/lib/debug/tracer.rb
@@ -40,9 +40,6 @@ module DEBUGGER__
       enable
     end
 
-    def close
-    end
-
     def header depth
       "DEBUGGER (trace/#{@type}) \#th:#{Thread.current.instance_variable_get(:@__thread_client_id)} \#depth:#{'%-2d'%depth}"
     end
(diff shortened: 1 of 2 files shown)


lib/debug/tracer.rb
```diff
@@ -58,14 +55,18 @@ def skip? tp
       if tp.path.start_with?(__dir__) ||
          tp.path.start_with?('<internal:') ||
          ThreadClient.current.management? ||
-         (@pattern && !tp.path.match?(@pattern) && !tp.method_id&.match?(@pattern)) ||
+         skip_with_pattern?(tp) ||
```
ko1, line 58: It seems to remove `!tp.method_id&.match...`. Does it intentional?

st0012: From my understanding, only `CallTracer`'s tp event would have `method_id`. So I extended this behavior in `CallTracer#skip_with_pattern?`

ko1:
could you rebase it?

PR #234, opened 2021-08-22
Apply skip_path to recorder

st0012:
If the `skip_path` option is configured, stepping through the matched frames would not be possible anyway. So the recorder should not record those frames to speed up recording.

In my Rails application, the recorder is basically unusable when it records every line execution, which could be hundreds of thousands of them (I don't know because it never finished recording). But after I applied the `skip_path` pattern I have (`/lib/`), I only need to wait a few extra seconds (with 685 records).

### DIFF
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -7,6 +7,17 @@ require_relative 'frame_info'
 require_relative 'color'
 
 module DEBUGGER__
+  module SkipPathHelper
+    def skip_path?(path)
+      (skip_path = CONFIG[:skip_path]) && skip_path.any?{|skip_path| path.match?(skip_path)}
+    end
+
+    def skip_location?(loc)
+      loc_path = loc.absolute_path || "!eval:#{loc.path}"
+      skip_path?(loc_path)
+    end
+  end
+
   class ThreadClient
     def self.current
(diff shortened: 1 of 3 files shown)


lib/debug/thread_client.rb
```diff
@@ -7,6 +7,17 @@
 require_relative 'color'
 
 module DEBUGGER__
+  module SkipPathHelper
+    def skip_path?(path)
+      CONFIG[:skip_path] && CONFIG[:skip_path].any? { |skip_path| path.match?(skip_path) }
```
ko1, line 12:
```suggestion
      (skip_path = CONFIG[:skip_path]) && skip_path.any?{|skip_path| path.match?(skip_path)}
```

lib/debug/thread_client.rb
```diff
@@ -7,6 +7,17 @@
 require_relative 'color'
 
 module DEBUGGER__
+  module SkipPathHelper
+    def skip_path?(path)
+      (skip_path = CONFIG[:skip_path]) && skip_path.any?{|skip_path| path.match?(skip_path)}
```
ko1, line 12:
```suggestion
      (skip_path = CONFIG[:skip_path]) && skip_path.any?{|path| path.match?(path)}
```

use different var name (sorry my previous suggestion missed it).

PR #239, opened 2021-08-27
Colorize tracer's output

st0012:
Tracer's coloring follows the same rule as frame info's:
- `Cyan` for argument names.
- `Green` for paths.
- `Blue` for call identifier information.
- `Magenta` for return values.

## Screenshots

(The colors may be affected by my machine's settings and could be different from yours.)

### Trace Call

colorized trace call

### Trace Raise

colorized trace raise

### Trace Pass

colorized trace pass

### DIFF
--- a/lib/debug/color.rb
+++ b/lib/debug/color.rb
@@ -80,6 +80,10 @@ module DEBUGGER__
       colorize(str, [:BLUE, :BOLD])
     end
 
+    def colorize_magenta(str)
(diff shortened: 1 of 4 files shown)


ko1:
```
      # * `trace ... into: <file>`
      #   * Save trace information into: `<file>`.
```

What happens with `into`?

st0012:
@ko1 I've added a commit to avoid colorizing trace sent into files.

PR #250, opened 2021-09-05
Provide helpful message when the user calls e[val] without expressoin

st0012:
It's likely that the user is trying to access a variable 'e' from the debugger (a common name for rescued exception). This happened on me for several times just this week 🤦‍♂️ 
And the current message doesn't provide much help when this happens:

```
(rdbg) e    # eval command
eval error: no implicit conversion of nil into String
```

This commit provides better suggestions about what the user should do
instead:

```
(rdbg) e    # eval command
eval error: you must provide an expression for the e[val] command
to evaluate variables named 'e' or 'eval', please use 'pp <var>' instead
```

### DIFF
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -778,7 +778,12 @@ module DEBUGGER__
               puts "allocated at #{alloc_path}:#{ObjectSpace.allocation_sourceline(result)}"
             end
           when :call
-            result = frame_eval(eval_src)
+            if eval_src.nil?
+              puts "eval error: you must provide an expression for the e[val] command"
+              puts "to evaluate variables named 'e' or 'eval', please use 'pp <var>' instead"
+            else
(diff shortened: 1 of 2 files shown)


ko1:
how about to provide a help message?

ko1:
```
(rdbg) e    # eval command
* `e[val] <expr>`
  * Evaluate `<expr>` on the current frame.

To evaluate the variable `e`, use `pp e` instead.
(rdbg) eval    # command
* `e[val] <expr>`
  * Evaluate `<expr>` on the current frame.

To evaluate the variable `eval`, use `pp eval` instead.
```

ko1:
drastic approach is calling `pp e`, but ...

st0012:
@ko1 👍 I think your version looks better, will change.

> drastic approach is calling pp e

I think that'll make `e` without argument an implicit alias of `pp e`, which may not be what we want 🤔

Issue #246, opened 2021-09-02
Trace exception rescuing with ExceptionTracer?


st0012:
In addition to seeing where an exception is raised, it'll also help a lot to know when an exception is rescued. This will be very useful for debugging Rails applications or Rack middlewares. In these applications, rescue/re-raise exceptions based on different options is common. For example: https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/middleware/debug_exceptions.rb#L36-L40

So it'd really help if `ExceptionTracer` can also trace rescuing events. But I'm not sure if this would be technically possible when such events are not supported by `TracePoint`?

ko1:
No supports.

st0012:
I see 👍

ko1:
Now there is no support but we can consider about it.
Do you have a sample scenario?

ko1:
For `trace exception` we can see re-raised exception:

```
[master]$ exe/rdbg target.rb -e 'trace exception;; c'
DEBUGGER: Session start (pid: 12419)
[1, 8] in target.rb
     1|
     2| begin
=>   3|   raise "foo"
     4| rescue => e
     5|   raise "bar"
     6| end
     7|
     8| __END__
=>#0    <main> at target.rb:3
(rdbg:commands) trace exception
Enable ExceptionTracer (enabled)
(rdbg:commands) c
DEBUGGER (trace/exception) #th:1 #depth:1  #<RuntimeError: foo> at target.rb:3
DEBUGGER (trace/exception) #th:1 #depth:2  #<RuntimeError: bar> at target.rb:5
Traceback (most recent call last):
target.rb:3:in `<main>': foo (RuntimeError)
        1: from target.rb:2:in `<main>'
target.rb:5:in `rescue in <main>': bar (RuntimeError)
```

st0012:
I got this idea from debugging this issue: https://github.com/getsentry/sentry-ruby/issues/1563

In this case, the exception is raised from a middleware but it doesn't surface to the target middleware `Sentry::Rack::CaptureExceptions` as expected. Here's what I got from `trace exception`:

<img width="90%" alt="截圖 2021-09-03 下午3 45 28" src="https://user-images.githubusercontent.com/5079556/131969934-858e87a3-570b-43f4-90c9-3cd728119d72.png">

From the above information, I can know that:
1. It's re-raised by the `ActionDispatch::Callbacks` middleware
[...]

Issue #139, opened 2021-07-07
Inconsistent binding in CatchBreakpoint with primitive exceptions


st0012:
When stopped at a breakpoint, evaluating `binding` should reflect the current scope's binding.

<img width="80%" alt="截圖 2021-07-07 下午12 45 54" src="https://user-images.githubusercontent.com/5079556/124701522-50f3c580-df21-11eb-9865-eab22383986a.png">

But it's not the case when stopped at `CatchBreakpoint` with some primitive exceptions like `ZeroDivisionError`.

<img width="80%" alt="截圖 2021-07-07 下午12 45 38" src="https://user-images.githubusercontent.com/5079556/124701526-53eeb600-df21-11eb-9f3a-5bb5f0258313.png">

It's because those exceptions are raised in C source code, which's frame doesn't have a binding. So the debugger falls back to its own frame.

https://github.com/ruby/debug/blob/c15a460266dce5e2aebd40041176be0c5a31d5e6/lib/debug/thread_client.rb#L338-L345

I'm not sure if this edge case counts as a bug. But it'd be nice if the debugger can fall back to the nearest Ruby-level frame (with binding) in such cases.

ko1:
I think `self` should be C method's `self`, so select nearest Ruby binding will introduce confusing.

On the other hand, visible for local variables seems bad. I'll fix it.

```
(rdbg) p binding.local_variables
=> [:frame_self, :src]
```

2 variables `[:frame_self, :src]` are visible, but I don't have idea to hide them.

```patch
diff --git a/lib/debug/thread_client.rb b/lib/debug/thread_client.rb
index 792f966..eeac247 100644
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -331,6 +331,10 @@ module DEBUGGER__
       end
     end

+    def instance_eval_for_cmethod frame_self, src
+      frame_self.instance_eval(src)
+    end
+
     def frame_eval src, re_raise: false
       begin
         @success_last_eval = false
@@ -341,7 +345,7 @@ module DEBUGGER__
                    b.eval(src, "(rdbg)/#{f}")
                  else
                    frame_self = current_frame.self
-                   frame_self.instance_eval(src)
+                   instance_eval_for_cmethod(frame_self, src)
                  end
         @success_last_eval = true
         result
```

Another idea is to prohibit evaluation on C method frame. But it is useful to access self.

ko1:
BTW, I can't see the response of `catch` command.

[...]

PR #251, opened 2021-09-05
Allow accessing CatchBreakpoint's exception with `_ex_` expression

st0012:
Currently, there's no way to access the raised exception inside a `CatchBreakpoint`, which is not convenient for debugging. So this commit allows accessing the exception with the `_ex_` expression (similar to `Pry`).

But because the breakpoint can stop at both Ruby method frame (has binding) and C method frame (doesn't have binding), we can't assign the exception to a universally accessible container.

My implementation hijacks the user expression without actually evaluate it in the current scope. The drawback is that this `_ex_` expression won't show up in variables or methods list. I think this isn't a big problem as the exception isn't really stored in the program without debugger.

This feature also enables an useful combination that achieves the effect I mentioned in https://github.com/ruby/debug/issues/246#issuecomment-912363859

```ruby
binding.b(do: "catch Exception do: trace object _ex_") # traces the raised exception
```

**Example**

```ruby
binding.b(do: "catch Exception do: trace object _ex_")

def foo
  1/0
end

def compute_exception(e)
  puts(e)
end

begin
  foo
rescue => e
  compute_exception(e)
end

```

截圖 2021-09-06 上午12 53 32

### DIFF
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -438,7 +438,7 @@ module DEBUGGER__
         puts_variable_info '%return', current_frame.return_value, pat
       end
       if current_frame&.has_raised_exception
-        puts_variable_info "%raised", current_frame.raised_exception, pat
+        puts_variable_info "_ex_", current_frame.raised_exception, pat
       end
 
       if vars = current_frame&.local_variables
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -337,6 +337,11 @@ module DEBUGGER__
     end
 
     def frame_eval src, re_raise: false
+      if src == "_ex_"
+        @success_last_eval = true
+        return current_frame.raised_exception
+      end
+
       begin
         @success_last_eval = false
 
--- a/lib/debug/breakpoint.rb
+++ b/lib/debug/breakpoint.rb
@@ -263,7 +263,12 @@ module DEBUGGER__
             break
           end
         }
-        suspend if should_suspend
+        begin
+          prev, Thread.current[:DEBUGGER__last_exception] = Thread.current[:DEBUGGER__last_exception], @last_exc
+          suspend
+        ensure
+          Thread.current[:DEBUGGER__last_exception] = prev
+        end if should_suspend
       }
     end
 
@@ -467,3 +472,8 @@ module DEBUGGER__
     end
   end
 end
+
+def _ex_
+  Thread.current[:DEBUGGER__last_exception]
+end
+
--- a/test/debug/catch_test.rb
+++ b/test/debug/catch_test.rb
@@ -78,6 +78,14 @@ module DEBUGGER__
 
         type '_ex_'
         assert_line_text(/#<ZeroDivisionError: divided by 0>/)
+        type '_ex_.backtrace'
+
+        assert_line_text(
+          [
+            /.*\.rb:4:in `\/'/,
+            /.*\.rb:4:in `<main>'/
+          ]
+        )
         type 'continue'
 
         type 'continue'
@@ -99,6 +107,9 @@ module DEBUGGER__
 
         type '_ex_'
         assert_line_text(/#<RuntimeError: foo>/)
+
+        type '_ex_.backtrace'
+        assert_line_text(/\[.*:in `<main>'"\]/)
         type 'continue'
       end
     end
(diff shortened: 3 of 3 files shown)


lib/debug/thread_client.rb
```diff
@@ -337,6 +337,11 @@ def instance_eval_for_cmethod frame_self, src
     end
 
     def frame_eval src, re_raise: false
+      if src == "_ex_"
```
ko1, line 340: The implementation is not acceptable because it can not call methods like `_ex_.cause`.

st0012: ah you're right. I'll fix it.

ko1:
```
[master]$ exe/rdbg target.rb  -n
DEBUGGER: Session start (pid: 10450)
[1, 5] in target.rb
     1|
=>   2| binding.b do: 'catch Exception'
     3| raise "foo"
     4|
     5| __END__
=>#0    <main> at target.rb:2
(rdbg:binding.break) catch Exception
#0  BP - Catch  "Exception"
[1, 5] in target.rb
     1|
     2| binding.b do: 'catch Exception'
=>   3| raise "foo"
     4|
     5| __END__
=>#0    <main> at target.rb:3

Stop by #0  BP - Catch  "Exception"
(rdbg) i    # info command
%self = main
_ex_ = #<RuntimeError: foo>
(rdbg) p _ex_    # command
=> #<RuntimeError: foo>
(rdbg) p _ex_.cause    # command
=> nil
(rdbg)
```

how about that?

```patch
diff --git a/lib/debug/breakpoint.rb b/lib/debug/breakpoint.rb
index f49b285..3b84228 100644
--- a/lib/debug/breakpoint.rb
+++ b/lib/debug/breakpoint.rb
@@ -263,7 +263,12 @@ module DEBUGGER__
             break
           end
         }
-        suspend if should_suspend
+        begin
+          prev, Thread.current[:DEBUGGER__last_exception] = Thread.current[:DEBUGGER__last_exception], @last_exc
+          suspend
+        ensure
+          Thread.current[:DEBUGGER__last_exception] = prev
+        end if should_suspend
       }
     end

@@ -467,3 +472,7 @@ module DEBUGGER__
     end
   end
 end
+
+def _ex_
+  Thread.current[:DEBUGGER__last_exception]
+end
diff --git a/lib/debug/thread_client.rb b/lib/debug/thread_client.rb
index c765f60..6bb48f9 100644
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -433,7 +433,7 @@ module DEBUGGER__
         puts_variable_info '%return', current_frame.return_value, pat
       end
       if current_frame&.has_raised_exception
-        puts_variable_info "%raised", current_frame.raised_exception, pat
+        puts_variable_info "_ex_", current_frame.raised_exception, pat
       end

       if vars = current_frame&.local_variables

```

ko1:
I'm not sure `_ex_` is the best name.

ko1:
if we introduce `_ex_` or similar method, maybe we want to have a `_return_`.

st0012:
> I'm not sure _ex_ is the best name.

I'm ok with other names like `_raised_exception`. I picked `_ex_` because that reduces one less thing to learn for Pry users.

> how about that?

I think the approach is clever. But I'm not sure if switching a fiber local like this will make things messy in the long term? Also I don't think `Breakpoint#setup` is the best place to mutate this state (I understand that it needs to wrap around `#suspend` to it's easier there).

IMO, the best approach would be finding a way to store the exception in a frame-level variable for C methods. I'm not sure if it's possible at all though 🤔

st0012:
@ko1 I think we may be able to "leak" the `_ex_` variable to C method frames like this https://github.com/ruby/debug/issues/139

ko1:
why do you prefer local variables? In this case I think it is better to use dynamic scope variable (like method).

st0012:
Because it means we don't need to reset it to the previous state.

st0012:
@ko1 since I can't come up with any valid case to oppose the approach (I thought nested catch breakpoint would be an issue, but we can't really trigger another catch breakpoint inside a catch breakpoint atm), I think it's more important to ship this feature first. So I've updated the PR.

ko1:
Let's think for v1.1.

I agree if we allow nested exception, it should be a better to use variables.
The remaining issue is the name.

st0012:
@ko1 should I continue this with the local variable approach?

PR #256, opened 2021-09-08
Improve readme

st0012:
- Add command section anchor links to most usage entries.
- Add `information` and `tracing` entries to the usage section.
- Add anchor links to `binding.break` references.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,8 @@ To use a debugger, roughly you will do the following steps:
 2. Run a program with the debugger.
 3. At the breakpoint, enter the debugger console.
 4. Use debug commands.
-    * [Query the program status](#evaluate) (e.g. `p lvar` to see the local variable `lvar`).
+    * [Evaluate Ruby expressions](#evaluate) (e.g. `p lvar` to see the local variable `lvar`).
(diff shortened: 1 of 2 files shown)


ko1:
I want to leave `p lvar` because it uses debug command and no confusion.

st0012:
The point is that I don't want to make it clear that the debugger can directly evaluate a Ruby expression without `p`.
Many Rubyists that have read my debugger article may still think that it's not supported (which I updated a bit later but most people had already read it then).

ko1:
> I want to make it clear that the debugger can directly evaluate a Ruby expression without p.

Should not described here because evaluating Ruby expression needs some description.

st0012:
ok but currently the readme doesn't have an expression section, I think you'll add it?
I'll revert that change.

st0012:
@ko1 sorry please hold off the merge. I'll add more anchor links.

st0012:
@ko1 I've updated the changed and the PR description.

st0012:
@ko1 it's ready for review now. can you give it a look?

st0012:
@ko1 thank you!

PR #265, opened 2021-09-11
DAP(setExceptionBreakpoints): Handle `filters` argument

hanazuki:
Handles the `filters` argument passed to a `setExceptionBreakpoints` request.
Some editors use this mandatory argument instead of optional `filterOptions`.

ref: https://microsoft.github.io/debug-adapter-protocol/specification#Requests_SetExceptionBreakpoints

tested with dap-mode.el

### DIFF
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -153,8 +153,8 @@ module DEBUGGER__
         when 'setFunctionBreakpoints'
           send_response req
         when 'setExceptionBreakpoints'
-          filters = args.dig('filterOptions').map{|bp_info|
-            case bp_info.dig('filterId')
+          process_filter = ->(filter_id) {
+            case filter_id
             when 'any'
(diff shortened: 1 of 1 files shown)


lib/debug/server_dap.rb
```diff
@@ -163,10 +163,19 @@ def process
               bp = nil
             end
             {
-              verifiled: bp ? true : false,
```
ko1, line 166: 🙈

lib/debug/server_dap.rb
```diff
@@ -163,10 +163,19 @@ def process
...
+              verified: bp ? true : false,
               message: bp.inspect,
             }
           }
+
+          filters = args.fetch('filters').map {|filter_id|
```
ko1, line 171: ah, `filters` is a mandatory parameter.

ko1:
Thank you for trying and modifying!

PR #266, opened 2021-09-12
Expand explanation of usage in source code

brandoncc:
When I read the documentation, I thought it was saying I could use debug with `binding.irb` and `binding.pry` because of the wording that was chosen. I hope this change helps others to understand that the reference to those methods are simply examples to help readers understand how to use `binding.break`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ To use a debugger, roughly you will do the following steps:
 
 There are several options for (1) and (2). Please choose your favorite way.
 
-### Modify source code as `binding.pry` and `binding.irb`
+### Modify source code similar to `binding.pry` and `binding.irb`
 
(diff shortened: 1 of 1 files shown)


adam12 (approved):

ko1:
ah, merged, but it should be modified `misc/README.md.erb`. Could you change it too?

adam12:
> it should be modified `misc/README.md.erb`

I missed this too :(

It's come up frequently enough in past reviews. Is it worth exploring a Github Action that ensures that they are both updated consistently?

brandoncc:
I opened the PR as requested, sorry about that. Is README.md a generated file that comes from README.md.erb?

ko1:
> It's come up frequently enough in past reviews. Is it worth exploring a Github Action that ensures that they are both updated consistently?

Good idea. or simply adding a test is another idea.
The problem is Github Actions doesn't run for guest users...

PR #280, opened 2021-09-14
Fix `ForkInterceptor`'s `at_exit` callback

st0012:
When running with a puma server with 2+ workers (processes), the debugger causes 2 errors when exiting the server:

截圖 2021-09-14 下午10 23 33

I fixed them in separate commits. You can see the before/after comparison with debugging output in the commit messages.

### DIFF
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -1643,7 +1643,18 @@ module DEBUGGER__
 
           at_exit{
             trap(:SIGINT, :IGNORE)
-            Process.waitpid(child_pid) if Process.pid == parent_pid
+
+            # only check child process from its parent
+            if Process.pid == parent_pid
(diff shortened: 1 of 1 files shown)


ko1:
Ah, puma already collect child processes?

st0012:
I think puma makes sure all the child processes are dead before fully existing:

https://github.com/puma/puma/blob/6e4257fece8f822cb3cf7624e2146f7214e5b884/lib/puma/cluster.rb#L500-L521

ko1:
I forget why waitpid is here on parent process... why I put it here...

ko1:
Ah, I remember that. To keep remote connection port.

ko1:
or STDIN/OUT/ERR? I need to check it.

PR #282, opened 2021-09-14
Pin objects that are defined in Ruby.

tenderlove:
`DEBUGGER__` and `FrameInfo` are defined in Ruby, so they are not
automatically pinned.  `rb_mDebugger` and `rb_cFrameInfo` are static
global variables that hold references to these objects.  If compaction
happens, these static references can go bad.  To fix this, we register
the two references as "mark objects" so that the GC will automatically
pin the references.

Fixes: #272

### DIFF
--- a/ext/debug/debug.c
+++ b/ext/debug/debug.c
@@ -122,6 +122,11 @@ Init_debug(void)
 {
     rb_mDebugger = rb_const_get(rb_cObject, rb_intern("DEBUGGER__"));
     rb_cFrameInfo = rb_const_get(rb_mDebugger, rb_intern("FrameInfo"));
+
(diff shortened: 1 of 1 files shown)


ko1:
wow now Class/Module objects are not pinned...

tenderlove:
@ko1 Class and module are pinned if they are defined via C api.  `rb_define_class` will automatically pin the class, but `class Foo; end` will not.  IOW `cGlobal = rb_define_class(...)` is safe, but `cGlobal = rb_const_get(...)` is not safe.

ko1:
Yeah thank you for explanation. I misunderstand the specification.

PR #286, opened 2021-09-15
Small cleanup & refactoring

st0012:


lib/debug/thread_client.rb
```diff
@@ -659,18 +659,17 @@ def wait_next_action
 
     def wait_next_action_
       # assertions
-      raise "@mode is #{@mode}" unless @mode == :waiting
+      raise "@mode is #{@mode}" if !waiting?
```
st0012, line 662: I think `if !state?` conveys the expectation clearer.

ko1: `unless` is clear for me.

PR #289, opened 2021-09-16
Fix Tracer's handling on block calls

st0012:
**Before**

```
❯ ruby -Ilib -r debug target.rb
[1, 9] in target.rb
=>   1| binding.b(do: "trace object 1")
     2|       def foo
     3|         yield(1)
     4|       end
     5|
     6|       foo do |int|
     7|         puts(int)
     8|       end
     9|
=>#0    <main> at target.rb:1
(rdbg:binding.break) trace object 1
Enable ObjectTracer for 1 (enabled)
        4: from target.rb:6:in `<main>'
        3: from target.rb:3:in `foo'
        2: from target.rb:6:in `block in <main>'
        1: from /Users/st0012/projects/debug/lib/debug/tracer.rb:206:in `block in setup'
/Users/st0012/projects/debug/lib/debug/tracer.rb:100:in `minfo': undefined method `singleton_class?' for nil:NilClass (NoMethodError)

```

**After

```
❯ ruby -Ilib -r debug target.rb
[1, 9] in target.rb
=>   1| binding.b(do: "trace object 1")
     2|       def foo
     3|         yield(1)
     4|       end
     5|
     6|       foo do |int|
     7|         puts(int)
     8|       end
     9|
=>#0    <main> at target.rb:1
(rdbg:binding.break) trace object 1
Enable ObjectTracer for 1 (enabled)
DEBUGGER (trace/object) #th:1 #depth:5  1 is used as a parameter int of {block} at target.rb:6
DEBUGGER (trace/object) #th:1 #depth:5  1 receives #to_s (Integer#to_s) at target.rb:7
1
```

### DIFF
--- a/lib/debug/tracer.rb
+++ b/lib/debug/tracer.rb
@@ -201,7 +201,7 @@ module DEBUGGER__
             end
 
           out tp, " #{colorized_obj_inspect} receives #{colorize_blue(method_info)}"
-        else
+        elsif !tp.parameters.empty?
           b = tp.binding
           method_info = colorize_blue(minfo(tp))
 
--- a/lib/debug/tracer.rb
+++ b/lib/debug/tracer.rb
@@ -95,6 +95,8 @@ module DEBUGGER__
     end
 
     def minfo tp
+      return "{block}" if tp.event == :b_call
+
       klass = tp.defined_class
 
       if klass.singleton_class?
--- a/test/debug/trace_test.rb
+++ b/test/debug/trace_test.rb
@@ -320,6 +320,47 @@ module DEBUGGER__
       end
     end
 
+    def test_top_level_block_doesnt_break_tracer
+      program = <<~RUBY
+     1| tap do
+     2|   puts(1)
+     3| end
+     4|
+     5| binding.b
+      RUBY
+
+      debug_code(program) do
+        type 'trace object 1'
+        assert_line_text(/Enable ObjectTracer/)
+        type 'c'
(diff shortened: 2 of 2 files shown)


lib/debug/tracer.rb
```diff
@@ -201,7 +203,7 @@ def setup
             end
 
           out tp, " #{colorized_obj_inspect} receives #{colorize_blue(method_info)}"
-        else
+        elsif !tp.parameters.empty?
```
st0012, line 206: This isn't the fix but it makes the condition more clear + saves unnecessary method info computation.

lib/debug/tracer.rb
```diff
@@ -95,6 +95,8 @@ def puts msg
     end
 
     def minfo tp
+      return "{block}" if tp.event == :b_call
```
st0012, line 98: Since `block` is a valid method name, I think we should use additional symbols to distinguish that it's a Ruby block call.

```
# block call
DEBUGGER (trace/object) #th:1 #depth:5  1 is used as a parameter int of {block} at target.rb:6
# method call named block
DEBUGGER (trace/object) #th:1 #depth:4  1 is used as a parameter int of Object#block at target.rb:10
```

I also thought about `&block` but users may assume `block` to be the block argument's name?

ko1:
On backtrace we use `block{}` so it should be better than `{block}`.

ko1:
> On backtrace we use `block{}` so it should be better than `{block}`.

Because I think it should also output with other parameters like `block{|a=1, ...`

PR #293, opened 2021-09-19
Use local variable instead of instance variable to store the current thread client

st0012:
In the current design, thread clients' relationship with the debugger session looks like this

```
<---------------------------------- Session ---------------------------------->
<----- event (TC-1) ----> <----- event (TC-2) -----> <----- event (TC-1) ----->
```

It always comes with an event and doesn't stay in the session after the event is processed.

But because it's currently stored in the session's instance variable `@tc`, we need to manage its state manually to make sure 
1. it's assigned before an event is started
2. it's cleared after the event is finished

https://github.com/ruby/debug/blob/07c1a7c23881b7c4ce1ba953762ebf4c5e18a4d4/lib/debug/session.rb#L296-L298
https://github.com/ruby/debug/blob/07c1a7c23881b7c4ce1ba953762ebf4c5e18a4d4/lib/debug/session.rb#L1344-L1353

Also, its availability is not guaranteed when processing different commands. For example, if the command is `step`, the `@tc` will be `nil` (reseted by `restart_all_threads`) until the next event sets its thread client. But if the command is `bt`, `@tc` will be remain unchanged.

```
<---------------------------------- Session -------------------------------->
<-------------- step cmd ------------> <-------------- bt cmd --------------> 
<--- @tc (TC-1) ---> <-- @tc (nil) --> <------------ @tc (TC-2) ------------>
```

Although none of these behaviors are causing issues, I think by making it a local variable we can:
- save the work to manage the ivar's state and avoid future bugs.
- make thread client's availability clear just by reading the code.

### DIFF
--- a/test/debug/thread_test.rb
+++ b/test/debug/thread_test.rb
@@ -4,43 +4,49 @@ require_relative '../support/test_case'
 
 module DEBUGGER__
   class ThreadTest < TestCase
+    POSSIBLE_STATES = "(sleep|run)"
+
     def program
       <<~RUBY
-     1| Thread.new do
-     2|   i = 0
-     3|   while true do
-     4|     i += 1
+     1| def fib(n)
+     2|   first_num, second_num = [0, 1]
+     3|   (n - 1).times do
+     4|     first_num, second_num = second_num, first_num + second_num
      5|   end
-     6| end
-     7| sleep 0.1
-     8| binding.b
+     6|   first_num
+     7| end
+     8|
+     9| Thread.new do
+    10|   fib(1_000_000)
+    11| end
+    12|
+    13| sleep(0.1)
+    14| binding.b
       RUBY
     end
 
-    def test_prints_all_threads
+    def test_prints_all_threads_and_can_switch_to_another_one
       debug_code(program) do
         type 'c'
         type 'thread'
         assert_line_text(
           [
-            /--> #0 \(sleep\)@.*:8:in `<main>'/,
-            /#1 \(sleep\)/
+            /--> #0 \(#{POSSIBLE_STATES}\)@.*:\d+:in `<main>'/,
+            /#1 \(#{POSSIBLE_STATES}\)/
           ]
         )
-        type 'q!'
-      end
-    end
 
-    def test_switches_to_the_other_thread
-      debug_code(program) do
-        type 'c'
-        type 'thread 1'
-        assert_line_text(
-          [
-            /#0 \(sleep\)@.*:8:in `<main>'/,
-            /--> #1 \(sleep\)/
-          ]
-        )
+        # currently there's no easy way to make sure that a thread will be "under control" (can be switched to) consistently
+        if false
+          type 'thread 1'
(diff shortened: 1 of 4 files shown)


test/debug/thread_test.rb
```diff
@@ -0,0 +1,54 @@
...
+          ]
+        )
+
+        # currently there's no easy way to make sure that a thread will be "under control" (can be switched to) consistently
+        if false
+          type 'thread 1'
+          assert_line_text(
+            [
+              /#0 \(#{POSSIBLE_STATES}\)@.*:\d+:in `<main>'/,
+              /--> #1 \(#{POSSIBLE_STATES}\)/
```
st0012, line 45: @ko1 I've tried various approaches but I can't find a way to make sure a thread is in the `under control` state, no matter it's running or sleeping. So I need to skip this case first to avoid randomly breaking tests.

ko1:
> In the current design, thread clients' relationship with the debugger session looks like this
> 
> ```
> <---------------------------------- Session ---------------------------------->
> <----- event (TC-1) ----> <----- event (TC-2) -----> <----- event (TC-1) ----->
> ```

Sorry I can't understand the chart. How to read it?

st0012:
```
<---------------------------------- Session ---------------------------------->
<----- event (TC-1) ----> <----- event (TC-2) -----> <----- event (TC-1) ----->
```

`event` means events from `@q_evt.pop` and the y-axis is time. I want to demonstrate that when the session handles several events in `session_server_main`, each event can come with different thread clients.

ko1:
>  y-axis is time

So do `Session` and 3 events come at same time?

I'm not sure (1) why there are two events from TC-1 at the same time (2) why session is finished.

st0012:
@ko1 that's a typo 🤦‍♂️ I meant x-axis is time, sorry.

ko1:
Ah, I see.

Your opinion is, using local variables is cleaner than using ivars. Let me consider about that.

ko1:
Now, I'm thinking current ivar approach is better because most of methods are limited run in session server and there is no danger to introduce wrong ivar (`@tc`). In other words, if wrong `@tc` is set and run, it should be bug.

This assertion (methods should run only in session server) should be guaranteed by the code.

st0012:
@ko1 I know it'd be a bug if `@tc` is incorrectly set. The problem is, because it's an instance variable, it's not easy to track its status change and debug. 

For example, `restart_all_threads` would set `@tc = nil`. So if you call a method that implicitly relies on `@tc`, like `repl_add_breakpoint`, after `restart_all_threads` is called, it'll raise `NoMethodError`. 
This means that there are ordering dependencies between session helpers, like

```rb
# this would work
step_command
repl_add_breakpoint

# this would break
repl_add_breakpoint
step_command
```

I can see myself getting bitten by this kind of implicit ordering when developing features in the future. (I spotted this when experimenting with TUI support).

So the idea is that if we can make `tc` a local variable and pass it around as an explicit dependency of helpers, it'll be easy to spot future problems.

Issue #298, opened 2021-09-23
Ruby 3.0.2 claims to ship 0.1.0 but 0.1.0 gem has been yanked


headius:
This is an out of sync gem change. It is not possible to install the same debug gem as shipped with Ruby 3.0.2.

https://github.com/ruby/ruby/blob/ruby_3_0/lib/debug.gemspec#L3

The 0.1.0 gem has been yanked.

headius:
The resolv library has a few tweaks for JRuby that may or may not be things we can fix.

Diffs are here: https://gist.github.com/headius/6508dfe066630bfa42e4d53b70c9c72e

hsbt:
I yanked debug-0.1.0 because it's the different implementation with ruby/debug. I will backport debug-0.2.0 to Ruby 3.0.3 at https://github.com/ruby/ruby/pull/5104/commits/b5e338f97e264aee7ecd87616838a63ad06f85d3.

You can use debug-0.2.0 at https://github.com/ruby/debug/releases/tag/v0.2.0

PR #301, opened 2021-09-23
Support JRuby and update from CRuby 3.0

headius:
This contains changes to support JRuby:

* Mask out continuation usage
* Mask out iseq usage

It also contains changes to align with CRuby 3.0.2, which claims to ship 0.1.0 (#298) but does not actually do so.

* "expand filenames in breakpoints" from https://github.com/ruby/ruby/commit/ed5d032ea33a465fd7152a236a4aca0ff5653bf3
* doc fixes from https://github.com/ruby/ruby/commit/1819652578e8f9fe3606f7a716ec4e427fc55f0a

Once this is merged it should be released as 0.2.1 and Ruby 3.x should update to use that version.

### DIFF
--- a/lib/debug.rb
+++ b/lib/debug.rb
@@ -3,8 +3,6 @@
 # Copyright (C) 2000  Information-technology Promotion Agency, Japan
 # Copyright (C) 2000-2003  NAKAMURA, Hiroshi  <nahi@ruby-lang.org>
 
-require 'continuation'
-
 if $SAFE > 0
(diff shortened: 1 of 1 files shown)


hsbt:
@headius You hope to release v0.2.1 with this patch?

hsbt:
I've requested v0.2.1 to Ruby 3.0.3 with this changes. Thanks!

headius:
Thank you! I will update our default version of debug for JRuby 9.4.

PR #313, opened 2021-09-30
Refactor Client's util usage

st0012:


lib/debug/config.rb
```diff
@@ -313,7 +313,17 @@ def self.parse_argv argv
 
         o.on('--util=NAME', 'Utility mode (used by tools)') do |name|
           require_relative 'client'
-          Client.new(name)
+
```
ko1, line 316: Please move it into Client like `Client.util(name)`. config.rb is loaded every time and it should be small.

PR #314, opened 2021-09-30
Unify test commands

st0012:
Currently, local tests are ran with `ruby -r debug/start` while remote tests are ran with `exe/rdbg` executable. This mix of commands causes confusion and makes it harder to apply command options (which only works with executable). So this PR changes local tests to start with `exe/rdbg` as well. 

It also introduces a new `execute_without_debugger` helper for scenarios that don't require debugger at the beginning.


st0012:
@ko1 It runs `ruby -Ilib script.rb`, which doesn't require the debugger. It's only for testing `require "debug"` or `require "debug/start"` manually in the code.

ko1:
Thank you. I understand.
For me it is special case of `debug_code` because it uses debugger after launching.
So `debug_code` with option seems enough, I think.

Now I remember that this is why we use `-r debug/start` instead of `rdbg` to achieve start a debuggee without the debugger.

ko1:
So your proposal is:

* `debug_code`: use `rdbg`
* `execute_without_debugger`: run with bare ruby

How about that?

* `debug_code do ... end` -> `run_rdbg script do ... end`
* `run_ruby script do ... end`

and they are explicit.

@ono-max san do you have any opinion?

st0012:
Yeah I like the idea 👍

ko1:
ah, another opinion.

if we use `debug_code`, we can tests with `rdbg` and `-r debug/...` internally, if possible. So current `debug_code` has also advantage.

st0012:
Do we need a full coverage for `-r debug/start`? I think having a case like `run_ruby script, option: "-r debug/start"` to test the debugger startup should be enough.

ono-max:
I'm weak negative about dividing `debug_code` into several methods. Because most of the test cases are executed by `rdbg`, other test cases such as `-r debug/start` are the special cases. They are enough with options of `debug_code`.

st0012:
@ono-max The reason to separate `rdbg` and `ruby` commands is because we'll add tests for command line options, which only works with `rdbg`. If we keep everything under `debug_code`, we'll have combinations like

```rb
debug_code(program) # test ruby + rdbg 
debug_code(program, remote: false) # test ruby + rdbg local
debug_code(program, executable_options: "--nonstop") # test rdbg with options
debug_code(program, boot_options: "", remote: false) # test ruby
```

And this is confusing to people who don't understand the history.

So unless it's necessary to cover all command tests with `ruby -r debug/start`, I prefer a clear separation like:

```rb
run_rdbg(program) # test rdbg 
run_rdbg(program, options: "--nonstop") # test rdbg with options
run_ruby(program) # test ruby
run_ruby(program, options: "-r debug/start") # test ruby with options
```

st0012:
@ko1 I've renamed the 2 commands and added a case for `-r debug/start`.

ono-max:
I see. Thank you for explaining to me. Let me ask this question.

@ko1 What do you think about this opinion?

> Do we need a full coverage for -r debug/start? I think having a case like run_ruby script, option: "-r debug/start" to test the debugger startup should be enough.

ono-max:
Another solution is https://github.com/ruby/debug/pull/317

st0012:
no, #317 doesn't solve the problem. how come different values of `boot_options` can change the underlying command to start the debugger? any developer outside this conversation would not expect that. it feels like a workaround to me. also, another problem of that solution is `boot_options` only applies to local tests instead of all tests.

ko1:
> And this is confusing to people who don't understand the history.

Sorry, why is it confusing?

```ruby
debug_code(program) # test ruby + rdbg 
debug_code(program, remote: false) # test ruby + rdbg local
debug_code(program, executable_options: "--nonstop") # test rdbg with options
debug_code(program, boot_options: "", remote: false) # test ruby
```

For example, to make it clear,

```ruby
debug_code program do ... end # test ruby + rdbg (do not care)
debug_code program, remote: false # test ruby + rdbg (do not care) local
debug_code(program, rdbg: '--nonstop') # test rdbg with options
debug_code(program, ruby: `...`, remote: false) # test with ruby
```

they are clear for me.

If we introduce convenience methods `run_...` we can introduce the following.

* `debug_code program do ... end` # don't care how to run, and many cases. we can introduce other words like `run_scenario program ...`
* `debug_code` can be replaced with `run_code` for example.
* `run_rdbg program, option: "--xxx" do ... end` # run only on rdbg
* `run_ruby program, ... do ... end` # run only on ruby

are fine for me, for example. But I'm not sure how many `run_ruby/rdbg` is used.

st0012:
```rb
debug_code(program, rdbg: '--nonstop') # test rdbg with options
debug_code(program, ruby: `...`, remote: false) # test with ruby
```

This is the confusing part. Why is the `ruby` option need to be used with `remote: false`? This creates implicit dependencies between options that's not intuitive to API users.

I can agree with this format if `debug_code(program, ruby: '...')` doesn't require the `remote: false`.

test/support/utils.rb
```diff
@@ -71,44 +71,50 @@ def collect_debuggee_backlog test_info
...
-      inject_lib_to_load_path
-
-      ENV['RUBY_DEBUG_NO_COLOR'] = 'true'
-      ENV['RUBY_DEBUG_TEST_MODE'] = 'true'
-      ENV['RUBY_DEBUG_NO_RELINE'] = 'true'
+    def debug_code(program, ruby: nil, rdbg: nil, remote: true, &test_steps)
+      if ruby
+        run_ruby(program, options: ruby, &test_steps)
+      else
+        run_rdbg(program, remote: remote, &test_steps)
```
st0012, line 78: I will support the `options` argument when adding tests for `rdbg` command, which will be the next PR after this is merged.

st0012:
@ko1 I've updated the methods with this specification

```rb
debug_code(program) # run rdbg
debug_code(program, ruby: "") # run ruby
debug_code(program, rdbg: "") # run rdbg
run_ruby(program, options: "") # run ruby
run_rdbg(program, options: "") # run rdbg
```

And regarding edge cases like

```rb
# should we prohibit these?
debug_code(program, ruby: "", remote: false) 
debug_code(program, rdbg: "", ruby: "")
```

I want to do that in the future because the current priority should be adding tests for `rdbg` options.

ono-max:
As @ko1 said, as shown below, we’re not sure how many `run_ruby/rdbg` is used.

> But I’m not sure how many run_ruby/rdbg is used.

How about starting from `debug_code(program)`, `debug_code(program, ruby: “”) `, `debug_code(program, rdbg: “”)`? Then, if we need a lot of test cases of running pure Ruby, we can introduce `run_ruby(program, options: “”)` or `run_rdbg(program, options: “”)`.

st0012:
The latest code already does that. (except `debug_code(program, rdbg: “”)` because currently we have no use case for it yet. I'll add it with new options tests")

st0012:
```rb
debug_code(program, ruby: "", remote: false) 
```

We currently don't have remote tests with Ruby yet. So we cant support this later if it's necessary.

```rb
debug_code(program, rdbg: "", ruby: "")
```

I've added a check to prohibit that.

Issue #308, opened 2021-09-29
`irb` command doesn't restore the prompt


ko1:
```
[master]$ exe/rdbg target.rb
[1, 5] in target.rb
=>   1| require 'reline'
     2| binding.irb
     3| prompt = Reline.readmultiline("hoge> ", true){ true }
     4|
     5| __END__
=>#0    <main> at target.rb:1
(rdbg) irb    # command
:irbrc_loaded
irb(main):001:0> exit
irb(main):002:0>             # now it is debugger console, but prompt is still irb's
```

This may be a bug of `reline`/`binding.irb`:

```
require 'reline'
binding.irb
prompt = Reline.readmultiline("hoge> ", true){ true }
```

and I want to find out the workaround.

ko1:
`Reline.prompt_proc = nil` fixes this issue.

PR #353, opened 2021-10-31
Fix bug that test framework gets stuck when debuggee outputs many lines

### LINKED ISSUE #318: Test doesn't work if debuggee outputs many lines
**Your environment**

* `ruby -v`:  3.1dev
* `rdbg -v`: 1.2.3

**Describe the bug**

```ruby

module DEBUGGER__
  class FooTest < TestCase
    def test_foo
      debug_code ' 1| 200.times{|i| p i}' do
        type 'c'
        assert_finish
      end
    end
  end
end
```

becomes timeout with remote tests.

**Expected behavior**

it should not fail.

### LINKED ISSUE #322: The output from the remote debuggee can't be checked in the current test framework
@ko1-san  pointed out that there is a problem such as the content of the title.

After #318 is solved(I created [the PR](https://github.com/ruby/debug/pull/321) to solve it), I'll fix this problem.

ono-max:
Fixes #318

I also defined `assert_debuggee_out` method to fix #322

### DIFF
--- a/test/support/assertions.rb
+++ b/test/support/assertions.rb
@@ -57,6 +57,19 @@ module DEBUGGER__
       })
     end
 
+    def assert_debuggee_line_text text
+      @scenario.push(Proc.new {|test_info|
+        next if test_info.mode == 'LOCAL'
+
+        log = test_info.remote_info.debuggee_backlog.join
+        msg = "Expected to include `#{text.inspect}` in\n(\n#{log})\n"
+
+        assert_block(FailureMessage.new{create_message(msg, test_info)}) do
+          log.match? text
+        end
+      })
+    end
+
     def assert_block msg
       if multithreaded_test?
         # test-unit doesn't support multi thread
--- a/test/support/utils.rb
+++ b/test/support/utils.rb
@@ -23,14 +23,12 @@ module DEBUGGER__
 
       debuggee_msg =
         if test_info.mode != 'LOCAL'
-          debuggee_backlog = collect_debuggee_backlog(test_info)
-
           <<~DEBUGGEE_MSG.chomp
             --------------------
             | Debuggee Session |
             --------------------
 
-            > #{debuggee_backlog.join('> ')}
+            > #{test_info.remote_info.debuggee_backlog.join('> ')}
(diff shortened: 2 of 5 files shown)


test/support/utils.rb
```diff
@@ -316,6 +298,7 @@ def check_error(error, test_info)
     def kill_remote_debuggee remote_info
       return unless remote_info
 
+      sleep 0.1 until remote_info.reader_thread.kill
```
ko1, line 301: `Thraed#kill` returns self. Why do you use `until`?

test/support/assertions.rb
```diff
@@ -57,6 +57,21 @@ def assert_no_line_text(text)
       })
     end
 
+    def assert_debuggee_out exp
+      @scenario.push(Proc.new {|test_info|
+        if test_info.mode == 'LOCAL'
+          log = test_info.last_backlog.join
+        else
+          log = test_info.remote_info.debuggee_backlog.join
```
ko1, line 65: do you need `last_debuggee_backlog`?

test/support/assertions.rb
```diff
@@ -57,6 +57,21 @@ def assert_no_line_text(text)
       })
     end
 
+    def assert_debuggee_out exp
```
ko1, line 60: I think it should be consistent with `assert_line_text` (name and parameters)

ono-max: https://github.com/ruby/debug/pull/365

ko1: Just now it is not renamed yet, so it should be renamed after #308.
Or merge this pull request after #308?

ono-max: Either is fine, but if I had to say, I prefer the former. To make it consistent with #365, we may need to change `assert_debuggee_out`, too.

ko1: Former option means introduce current consistent name (maybe `assert_debugee_line_text` or similar) and rename it with #365. Could you try it?

ono-max: I renamed it to "assert_debuggee_line_text".

PR #364, opened 2021-11-03
Create pull_request_template.md

st0012:
For now it's mainly for guiding readme changes. But in the long-term, I think we can introduce more instructions here as well.


.github/pull_request_template.md
```diff
@@ -0,0 +1,9 @@
+Thanks for your Pull Request 🎉 
+
+**Please follow these instructions to help us review it more efficiently:**
+
+- Add references of related issues/PRs in the description.
```
ko1, line 5: "... if available"?

PR #365, opened 2021-11-04
Rename some assert methods

ono-max:
In https://github.com/ruby/debug/pull/314, we changed some parts in the test framework. With that background and after the result of the discussion with @ko1-san, we think that renaming assert methods are a good opportunity now. 

- `assert_line_text` -> `assert_debugger_out`
- `assert_no_line_text` -> `assert_debugger_noout`
- (Align all arguments to `exp`)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -240,7 +240,7 @@ Passes if `expected` is equal to the location where debugger stops.
 
 Passes if `text` is included in the last debugger log.
 
-- assert_no_line_text(text)
+- assert_debugger_noout(text)
 
 Passes if `text` is not included in the last debugger log.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -232,17 +232,17 @@ The following table shows examples of the gentest options.
 
 ### Assertions
 
-- assert_line_num(expected)
+- assert_line_num(exp)
 
-Passes if `expected` is equal to the location where debugger stops.
+Passes if `exp` is equal to the location where debugger stops.
 
-- assert_debugger_out(text)
+- assert_debugger_out(exp)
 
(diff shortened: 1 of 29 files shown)


CONTRIBUTING.md
```diff
@@ -232,17 +232,17 @@ The following table shows examples of the gentest options.
 
 ### Assertions
 
-- assert_line_num(expected)
+- assert_line_num(exp)
```
st0012, line 235: I'm not sure if this abbreviation is necessary. `expected` is already a common word while `exp` is a bit ambiguous (can be interpreted as `expression`).

st0012 (commented):
I don't usually see usages of words like `out` or `noout` and perhaps `assert_debugger_output` and `assert_no_debugger_output` would be clearer.

ko1:
How about to use `assert_out` (or `assert_output`) because checking debuggee's output is relatively low frequency.

ko1:
or `assert_line` `assert_text` `assert_debuggee_text` to make them shorter.

CONTRIBUTING.md
ono-max: `exp` is a popular word in minitest.
http://docs.seattlerb.org/minitest/Minitest/Assertions.html

ono-max: I understand what you are saying. I'll revert it.

ono-max:
I prefer `assert_out` or `assert_output` because they're used in test-unit and byebug.

test-unit: `assert_output`
byebug: `check_output_includes`

ko1:
> test-unit: `assert_output`

So we can't this name, couldn't?

ono-max:
> So we can't this name, couldn't?

Oh, that's right.

ono-max:
So, `assert_out` is fine for me.

PR #369, opened 2021-11-06
Catch StandardError when no catch argument is given

st0012:
Currently, using `catch` command without argument prints all breakpoints, which isn't helpful as we already have `b[reak]` command for that. I think it'll be convenient if we can change that to `catch StandardError` (which I do quite often) and save users from typing the long word.

### DIFF
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -558,12 +558,15 @@ module DEBUGGER__
       when 'catch'
         check_postmortem
 
-        if arg
-          bp = repl_add_catch_breakpoint arg
-          show_bps bp if bp
-        else
-          show_bps
-        end
+        bp =
+          if arg
+            repl_add_catch_breakpoint arg
+          else
+            repl_add_catch_breakpoint "StandardError"
+          end
+
+        show_bps bp if bp
+
         return :retry
 
       # * `watch @ivar`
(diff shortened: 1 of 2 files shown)


ko1:
I'm negative because there is a convention to show all configurations are printed without any arg like `break`, `config`, `display` and so on. It break the consistency. Now it shows breakpoints, but should it only `CatchBreakpoint`?

I think this kind of support should be provided by completions.

Anothr comment is: Is it ok to use `StandardError`? (for example, `RuntimeError`)

st0012:
> Is it ok to use StandardError? (for example, RuntimeError)

In integration tests (e.g. Rails' request specs), the test subject usually rescues errors and wraps it with a more generic result (e.g. you get a 500 response when something raised internally). And in my app, it's possible to have 500 caused by `RuntimeError` or our custom exception class (which inherits `StandardError`). So at that point I don't know the exception and what I usually do is using `binding.b(do: "catch StandardError")` to see what went wrong. (Postmortem mode doesn't work in this case as the process doesn't finish with exception).

> I'm negative because there is a convention to show all configurations are printed without any arg like break, config, display and so on. It break the consistency. Now it shows breakpoints, but should it only CatchBreakpoint?

Yeah it breaks the consistency. But I don't feel `catch` needs its own display command because unlike `break` or `watch`, there's a limited amount of `catch` breakpoint you can put without overlapping and I certainly never used it. So I think it'll be great if we can give the spot to something used more frequently.

PR #370, opened 2021-11-06
Improve auto completion

st0012:
This is a simple implementation to achieve the following completion:

- Instance variable

instance variable completion

- Local variable + method
local variable + method completion

- Constant
constant completion

- Top level constant

It doesn't multi-level completion (e.g. `Foo::Bar`'s `Foo` can be completed but `Bar` can't). If this PR's approach works, I'll add future support for that.

Also, I haven't found a way to simulate `tab` input in test environment so I can't write any tests. Still trying though.

### DIFF
--- a/lib/debug/console.rb
+++ b/lib/debug/console.rb
@@ -1,6 +1,30 @@
 # frozen_string_literal: true
 module DEBUGGER__
   class Console
+    COMPLETION_PROC = -> given do
+      thread_client = SESSION.managed_thread_clients.first
+      binding = thread_client.current_binding
+      frame_self = thread_client.current_frame&.self
+
+      candidates =
+        case given
+        when /\A@\w/
+          frame_self&.instance_variables
+        when /\A[A-Z]/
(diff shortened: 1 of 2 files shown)


lib/debug/console.rb
```diff
@@ -1,6 +1,39 @@
...
+        when /\A@\w/
+          frame_self.instance_variables
+        when /\A[A-Z]/
+          if frame_self.is_a?(Module)
+            frame_self.constants
+          else
+            frame_self.class.constants
+          end
+        when /^::([A-Z][^:\.\(\)]*)$/
+          # Absolute Constant or class methods
```
st0012, line 24: @ko1 this case's implementation is borrowed from `irb`. I've added acknowledgement in the commit message, but I'm not sure if I should add one in code comment too?

ko1:
This approach is not acceptable because of Ractor support.

st0012:
@ko1 because the way it retrieves thread client?

PR #374, opened 2021-11-08
Add more rdbg tests

st0012:
This PR adds a few new rdbg tests. But more importantly, it excludes user-defined `~/.rdbgrc` from tests and restore it later.


test/support/test_case.rb
```diff
@@ -10,9 +10,16 @@ module DEBUGGER__
   class TestCase < Test::Unit::TestCase
     include TestUtils
     include AssertionHelpers
+    RC_FILENAME = File.join(Dir.home, ".rdbgrc")
```
ko1, line 13: Do not modify the `~/.rdbgrc` on a test.

st0012: It only renames the file while the test started and rename it back when it's finished. So eventually the file isn't changed at all. Without this, existing `~/.rdbgrc` will break the tests.

ko1: Surprisingly, there are platforms that doesn't have HOME (and Ruby tests run on them).

Please use `Dir.chroot` to manipulate home dir.
(make a temp dir and make home dir under the temp dir. And use chroot)

test/support/test_case.rb
```diff
@@ -16,10 +16,14 @@ class TestCase < Test::Unit::TestCase
 
     def setup
       @temp_file = nil
+      @original_root = Dir.home("root")
+      dir = Dir.mktmpdir
+      Dir.chroot(dir)
```
st0012, line 21: @ko1  I think `Dir.chroot` requires privileged permission to call and I'm not able to find a replacement for it. Do you know any alternatives?

ko1: I see.
@nobu had advised that changing `ENV['HOME']` is easy way for this kind of hack.

test/support/utils.rb
```diff
@@ -178,7 +178,7 @@ def prepare_test_environment(program, test_steps, &block)
     TIMEOUT_SEC = (ENV['RUBY_DEBUG_TIMEOUT_SEC'] || 10).to_i
 
     def run_test_scenario cmd, test_info
-      PTY.spawn(cmd) do |read, write, pid|
+      PTY.spawn({ "HOME" => @pty_home_dir }, cmd) do |read, write, pid|
```
st0012, line 181: @ko1 I think this is better than updating `ENV["HOME"]`. Wdyt?

ko1: It seems good idea but Dir.tmpdir is not removed automatically.

st0012: Good catch. I've updated it.

ko1: It seems too much to make a directory per tesst. `at_exit` will help you.

st0012: Yeah. I eventually picked the [startup/shutdown callbacks](https://www.rubydoc.info/gems/test-unit/Test/Unit/TestCase) instead.

peterzhu2118 (approved):
Looks good, thanks!

PR #382, opened 2021-11-10
fix warning about regexp division ambiguity

dorianmariefr:
I was getting:

```
/Users/dorianmariefr/src/clones/debug/test/debug/control_flow_commands_test.rb:207: warning: ambiguity between regexp and two divisions: wrap regexp in parentheses or add a space after `/' operator
```

I'm on ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-darwin20]

kind of surprising that this could be seen as multiple divisions but sure


ono-max (approved):
Good catch!

Thank you

PR #383, opened 2021-11-10
add fish support and instructions for init command

### LINKED ISSUE #377: Support for fish for `rdbg --util=init -`
**Your environment**

* `ruby -v`: ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-darwin20]
* `rdbg -v`: rdbg 1.3.4

**Describe the bug**

I'm trying to have `debugger` available in all my gems and apps when debugging, so I would like to use the init feature, but I'm using fish and have to do it manually:

```
set -x RUBYOPT "-r /Users/dorianmariefr/.rvm/gems/ruby-3.0.2/gems/debug-1.3.4/lib/debug/prelude" $RUBYOPT
```

**To Reproduce**

`eval (rdbg --util=init -)` in a fish shell

**Expected behavior**

it works and preloads debug gem

dorianmariefr:
fixes #377 

> To give a variable to an external command, it needs to be "exported". Unlike other shells, fish does not have an export command. Instead, a variable is exported via an option to set, either --export or just -x.

https://fishshell.com/docs/current/tutorial.html#exports-shell-variables

```
~/s/c/debug> rdbg --util=init
# add the following lines in your .bash_profile
eval "$(rdbg --util=init -)"

# Add `Kernel#bb` method which is alias of `Kernel#debugger`
# export RUBY_DEBUG_BB=1

# for fish shell, in your ~/.config/fish/config.fish
# eval (rdgb --util=init - --fish)
```

```
~/s/c/debug> rdbg --util=init - --fish
set -x RUBYOPT "-r /Users/dorianmariefr/.rvm/gems/ruby-3.0.2/gems/debug-1.3.4/lib/debug/prelude" $RUBYOPT
```

- [x] with tests

### DIFF
--- a/lib/debug/client.rb
+++ b/lib/debug/client.rb
@@ -14,26 +14,46 @@ module DEBUGGER__
 
   class Client
     class << self
-      def util name
+      def util name, argv: []
         case name
+        when 'current-shell'
+          puts current_shell
         when 'gen-sockpath'
           puts DEBUGGER__.create_unix_domain_socket_name
         when 'list-socks'
           cleanup_unix_domain_sockets
           puts list_connections
         when 'init'
-          if ARGV.shift == '-'
-            puts <<~EOS
-            export RUBYOPT="-r #{__dir__}/prelude $(RUBYOPT)"
-            EOS
+          if argv.shift == '-'
+            case current_shell
+            when :fish
+              puts <<~EOS
+              set -x RUBYOPT "-r #{__dir__}/prelude" $RUBYOPT
+              EOS
+            else
+              puts <<~EOS
+              export RUBYOPT="-r #{__dir__}/prelude $RUBYOPT"
+              EOS
+            end
           else
-            puts <<~EOS
-            # add the following lines in your .bash_profile
-            eval "$(rdbg --util=init -)"
+            case current_shell
+            when :fish
+              puts <<~EOS
+              # add the following lines to your .config/fish/config.fish
+              eval (rdbg --util=init -)
 
-            # Add `Kernel#bb` method which is alias of `Kernel#debugger`
-            # export RUBY_DEBUG_BB=1
-            EOS
+              # Add `Kernel#bb` method which is alias of `Kernel#debugger`
+              # set -x RUBY_DEBUG_BB 1
+              EOS
+            else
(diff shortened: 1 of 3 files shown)


ono-max:
Thanks!

Could you move `test/debug/client_test.rb` to `test/client_test.rb` or make a new directly like `test/unit/client_test.rb`?

The `test/debug` directly is for integration tests 👍

dorianmariefr:
@ono-max done

lib/debug/client.rb
```diff
@@ -56,6 +76,15 @@ def cleanup_unix_domain_sockets
       def list_connections
         Dir.glob(DEBUGGER__.create_unix_domain_socket_name_prefix + '*')
       end
+
+      def current_shell
+        case `ps -p #{Process.ppid} -o 'args='`.strip
+        when 'fish', '-fish'
+          :fish
```
JoshCheek, line 83: I'm not sure whether ps works on Windows or not. I googled it to see (I don't have a Windows box to try it on), and the internet suggested that it's called tasklist on Windows.

Anyway, when I run it, it returns the full path, which won't match the case statement:

```sh
$ ruby -e 'p `ps -p #{Process.ppid} -o args=`.strip'
"/opt/homebrew/bin/fish"
```

Probably a regex would be more resilient, something like `/\bfish\z/`

dorianmariefr: Thanks, didn't think about Windows, I can't test on Windows though

dorianmariefr: I did `File.basename(...)` so that it works on your setup too, on mine I get:

```
> ruby -e 'p `ps -p #{Process.ppid} -o args=`.strip'
"-fish"
```

dorianmariefr: I'm downloading a Windows VM to test on Windows

ko1: On windows, there is no place to use this feature, so no problem.

dorianmariefr: yeah adding windows support would be another larger pull request

ko1:
I made another patch with reconsidering whole of this feature: https://github.com/ruby/debug/pull/428
with fish support. Thank you for your contribution!

PR #384, opened 2021-11-10
add *.gem to .gitignore

dorianmariefr:
Doing `gem build` generates `.gem` files and we don't want those in git.

I also added it to my gitignore global but I thought this could help others

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@
 /Gemfile.lock
 /lib/debug/debug.so
 .ruby-version
+*.gem


ko1:
thanks, but .gem is located in `tmp/` so now it is needed.

dorianmariefr:
@ko1 I just do `gem build` and it's not in `tmp/` folder but in the main folder, maybe you use another command? but that's the most common (standard) command as far as I know

hsbt:
Use `rake build` instead of `gem build`.

dorianmariefr:
ah okay makes sense

PR #388, opened 2021-11-13
Refactor the logic by removing an intermediate variable

ono-max:


ono-max:
Oh, this change seems to be wrong.

PR #389, opened 2021-11-13
Fix unused block arguments

ono-max:


st0012 (commented):
Good catches 👍 And perhaps we can change them from `#each` to `#each_value`?

ono-max:
Thanks, that's a good suggestion. Let me change it.

PR #392, opened 2021-11-15
Fix several warnings when it runs tests

ono-max:
The following warnings are fixed in this PR.

```shell
warning: instance variable @temp_file not initialized
warning: ambiguous first argument; put parentheses or a space even after `/' operator
```


olleolleolle (approved):
These are good repairs, reduces the output amount when running tests.

PR #393, opened 2021-11-15
Add condition and command support to watch command

st0012:
`watch` is the only breakpoint command that doesn't support condition and command options (`if`, `do`, and `pre`). It's not by design and the inconsistency between it and other commands may confuse the users. So after discussing with @ko1, we're supporting those options on `watch` too.

### DIFF
--- a/lib/debug/breakpoint.rb
+++ b/lib/debug/breakpoint.rb
@@ -328,12 +328,15 @@ module DEBUGGER__
   end
 
   class WatchIVarBreakpoint < Breakpoint
-    def initialize ivar, object, current
+    def initialize ivar, object, current, cond: nil, command: nil
       @ivar = ivar.to_sym
       @object = object
       @key = [:watch, @ivar].freeze
 
       @current = current
+
+      @cond = cond
+      @command = command
       super()
     end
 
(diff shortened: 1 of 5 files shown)


st0012:
I think the binding for condition evaluation needs more work.

lib/debug/breakpoint.rb
```diff
@@ -328,12 +328,15 @@ def to_s
   end
 
   class WatchIVarBreakpoint < Breakpoint
-    def initialize ivar, object, current
+    def initialize ivar, object, current, cond: nil, command: nil
       @ivar = ivar.to_sym
       @object = object
-      @key = [:watch, @ivar].freeze
+      @key = [:watch, object.object_id, @ivar].freeze
```
st0012, line 334: WatchBreakpoint could be initialized for multiple instances, so we should take that into consideration when filtering duplicated bps.

lib/debug/breakpoint.rb
```diff
@@ -343,7 +346,10 @@ def watch_eval
         begin
           @prev = @current
           @current = result
-          suspend
+
+          if @cond.nil? || @object.instance_eval(@cond)
```
st0012, line 350: IMO, the if condition of WatchBreakpoint should also be evaluated on the target object.

PR #397, opened 2021-11-17
Implement Debugger Locals (`_raised_` and `_returned_`)

st0012:
I want to introduce a new concept called `Debugger locals`. The 2 I introduced in this PR are:

- `_raised_` - stores the last raised exception (after a catch breakpoint being triggered)
- `_returned_` - stores the last returned value (after a normal breakpoint being triggered on method return)

They're designed to help pass values between breakpoint and commands. For example:

- `catch Exception do: trace object _raised_` - pass caught exception to object tracer (super handy)
- `b FILE:line do:  p _returned_` - evaluate a method's return value without stopping the program 

They only exist in debugger's REPL and commands. And they **won't** affect user program's binding and are not accessible from user program.

Because it could be hard to correctly anticipate when those values are stored (especially `_returned_`), I think they should always exist with a default value `nil` instead of causing `NameError` when called unexpectedly.

Regarding the names, I think they can still be changed. But I don't want to use longer names like `raised_exception`, as it'll be too lengthy to type in the REPL. Also, since the variables have underscore for both prefix and postfix, I think the chance of this to happen is extremely low.

### DIFF
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -340,10 +340,20 @@ module DEBUGGER__
       frame_self.instance_eval(src)
     end
 
+    SPECIAL_LOCALS = [
+      [:raised_exception, "_raised_"],
+      [:return_value, "_returned_"],
+    ]
+
     def frame_eval src, re_raise: false
       @success_last_eval = false
 
       b = current_frame&.eval_binding || TOPLEVEL_BINDING
+      b = b.dup
+
+      SPECIAL_LOCALS.each do |m, local_name|
+        b.local_variable_set(local_name, current_frame.send(m))
+      end
 
       result = if b
                   f, _l = b.source_location
--- a/test/debug/debugger_local_test.rb
+++ b/test/debug/debugger_local_test.rb
@@ -1,0 +1,89 @@
+# frozen_string_literal: truk
+
+require_relative '../support/test_case'
+
+module DEBUGGER__
+  class DebuggerLocalsTest < TestCase
+    class RaisedTest < TestCase
+      def program
+        <<~RUBY
+     1| _rescued_ = 1111
+     2| foo rescue nil
+     3|
+     4| # check repl variable doesn't leak to the program
+     5| result = _rescued_ * 2
+     6| binding.b
+        RUBY
+      end
+
+      def test_raised_is_accessible_from_repl
+        debug_code(program) do
+          type "catch Exception"
+          type "c"
+          type "_raised_"
+          assert_line_text(/#<NameError: undefined local variable or method `foo' for main:Object/)
+          type "c"
+          type "result"
+          assert_line_text(/2222/)
+          type "c"
+        end
+      end
+
+      def test_raised_is_accessible_from_command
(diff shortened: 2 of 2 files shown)


lib/debug/thread_client.rb
```diff
@@ -340,11 +340,21 @@ def instance_eval_for_cmethod frame_self, src
...
+      [:raised_exception, "_raised_"],
+      [:return_value, "_returned_"],
+    ]
+
     def frame_eval src, re_raise: false
       begin
         @success_last_eval = false
 
         b = current_frame&.eval_binding || TOPLEVEL_BINDING
+        b = b.dup
```
ko1, line 353: why dup it?

lib/debug/thread_client.rb
```diff
@@ -340,11 +340,21 @@ def instance_eval_for_cmethod frame_self, src
...
+
+        SPECIAL_LOCALS.each do |m, local_name|
+          b.local_variable_set(local_name, current_frame.send(m))
```
ko1, line 356: do you need to define lvar if it is not provided?

lib/debug/thread_client.rb
st0012: I initially thought it'd pollute the program's binding if we use the binding directly. But after testing it doesn't seem to happen. I'll remove it.

lib/debug/thread_client.rb
st0012: in other tools like `pry`, accessing special variables (e.g. `_ex_`) when they don't exist it just return `nil` instead raising exceptions. I think this would be a nicer experience to have.
also, because `WatchBreakpoint` listens to both `*_return` (has return value) and `call` events, it'll be easier to script bp commands if it doesn't raise exceptions.

```
watch .... do: p _returned_
# instead of 
watch .... do: p _returned_ if: local_variables.include?(:_returned_)
```

lib/debug/thread_client.rb
st0012: well based on the CI result, it does pollute the program's binding
截圖 2021-11-25 10 18 45

ko1: Oh, I didn't know that `Binding#dup` can make another environment (even though I made it!!)

st0012:
@ko1 I've rebased the PR.

lib/debug/thread_client.rb
st0012: Haha, that's a great feature 😄

lib/debug/thread_client.rb
ko1: I don't agree because showing these variables is some kind of signals to show the special lines.
I'll merge this PR and change this points (and variable names).

I'll remove `#return` and use these variables.

```
(rdbg) i    # info command
%self = main
%return = 10
```

st0012: I'm not sure what the `#return` means here? so just `_raised_` and it'll be renamed as `%raised`?

ko1: sorry I meant removing `%return`  and show only `_return`

PR #405, opened 2021-11-21
Add path option to breakpoints

st0012:
After a discussion with @ko1, we decided to add `path:` option to breakpoint points. Specifically, they are:

- `MethodBreakpoint` - e.g. `break Class#method path: path_pattern`
- `CheckBreakpoint` - e.g. `break if: a == 1 path: path_pattern`
- `CatchBreakpoint` - e.g. `catch Exception path: path_pattern`
- `WatchBreakpoint` - e.g. `watch @ivar path: path_pattern`

When `path:` is provided, the breakpoint will only stop when all the conditions are fulfilled + the event that triggers it (method call, raised exception...etc.) happens at the path that matches the option.


test/support/extra_file_helper.rb
```diff
@@ -0,0 +1,15 @@
+require 'securerandom'
+
+module DEBUGGER__
```
ko1, line 3: how about to add to TestCase directly?

PR #409, opened 2021-11-24
fix typo

wonda-tea-coffee:
## Description
fix typo in document.


ono-max (approved):
Good catch!
Thank you.

PR #419, opened 2021-12-01
change the special local variables

ko1:
* `_returned_` -> `_return`
* `_raised_` -> `_raised`
* On `info` command, show `_...` instead of `%...`
* But if there is a local variable conflicts with special
  variables, use `%...` and do not introduce special lvars.

### DIFF
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -340,9 +340,9 @@ module DEBUGGER__
       frame_self.instance_eval(src)
     end
 
-    SPECIAL_LOCALS = [
-      [:raised_exception, "_raised_"],
-      [:return_value, "_returned_"],
+    SPECIAL_LOCAL_VARS = [
+      [:raised_exception, "_raised"],
+      [:return_value,     "_return"],
     ]
 
     def frame_eval src, re_raise: false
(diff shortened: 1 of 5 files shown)


ko1:
Another idea is `$_return`.

test/debug/info_test.rb
```diff
@@ -49,14 +49,32 @@ def test_info_prints_locals_by_default
...
+      3| end
+      4| foo :ret
+      RUBY
+
+      debug_code code do
+        type 'b 3'
+        type 'c'
+        assert_line_num 3
+        type 'i'
+        assert_line_text [/#return = :ok/, /_return = :ret/]
```
st0012, line 73: I thought you meant `%` instead of `#` in the description? I'm fine with either though.

lib/debug/thread_client.rb
```diff
@@ -340,9 +340,9 @@ def instance_eval_for_cmethod frame_self, src
       frame_self.instance_eval(src)
     end
 
-    SPECIAL_LOCALS = [
-      [:raised_exception, "_raised_"],
-      [:return_value, "_returned_"],
+    SPECIAL_LOCAL_VARS = [
+      [:raised_exception, "_raised"],
+      [:return_value,     "_return"],
```
st0012, line 345: I think it should be `returned` instead of `return` to present the passive form, similar to `raised`.
which means `return_value` should also be changed into `returned_value`.

ko1: In this timing, not returned yet.

PR #422, opened 2021-12-05
Add special local `_` to store previously evaluated value

st0012:
Specs:

- Like `irb`'s `_` variable, it stores the value returned by the
  previously evaluated expression in the REPL.
- The default value is `nil`.
- All the threads under the same session share the previously evaluated
  value. So when switching to other threads, you can use `_` to get the value
  returned by the previous thread.
- If the previous evaluation caused an exception, its result won't be stored.

### DIFF
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -87,6 +87,7 @@ module DEBUGGER__
       @obj_map = {} # { object_id => obj } for CDP
       @recorder = nil
       @mode = :waiting
+      @prev_evaled_value = nil
       set_mode :running
       thr.instance_variable_set(:@__thread_client_id, id)
 
@@ -358,6 +359,8 @@ module DEBUGGER__
         b.local_variable_set(name, var) if /\%/ !~ name
       end
 
+      b.local_variable_set(:_, @prev_evaled_value)
+
       result = if b
                   f, _l = b.source_location
                   b.eval(src, "(rdbg)/#{f}")
@@ -366,6 +369,8 @@ module DEBUGGER__
                   instance_eval_for_cmethod(frame_self, src)
                 end
       @success_last_eval = true
+
+      event! :sync_prev_evaled, result
       result
 
     rescue Exception => e
@@ -824,6 +829,8 @@ module DEBUGGER__
           end
 
           event! :result, result_type, result
+        when :store_prev_evaled
+          @prev_evaled_value = args[0]
         when :frame
           type, arg = *args
           case type
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -359,7 +359,10 @@ module DEBUGGER__
         b.local_variable_set(name, var) if /\%/ !~ name
       end
 
-      b.local_variable_set(:_, @prev_evaled_value)
+      # this check needs to be performed on the original binding
+      if !current_frame.binding&.local_variable_defined?(:_)
+        b.local_variable_set(:_, @prev_evaled_value)
+      end
 
       result = if b
                   f, _l = b.source_location
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -286,7 +286,10 @@ module DEBUGGER__
         end
 
         wait_command_loop tc
-
+      when :sync_prev_evaled
+        @th_clients.each_value do |tc|
+          tc << [:store_prev_evaled, ev_args[0]]
+        end
       when :dap_result
         dap_event ev_args # server.rb
         wait_command_loop tc
--- a/test/debug/debugger_local_test.rb
+++ b/test/debug/debugger_local_test.rb
@@ -161,5 +161,59 @@ module DEBUGGER__
         end
       end
     end
+
+    class UnderscoreTest < TestCase
+      def program
+        <<~RUBY
+     1| a = 10
+     2| b = 100
+     3| binding.b
+        RUBY
+      end
+
+      def test_underscore_returns_nil_by_default
+        debug_code(program) do
+          type "_"
+          assert_line_text(/nil/)
+          type "q!"
+        end
+      end
+
+      def test_underscore_returns_the_previous_value
+        debug_code(program) do
+          type "c"
+          type "a"
+          assert_line_text(/10/)
+          type "_ + 40"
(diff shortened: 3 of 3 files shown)


ko1:
I'm afraid to conflict with already existing parameter `_`.
Interestingly, `binding.irb` and `binding.pry` has same issue.

```
]$ ruby -r pry target.rb

From: /mnt/c/ko1/src/rb/ruby-debug/target.rb:3 Object#foo:

    2: def foo _
 => 3:   binding.pry
    4: end

[1] pry(main)> p _
nil
=> nil
[2] pry(main)> 42
=> 42
[3] pry(main)> _
=> 42
```

st0012:
I know this behavior. But from my experience, `_` in normal Ruby program usually means "won't be used value". So I think this problem is negligible in most cases and aligning user experience with `irb` and `pry` is more important. Take myself for example, because I switch between the debugger and Rails console everyday, I almost see this everyday too:

```
(rdbg) _    # ruby
eval error: undefined local variable or method `_' for main:Object
nil
```

And if we do get an issue report for this behavior, I think it'd also happen in `irb` too. We can then discuss an approach with the `irb` team and come up with a consistent approach to address it.

ko1:
I decided to postpone this feature.
Reasons:

* already existing `_` variable should not be overridden. debugger should be able to see all variables.
* Previous value of this patch is managed by ThreadClient object, so if the thread context is switched, the previous value is not "previous" in the literal sense.
* I want to introduce previous values like gcc:

```
(gdb) p 1
$1 = 1
(gdb) p 2+3
$2 = 5
(gdb) p $1 + $2
$3 = 6
```
`$[n]` is not acceptable on Ruby, so we need to invent new keywords.`$_[n]` (`$_1`, ...) is one idea but a few gems use this kind of global variables:

```
ko1@aluminium:~$ gem-codesearch '\$_\d+\b'
/srv/gems/ruby-compiler-0.1.1/vendor/ruby/test/ruby/test_rubyoptions.rb:      write_file("test-script", "$_0 = $0.dup; Process.setproctitle('hello world'); $0 == $_0 or Process.setproctitle('$0 changed!'); sleep 60")
/srv/gems/syntheticore-perlize-0.1.5/perlize.rb:            $_2 = arg if $_1
/srv/gems/syntheticore-perlize-0.1.5/perlize.rb:            $_1 = arg if $_1.nil?
/srv/gems/syntheticore-perlize-0.1.5/perlize.rb:            $__ = [$_1, $_2].compact.last
/srv/gems/syntheticore-perlize-0.1.5/perlize.rb:            $_1 = nil if $_2.nil?
/srv/gems/syntheticore-perlize-0.1.5/perlize.rb:            $_2 = nil
/srv/gems/syntheticore-perlize-0.1.5/perlize.rb:            $__ = [$_1, $_2].compact.last
```

Anyway, last value expression should be consistent with them.

st0012:
> already existing `_` variable should not be overridden. debugger should be able to see all variables.

Perhaps we can use the same approach for `_raised`? Which is not assigning the value when `_` is set by the user, but the rest of the time it'll be set. I think it's a balance between keeping the (probably few) user's program behavior while giving a consistent experience for most users. 

> Previous value of this patch is managed by ThreadClient object, so if the thread context is switched, the previous value is not "previous" in the literal sense.

I don't do concurrent programming often so I'm not sure what to expect in such case. But I imagine a shared value will be more complicated to manage than the current solution? If that's true, I'd go with the current approach and modify it when we receive more feedback.

> I want to introduce previous values like gcc:

I think it's a good idea. But I also hope to have `_` as well.

ko1:
> If that's true, I'd go with the current approach and modify it when we receive more feedback.

It is difficult to change such basic design.

st0012:
@ko1 ok, here's my plan to implement it:
1. To make the value shared, we need to keep it in `Session`.
2. The `_` variable needs be accessible in the frame binding to serve usages like `a = _ + 5`.
3. Given the above requirements, whenever a line of code is evaluated, the debugger needs to:
    1. Return the value to `Session` with `event!` method. This is simple.
    2. The session should then dispatch that value to all thread clients with a new `@tc <<` call.
    3. Before a `ThreadClient#frame_eval` is called, it needs to assign the value to the duplicated binding like we do for `SPECIAL_LOCALS`.

Wdyt?

st0012:
@ko1 I've addressed this 2 issues:

>  already existing _ variable should not be overridden. debugger should be able to see all variables.
> Previous value of this patch is managed by ThreadClient object, so if the thread context is switched, the previous value is not "previous" in the literal sense.

And regarding:

> I want to introduce previous values like gcc:

I think it could use more discussion and can be addressed in another PR.

lib/debug/thread_client.rb
```diff
@@ -366,6 +372,8 @@ def frame_eval src, re_raise: false
                   instance_eval_for_cmethod(frame_self, src)
                 end
       @success_last_eval = true
+
+      event! :sync_prev_evaled, result
```
ko1, line 376: Please send `SESSION'`s method. I'll make it.

st0012:
@ko1 if #503 is merged, we can use internal data channel and metadata to pass things around, which means the new events added here won't be needed. do you want to hold this PR's merge for that?

ko1:
Considering about this feature, there are same issues on multiple processes like ractors and we can not solve the issue.
On multi-process mode, the confusion can be bigger but I agree it is useful feature.

* inter-threads: share the value with `_`
* inter-ractors: share the value between a ractor? share the value if it is shareable? Q. what happens in another case?
* inter-process: Q. what happens?

For Q:
(1) Use the last value in a ractor/process
(2) Do not set `_` if the last value is not evaluated on different ractor/process

I need to check other debuggers (and similar systems)

st0012:
I don't have experience on debugging multi-process with a debugger, but I think it makes sense to not setting `_` in a new thread/ractor.

And regarding the implementation, I think we can use the same metadata/internal info approach I proposed in https://github.com/ruby/debug/pull/503. So after an expression is evaluated, its value will be sent back to the `SESSION` as part of the internal info. And every `ThreadClient` will receive the last-evaled value through commands' metadata. That way we won't need to introduce any new event/command types. Wdyt?

ko1:
Could you survey other REPL/debuggers?

PR #424, opened 2021-12-06
Fix REPL binding

st0012:
This PR fixes 2 issues:
- Currently, special local `_raised`  doesn't work in C method frames.
- Local variables assigned in REPL aren't stored in the binding.

### DIFF
--- a/lib/debug/frame_info.rb
+++ b/lib/debug/frame_info.rb
@@ -118,22 +118,6 @@ module DEBUGGER__
       "#{pretty_path}:#{location.lineno}"
     end
 
-    private def make_binding
-      __newb__ = self.self.instance_eval('binding')
-      self.local_variables.each{|var, val|
-        __newb__.local_variable_set(var, val)
-      }
-      __newb__
-    end
-
-    def eval_binding
-      if b = self.binding
-        b
-      elsif self.local_variables
-        make_binding
-      end
-    end
-
     def local_variables
       if lvars = self._local_variables
         lvars
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -352,7 +352,7 @@ module DEBUGGER__
     def frame_eval src, re_raise: false
       @success_last_eval = false
 
-      b = current_frame&.eval_binding || TOPLEVEL_BINDING
+      b = current_frame&.binding || TOPLEVEL_BINDING
 
       special_local_variables current_frame, binding: b do |name, var|
         b.local_variable_set(name, var) if /\%/ !~ name
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -355,7 +355,7 @@ module DEBUGGER__
       b = current_frame&.eval_binding || TOPLEVEL_BINDING
       b = b.dup
 
-      special_local_variables current_frame do |name, var|
+      special_local_variables current_frame, binding: b do |name, var|
         b.local_variable_set(name, var) if /\%/ !~ name
       end
 
@@ -442,10 +442,12 @@ module DEBUGGER__
 
     ## cmd: show
 
-    def special_local_variables frame
+    def special_local_variables frame, binding: nil
       SPECIAL_LOCAL_VARS.each do |mid, name|
         next unless frame&.send("has_#{mid}")
-        name = name.sub('_', '%') if frame.binding.local_variable_defined? name
+        binding ||= frame&.binding
+        next unless binding
+        name = name.sub('_', '%') if binding.local_variable_defined? name
         yield name, frame.send(mid)
       end
     end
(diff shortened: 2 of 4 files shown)


lib/debug/thread_client.rb
```diff
@@ -348,10 +348,9 @@ def instance_eval_for_cmethod frame_self, src
     def frame_eval src, re_raise: false
       @success_last_eval = false
 
-      b = current_frame&.eval_binding || TOPLEVEL_BINDING
-      b = b.dup
+      b = current_frame&.binding || TOPLEVEL_BINDING
 
-      special_local_variables current_frame do |name, var|
+      special_local_variables current_frame, binding: b do |name, var|
```
st0012, line 353: To evaluate in C frames, we should be able to fallback to `TOPLEVEL_BINDING`. So relying on the frame's binding is not enough here.

lib/debug/thread_client.rb
```diff
@@ -348,10 +348,9 @@ def instance_eval_for_cmethod frame_self, src
     def frame_eval src, re_raise: false
       @success_last_eval = false
 
-      b = current_frame&.eval_binding || TOPLEVEL_BINDING
-      b = b.dup
```
st0012, line 352: For some unknown reasons, I can't reproduce the polluted binding case anymore 🙈 . So I just remove the `.dup` call.

lib/debug/frame_info.rb
```diff
@@ -118,22 +118,6 @@ def location_str
       "#{pretty_path}:#{location.lineno}"
     end
 
-    private def make_binding
-      __newb__ = self.self.instance_eval('binding')
-      self.local_variables.each{|var, val|
```
st0012, line 123: This isn't correct as `local_variables` only yields keys but no values. But the fact that we didn't spot this probably means it's never called anyway.

lib/debug/thread_client.rb
st0012: Ah, it's because the current implementation only defines them at the frame that has those values. So it's not easy or even impossible to leak them into the program. 

For example, the `_return` is only defined at the end of the method definition, like 

```rb
def foo
 # ...
end # <= here
```

And theoretically no user program will be able to access it, so there's no need to worry about it polluting the program.

ko1: without `.dup`, `TOPLEVEL_BINDING` will be affected with assignments.
Maybe it is not intended.

lib/debug/thread_client.rb
```diff
@@ -344,29 +344,43 @@ def instance_eval_for_cmethod frame_self, src
...
                   frame_self = current_frame.self
                   instance_eval_for_cmethod(frame_self, src)
                 end
       @success_last_eval = true
+
+      # write non-special local variables back to the original binding
+      repl_binding.local_variables.each do |var|
+        next if SPECIAL_LOCAL_VAR_NAMES.include?(var)
+        original_binding.local_variable_set(var, repl_binding.local_variable_get(var))
+      end
```
st0012, line 382: After we dup the original binding, local variables changes made in the REPL won't be applied to it. So we need to manually propagate the values back to the original binding.

ko1: on master, it is propagated. which case do you considering?

```
[1, 5] in target.rb
=>   1| a = b = c = nil
     2| p a, b, c
     3|
     4|
     5| __END__
=>#0    <main> at target.rb:1
(rdbg) n    # next command
[1, 5] in target.rb
     1| a = b = c = nil
=>   2| p a, b, c
     3|
     4|
     5| __END__
=>#0    <main> at target.rb:2
(rdbg) eval a = b = c = :ok    # command
(rdbg) n    # next command
:ok
:ok
:ok
```

ko1: maybe you mean:

```
[1, 4] in target.rb
     1|
=>   2| a = 1
     3|
     4| __END__
=>#0    <main> at target.rb:2
(rdbg) y = 50    # ruby
50
(rdbg) y
eval error: undefined local variable or method `y' for main:Object
nil
```

but `y` is only valid on this debug subsession (when continue, the `y` will be gone). hmm

ko1: byebug's example.

```ruby
[master]$ byebug target.rb

[1, 10] in /mnt/c/ko1/src/rb/ruby-debug/target.rb
    1:
=>  2: a = 1
    3: b = 2
    4:
    5: __END__
    6:
    7: require 'foo'
    8:
    9: __END__
   10:
(byebug) y = 10
10
(byebug) p y
10
10
(byebug) n

[1, 10] in /mnt/c/ko1/src/rb/ruby-debug/target.rb
    1:
    2: a = 1
=>  3: b = 2
    4:
    5: __END__
    6:
    7: require 'foo'
    8:
    9: __END__
   10:
(byebug) y
*** NoMethodError Exception: undefined method `end_line=' for nil:NilClass
```

ko1:
I made #441 so please tell me if it is not enough.

st0012:
@ko1 I can confirm that #441 fixed both issues. thanks!

PR #428, opened 2021-12-08
rename `--util=init` to `--util=setup-autoload`

ko1:
`--util=init` is secret feature to help auto-loading debug.gem
without any additional wirting (like `gem debug` in Gemfile).

This fix is:

* renaming the command line option from `--util=init` to `--util=setup-autoload`
  to make clear the task.
* remove `--util=init -`.
* support zsh and fish.


ko1:
Thank you!

PR #433, opened 2021-12-14
Skip tests for DAP if the environmental variable is not set

ono-max:
To embed debug.gem in ruby 3.1, we have to make tests passes. Because the current test framework for DAP is not stable, we'll skip tests for DAP as an ad hock way.


st0012:
Perhaps we can add a new workflow for DAP tests? We can also add `continue-on-error: true` so when it failed the entire build would still be green.

ko1:
it will be tested on ruby/ruby so we need to stop the unstable DAP test ASAP.

st0012:
@ko1 yeah I understand that. my suggestion is to add a new job for the DAP tests (with the ENV enabled) so we can still get DAP tested on CI.

ono-max:
@st0012 
Sorry, for late reply.

> my suggestion is to add a new job for the DAP tests (with the ENV enabled) so we can still get DAP tested on CI.

I agree with your opinion. Let me create the PR to add a new job for DAP tests.

PR #438, opened 2021-12-16
support completion on DAP

ko1:
using `IRB::InputCompletor::retrieve_completion_data`.
This method doesn't return type information so maybe we need to
rewrite it.


ko1:
https://user-images.githubusercontent.com/9558/146324037-538e7d6e-3096-47dd-991d-50f492dd3383.mp4

st0012 (commented):
🙌 I'll use a similar approach to implement the REPL version.

PR #448, opened 2021-12-18
Use hidden iseq functions

ko1:


st0012:
Is this for performance gain?

ko1:
Yes. Not measured though.

PR #454, opened 2021-12-21
Unify internal path skipping logic

st0012:


ko1:
I'm not sure we can merge it. Now skipping `<internal:...>` locations is ad-hoc solution.

st0012:
I think it just means less code to remove once we want to get rid of the workaround?

ko1:
ok I'll revisit it.

PR #459, opened 2021-12-25
Add a new workflow to test for DAP

ono-max:


.github/workflows/ruby.yml
```diff
@@ -32,3 +32,8 @@ jobs:
         bundler-cache: true # runs 'bundle install' and caches installed gems automatically
     - name: Run tests
       run: bundle exec rake
+    - name: Run tests for DAP
```
st0012, line 35: This adds another step to all testing jobs. What you want to do may be adding `workflows/dap.yml`

ono-max:
@st0012 

Thanks for the review.

Rakefile
```diff
@@ -32,3 +31,12 @@ task :run => :compile do
   system(RbConfig.ruby, *%w(-I ./lib test.rb))
 end
 
+task :test_dap do
+  ENV['RUBY_DEBUG_DAP_TEST'] = '1'
```
st0012, line 35: This line can be moved under `Rake::TestTask.new(:test_dap) do |t|`

Rakefile
```diff
@@ -32,3 +31,12 @@ task :run => :compile do
...
+end
+
+Rake::TestTask.new(:test_dap) do |t|
+  t.libs << "test"
```
st0012, line 39: Based on what I tested in https://github.com/ruby/debug/pull/462, these `libs` assignment aren't necessary.

Rakefile
ono-max: Did you check if what you said work?

If I move `ENV['RUBY_DEBUG_DAP_TEST']` to Rake::TestTask.new(:test_dap) do |t|, it's always true even I run other rake task, though.

st0012: It can run `rake test_dap` but it looks like it'll also [leak the env var](http://dftrjy.blogspot.com/2019/01/rake-testtask-exports-environment.html).

st0012:
@ono-max an unrelated question: do you think you can write a guide on what the DAP payload means and how to write tests for the DAP server? I'd like to contribute to it and I'd love to have some pointers 😄

st0012 (approved):
The `test_dap` task is very handy 🚀

PR #461, opened 2021-12-26
Do not match any characters after the dot in an expression

### LINKED ISSUE #460: Debugger raise "NameError" when it hovers the mouse cursor over ".new" in VSCode
**Your environment**

* `ruby -v`: 
ruby 2.6.7p197 (2021-04-05 revision 67941) [x86_64-darwin20]

**Describe the bug**
When it hovers the mouse cursor over ".new", debugger raise "NameError" as follows:
```shell
$ exe/rdbg target.rb --open=vscode
DEBUGGER: Debugger can attach via UNIX domain socket (/var/folders/kv/w1k6nh1x5fl7vx47b2pd005w0000gn/T/ruby-debug-sock-501/ruby-debug-naotto-43763)
DEBUGGER: wait for debugger connection...
Launching: code /var/folders/kv/w1k6nh1x5fl7vx47b2pd005w0000gn/T/ruby-debug-vscode-20211226-43763-1vukogr/ /var/folders/kv/w1k6nh1x5fl7vx47b2pd005w0000gn/T/ruby-debug-vscode-20211226-43763-1vukogr/README.rb
DEBUGGER: Connected.
["DEBUGGER Exception: /Users/naotto/workspace/debug/lib/debug/thread_client.rb:986",
 #<NameError: wrong constant name Sample.new>,
 ["/Users/naotto/workspace/debug/lib/debug/server_dap.rb:813:in `const_defined?'",
  "/Users/naotto/workspace/debug/lib/debug/server_dap.rb:813:in `block (2 levels) in search_const'",
  "/Users/naotto/workspace/debug/lib/debug/server_dap.rb:812:in `all?'",
  "/Users/naotto/workspace/debug/lib/debug/server_dap.rb:812:in `block in search_const'",
  "/Users/naotto/workspace/debug/lib/debug/server_dap.rb:811:in `reverse_each'",
  "/Users/naotto/workspace/debug/lib/debug/server_dap.rb:811:in `search_const'",
  "/Users/naotto/workspace/debug/lib/debug/server_dap.rb:749:in `process_dap'",
  "/Users/naotto/workspace/debug/lib/debug/thread_client.rb:975:in `wait_next_action_'",
  "/Users/naotto/workspace/debug/lib/debug/thread_client.rb:676:in `wait_next_action'",
  "/Users/naotto/workspace/debug/lib/debug/thread_client.rb:284:in `suspend'",
  "/Users/naotto/workspace/debug/lib/debug/thread_client.rb:217:in `on_breakpoint'",
  "/Users/naotto/workspace/debug/lib/debug/breakpoint.rb:65:in `suspend'",
  "/Users/naotto/workspace/debug/lib/debug/breakpoint.rb:157:in `block in setup'",
  "target.rb:1:in `<main>'"]]
DEBUGGER: Disconnected.
/Users/naotto/workspace/debug/lib/debug/server_dap.rb:813:in `const_defined?': wrong constant name Sample.new (NameError)
	from /Users/naotto/workspace/debug/lib/debug/server_dap.rb:813:in `block (2 levels) in search_const'
	from /Users/naotto/workspace/debug/lib/debug/server_dap.rb:812:in `all?'
	from /Users/naotto/workspace/debug/lib/debug/server_dap.rb:812:in `block in search_const'
	from /Users/naotto/workspace/debug/lib/debug/server_dap.rb:811:in `reverse_each'
	from /Users/naotto/workspace/debug/lib/debug/server_dap.rb:811:in `search_const'
	from /Users/naotto/workspace/debug/lib/debug/server_dap.rb:749:in `process_dap'
	from /Users/naotto/workspace/debug/lib/debug/thread_client.rb:975:in `wait_next_action_'
	from /Users/naotto/workspace/debug/lib/debug/thread_client.rb:676:in `wait_next_action'
	from /Users/naotto/workspace/debug/lib/debug/thread_client.rb:284:in `suspend'
	from /Users/naotto/workspace/debug/lib/debug/thread_client.rb:217:in `on_breakpoint'
	from /Users/naotto/workspace/debug/lib/debug/breakpoint.rb:65:in `suspend'
	from /Users/naotto/workspace/debug/lib/debug/breakpoint.rb:157:in `block in setup'
	from target.rb:1:in `<main>'
```

**To Reproduce**
```ruby
class Sample
  def a
    p :a
  end
end

s = Sample.new
s.a

b = 1
c = 1
```

**Expected behavior**
It should not raise an error.

**Additional context**

ono-max:
Fixes #460.

The reason of this bug is that debugger matches characters after the dot when expressions are constant variables. That's why this PR ensure that debugger doesn't match any characters after the dot.

### DIFF
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -745,8 +745,8 @@ module DEBUGGER__
                   break false
                 end
               } and (message = "Error: Not defined global variable: #{expr.inspect}")
-            when /\A[A-Z]/
-              unless result = search_const(b, expr)
+            when /(\A[A-Z]\w*)/
+              unless result = search_const(b, $1)
                 message = "Error: Not defined constants: #{expr.inspect}"
               end
             else


lib/debug/server_dap.rb
```diff
@@ -745,8 +745,8 @@ def process_dap args
                   break false
                 end
               } and (message = "Error: Not defined global variable: #{expr.inspect}")
-            when /\A[A-Z]/
-              unless result = search_const(b, expr)
+            when /(\A[A-Z]\w*)/
```
st0012, line 748: just curious, why don't we use `frame_eval` to evaluate the `expr` directly but instead parsing them?

ono-max: `frame_eval` sometimes changes the value(FYI: https://github.com/ruby/debug/pull/395).

st0012: I think after https://github.com/ruby/debug/commit/756a2ed01d12a5fbdb81da829e48f159dd688ad6#diff-4208807b6db8555c61e42002b3e9c879003c5cfdb32d2b1a285a614b22fdf9d5 the changes made by `frame_eval` doesn't affect the original program anymore. So in this case it won't reproduce the issue again.

ono-max: > I think after 756a2ed#diff-4208807b6db8555c61e42002b3e9c879003c5cfdb32d2b1a285a614b22fdf9d5 the changes made by frame_eval doesn't affect the original program anymore. So in this case it won't reproduce the issue again.

Oh, I see. I didn't know that.
However, we can't use `frame_eval` for the following reasons:

- `frame_eval` uses current_frame, but we have to get a frame from request.
  - https://github.com/ruby/debug/blob/master/lib/debug/server_dap.rb#L715
- Even current_frame is the same as the target frame, we can't sometimes get the correct value such as instance of class. In the following case, showed value and value in backtrace information are not same.
  - Screen Shot 2021-12-27 at 9 47 56 AM

st0012: I don't get the example case. Are they different because they came from different frames? Or they're different but are evaluated from the same frame?
It looks like `search_const` may not benefit much from using `frame_eval`. But I think instance variables, global variables and normal expressions can be evaluated with `frame_eval`. And making `frame_eval` take arbitrary frame's binding shouldn't be an issue. 

I also see that the CDP server uses a similar but slightly different implementation for evaluation. So if we can utilize it in both server implementation, we can largely simplify these parts and make sure we're aligned on features (such as special locals).

ko1: We shouldn't use `eval` because expression can have side-effect.

PR #462, opened 2021-12-26
Clean up Rakefile

st0012:
1. Remove unnecessary task lib assignment.
2. Remove unnecessary declarations for `README.md` task.
3. Remove unused and broken `rake run` task.

```
❯ be rake run
# compile logs
/Users/st0012/.rbenv/versions/2.7.5/bin/ruby: No such file or directory -- test.rb (LoadError)
```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -2,8 +2,6 @@ require "bundler/gem_tasks"
 require "rake/testtask"
 
 Rake::TestTask.new(:test) do |t|
-  t.libs << "test"
-  t.libs << "lib"
   t.test_files = FileList["test/**/*_test.rb"]
 end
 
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -20,15 +18,9 @@ end
 
 task :default => [:clobber, :compile, 'README.md', :test]
 
-file 'README.md' => ['lib/debug/session.rb', 'lib/debug/config.rb',
```
ko1, line 23: why do you remove this dependencies?

st0012: Sorry that I didn't notice it's for checking timestamps 🤦‍♂️ I've dropped this change.

PR #482, opened 2022-01-06
Fix breakpoints' path option

st0012:
It should distinguish if path is a string or a regexp and match them with the path in different ways.


ko1:
Could you update doc too?

st0012:
@ko1 updated 👍

PR #495, opened 2022-01-16
Support command on CheckBreakpoint

st0012:
1. `break if: cond` doesn't support commands like `do:` or `pre:`. This PR adds the support.
2. To make missing option support noticeable in the future, I've centralized `Breakpoint`'s options processing to the parent class as keyword arguments. This requires the options (`path`, `command` and `cond` for now) to be explicitly provided. It should make missing support of them more noticeable.

### DIFF
--- a/lib/debug/breakpoint.rb
+++ b/lib/debug/breakpoint.rb
@@ -317,10 +317,10 @@ module DEBUGGER__
   end
 
   class CheckBreakpoint < Breakpoint
-    def initialize expr, path, command
-      @expr = expr.freeze
+    def initialize cond:, command: nil, path: nil
+      @cond = cond
       @command = command
-      @key = [:check, @expr].freeze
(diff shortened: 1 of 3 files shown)


lib/debug/breakpoint.rb
```diff
@@ -8,9 +8,13 @@ class Breakpoint
 
     attr_reader :key
 
-    def initialize do_enable = true
+    def initialize cond:, command:, path:, do_enable: true
```
ko1, line 11: could you use optional keywords?

st0012: Since all types of breakpoints should support those options by default, I think asking subclasses to explicitly pass them is a way to prevent similar issues in the future.

ko1: if it is mandatory, please use mandatory parameters.

PR #500, opened 2022-01-22
Mention Ruby version requirement in Readme

st0012:
I got feedback from Rubyists saying that they thought the debugger only works with Ruby 3.1. Given that it's mostly advertised with the Ruby 3.1 release, I can understand why they make that assumption. So I think we can mention the supported Ruby version and platform in the readme to make it clearer.

### DIFF
--- a/misc/README.md.erb
+++ b/misc/README.md.erb
@@ -2,7 +2,7 @@
 
 # debug.rb
 
-This library provides debugging functionality to Ruby.
+This library provides debugging functionality to `Ruby MRI 2.6+`.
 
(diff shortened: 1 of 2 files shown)


misc/README.md.erb
```diff
@@ -2,7 +2,7 @@
 
 # debug.rb
 
-This library provides debugging functionality to Ruby.
+This library provides debugging functionality to `Ruby MRI 2.6+`.
```
ko1, line 5: why use back quote?

```suggestion
This library provides debugging functionality to Ruby (MRI) 2.6 and later.
```

PR #465, opened 2021-12-28
Unify frame and locals collection implementation between ThreadClient and servers

st0012:
1. Use `get_frame` to retrieve frames with the given frame index.
2. Use `collect_locals` to retrieve the locals from the given frame.
    - This also fixes an issue that `%self` isn't displayed in DAP when stepping backward. 

Thanks to the DAP tests written by @ono-max, I feel a lot comfortable doing such refactoring now 👍

Issue #305, opened 2021-09-26
Proposal - A Built-in Text-based Debugger UI


st0012:
I want to propose a built-in Text UI that's similar to `GDB`'s TUI, which has one key feature: having multiple windows to display different information at the same time. For example:
  - Variables
  - Source code
  - Backtrace
  
Also, it may be able to display `debug.gem`-specific output as well, like:
- Information about replay debugging
- Tracer output 

And windows should support these operations:
- Focus on a certain window and scroll its content 
- Configurable information

As for features like key-bindings, layout, custom color schemes...etc. They'll be nice to have but probably aren't necessary in the first few versions.

(In the long-term, I hope the UI can be close to [ruby_jard's](https://rubyjard.org/) or provide APIs for community-based UIs)

### Usage

By default the debugger UI shouldn't be started. But users can use it with:

- `tui on|off` to open/close the UI.
  - `GDB` uses `tui enable|disable` but we already use `on|off` for `trace` and `record` commands.
- `tui focus <num>` to focus on a window.
- `tui up|down <lines>` for primitive scrolling.
- A new configuration option `RUBY_DEBUG_ACTIVATE_UI` - when it's set the UI will be activated whenever a breakpoint is triggered.

### Why do we need a built-in UI while there's already DAP support?

- **It'll be accessible to all Ruisysts**:  ATM, VSCode (used by 30~40% of the Rubyists) is the only editor that can work with this debugger via DAP. This means users of other editors can't access any debugger UI powered by this project. So if there's an UI bundled with this gem, all users will be able to use it.
  - It's not easy to support debugger UI for non-vscode editors atm. Because it usually requires:
    -  DAP support from editor plugins (for vim users, it relies on plugins like [vimspector](https://github.com/puremourning/vimspector)).
    - Extra work to connect the DAP plugin and [vscode-rdbg](https://github.com/ruby/vscode-rdbg). For example, https://github.com/ruby/vscode-rdbg/issues/7
  - (Editor usage source: [Rails survey 2020](https://rails-hosting.com/2020/#os-editors-servers)).
- **It'll require zero or a minimum setup**: Event if all major editors support DAP and work with this debugger, it'll still require different levels of setup + learning the editor's debugger commands to activate the debugger UI. But the built-in UI won't require any or much less of them.

### When don't we let the community build it?

[...]

PR #503, opened 2022-01-27
Text-based UI support

st0012:
This PR adds the TUI support proposed in #305. 

## Usage

- `tui` opens TUI with a default top window (locals)
- `tui on [type]`  opens TUI with a `type` top window (current has `locals` and `src`)
  - `tui on src,locals` can open 2 windows
- `tui off` closes TUI

## Demo

**Single-window**
(The top window shows the current frame's local variables)

https://user-images.githubusercontent.com/5079556/151341747-e0e88185-e18d-4444-b914-81d5bb987871.mov

**Multi-window**

https://user-images.githubusercontent.com/5079556/152705616-021f881c-2801-4697-9bd2-06d0144ee44f.mov

## Specs

These specs will be improved in the future.

**When disabled (default)**

It doesn't affect any existing funcionality.

**When enabled with `tui` command**

- The screen will be separated into top frame and the REPL
  - The top frame
    - Is not scrollable
    - Has one window to display current frame's source code (`src`) or local variables (`locals`).
    - Has a fixed height of 15 lines (13 lines for content)
      - I'll make it configurable once this has been accepted
  - The REPL
    - Is not scrollable either
- Can be disabled with `tui off`

### More window types

- `breakpoints` - the list of registered breakpoints
- `tracers` - the list of activated tracers
- `trace` - tracers' output will be directed here
- `threads`
- `frames` - nearby frames (it's unlikely we'll be able to show all frames)

### Multi-window support

- `tui on [w1:h1,w2:h2]` opens windows with specified types. We'll start from stack layout. Example: `tui on src:20,locals:10`

```
----------
|  src   |
|        |
----------
| locals |
----------
|  REPL  |
----------
```

- `RUBY_DEBUG_TUI_WINDOWS=w1:h1,w2,h2` - configures default windows.
- `RUBY_DEBUG_TUI_HEIGHT=15` - configures the default window height.

### Scroll support

We can add scroll support to both the REPL and the top frame. But probably only command or key-binding scrolling at first.

### Additional data channel

TUI requires frequent data exchange between the Session and ThreadClient. More specifically:
  1. Session needs to tell TC what type of the data it needs with additional information like how many lines is expected.
  2. TC needs to give the Session requested TUI data, like local variables.
 
For 2), we already pass `@internal_info` in every `TC -> Session` events so we can use it directly.

However, we currently don't have anything for 1). So in this PR I added change to let every `Session -> TC` request also carry a `metadata` payload. It's a Hash that can carry some general information, like TUI info. I think this can also help simplify the implementation of https://github.com/ruby/debug/pull/422.

### Separate frame data's collection/display logic

Currently, frame data like file source, local variables...etc. are printed during collection. This is enough for the current local console but is not convenient when we need to process the data differently, like [DAP server](https://github.com/ruby/debug/blob/089ae90a5c556d7b3a6f57ff9cbc07f9f35b33b3/lib/debug/server_dap.rb#L642-L662) or the TUI. So I also extracted `get_*` methods (return collections) from some of the `show_*` methods (show collections). Part of this change was done in #465 and merged here and can be merged it separately.

(I know technically `puts` in ThreadClient doesn't "print" the data immediately. But it's hard to get the collection back from the `@output` variable as a collection anyway.)

### TODO: Reduce screen refreshing

Currently, screen is refreshed on each `UI#puts` call. So `lines.each {|l| @ui.puts(l)}` will refresh screen multiple times. If we can change that into `@ui.puts(lines)`, we can avoid unnecessary refreshes. It won't be a huge change (mainly [here](https://github.com/ruby/debug/blob/master/lib/debug/session.rb#L206) and maybe a few other places), but I want to wait for the general design is approved before making further changes.

### DIFF
--- a/lib/debug/local.rb
+++ b/lib/debug/local.rb
@@ -52,6 +52,10 @@ module DEBUGGER__
       end
     end
 
+    def puts_internal_test_info(internal_info)
+      $stdout.puts("INTERNAL_INFO: #{internal_info}")
+    end
+
     def puts str = nil
       case str
       when Array
@@ -103,6 +107,216 @@ module DEBUGGER__
           end
         end
       }
+    end
+  end
+
+  class UI_LocalTuiConsole < UI_LocalConsole
+    attr_reader :screen
+
+    def initialize(windows)
+      @screen = Screen.new(width, height, windows)
+      super()
+    end
+
+    def deactivate
+      super
+      clear_screen!
+    end
+
+    def height
+      if (w = IO.console_size[0]) == 0 # for tests PTY
+        80
+      else
+        w
+      end
+    end
+
+    def tui?
+      true
+    end
+
+    def store_prev_line(line)
+      @prev_line = line
+    end
+
+    def store_tui_data(data)
+      @ui_data = data
+    end
+
+    def puts str = nil
+      @screen.draw_windows(@ui_data)
+
+      if @prev_line
+        @screen.draw_repl(@prev_line)
+        @prev_line = nil
+      end
+
+      @screen.draw_repl(str)
+      @screen.render!
+    end
+
+    def windows_metadata
+      @screen.windows_metadata
+    end
+
+    def clear_screen!
+      @screen.clear_screen!
+    end
+
+    class Screen
+      attr_reader :width, :height, :windows, :repl
+
+      def initialize(width, height, windows)
+        @height = height
+        @width = width
+        @windows = windows
+        # we need to leave 1 line for the input and 1 line for overflow buffer
+        @repl = REPL.new("REPL", @width, height - windows.sum(&:height) - 2)
+        @windows << @repl
+      end
+
+      def render!
+        clear_screen!
+        @windows.each { |w| w.render!($stdout) }
+      end
+
+      def draw_windows(data)
+        @windows.each { |w| w.draw(data) }
+      end
+
+      def draw_repl(str)
+        case str
+        when Array
+          str.each{|line|
+            @repl.puts line.chomp
+          }
+        when String
+          str.each_line{|line|
+            @repl.puts line.chomp
+          }
+        when nil
+          @repl.puts
+        end
+      end
+
+      def windows_metadata
+        @windows.each_with_object({}) do |window, metadata|
+          metadata[window.name] = { width: window.content_width, height: window.content_height }
+        end
+      end
+
+      def clear_screen!
+        $stdout.print("\033c")
+      end
+    end
+
+    class Window
+      attr_reader :name, :width, :height, :content_width, :content_height
+
+      def initialize(name, width, height)
+        @name = name
+        @width = width
+        @content_width = @width
+        @height = height
+        @content_height = @height
+        @lines = []
+      end
+
+      def puts line = ""
+        if @lines.length >= @content_height
+          @lines.slice!(0)
+        end
+
+        # if the current line will break into multiple lines, we need to skip more lines for it. otherwise it'll affect the overall layout
+        if line.length > @content_width
+          n_lines = (line.length/@content_width.to_f).ceil
+
+          start = 0
+          n_lines.times do
+            if @lines.length >= @content_height
+              @lines.slice!(0)
+            end
+
+            @lines << line[start..(start + @content_width - 1)]
+            start += @content_width
+          end
+        else
+          @lines << line
+        end
+      end
+
+      def draw(data)
+        window_data = data[name]
+        return unless window_data
+
+        case window_data
+        when String
+          puts(window_data)
+        when Array
+          window_data.each do |d|
+            puts(d)
+          end
+        else
+          raise "unsupported window data type for window [#{name}]: #{data.class} (#{data})"
+        end
+      end
+
+      def render!(io)
+        @lines.each do |line|
+          io.puts(line)
+        end
+      end
+
+      def clear
+        @lines.clear
+      end
+    end
+
+    class FramedWindow < Window
+      def initialize(name, width, height)
+        super
+        # border and padding
+        @content_width = @width - 2
+        # top and bottom border
+        @content_height = @height - 2
+      end
+
+      def render!(io)
+        io.puts(top_border)
+
+        @content_height.times do |i|
+          if line = @lines[i]
+            io.puts("┃ " + line)
+          else
+            io.puts("┃")
+          end
+        end
+
+        io.puts(bottom_border)
+      end
+
+      def draw(data)
+        # framed windows doesn't preserve data
+        clear
+        super
+      end
+
+      private
+
+      def top_border
+        @top_border ||= begin
+          head = "┏━━" + " (#{@name}) "
+          tail = "━" * (@width - head.length)
+          head + tail
+        end
+      end
+
+      def bottom_border
+        @bottom_border ||= "┗" + "━" * (@width - 1)
+      end
+    end
+
+    class REPL < Window
     end
   end
 end
--- a/lib/debug/local.rb
+++ b/lib/debug/local.rb
@@ -114,13 +114,33 @@ module DEBUGGER__
     attr_reader :screen
 
     def initialize(windows)
-      @screen = Screen.new(width, height, windows)
       super()
+
+      @original_stdout = $stdout.dup
+      @original_stderr = $stderr.dup
+      @screen = Screen.new(width, height, windows, @original_stdout)
+
+      tui_console = self
+      original_stdout = @original_stdout
+      original_stderr = @original_stderr
+      Reline.output = original_stdout if defined?(Reline)
+
+      $stdout.define_singleton_method(:write) do |*args|
+        original_stdout.write(*args)
+        tui_console.puts(args.join)
+      end
+
+      $stderr.define_singleton_method(:write) do |*args|
+        original_stderr.write(*args)
+        tui_console.puts(args.join)
+      end
     end
 
     def deactivate
       super
       clear_screen!
+      $stdout.reopen(@original_stdout)
+      $stderr.reopen(@original_stderr)
     end
 
     def height
@@ -141,6 +161,10 @@ module DEBUGGER__
 
     def store_tui_data(data)
       @ui_data = data
+    end
+
+    def puts_internal_test_info(internal_info)
+      @original_stdout.puts("INTERNAL_INFO: #{internal_info}")
     end
 
     def puts str = nil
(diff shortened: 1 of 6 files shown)


st0012:
@ko1 this implementation requires some extractions made in https://github.com/ruby/debug/pull/465, so it'd be great if we can finish its discussion.

ono-max:
I've never seen `Expected the debuggee program to finish.` as random errors.

https://github.com/ruby/debug/runs/5010975029?check_suite_focus=true

st0012:
@ono-max Yeah it's weird but I don't see how this PR can cause the detaching test failure. The DAP-related changes are:

- Method extractions like `get_frame` and `collet_locals`, which are taken from https://github.com/ruby/debug/pull/465.  But I don't see that PR's tests failed. And it'd be weird for them to affect detaching as well.
- Replacing `tc <<` with the new `request_tc` interface. Since this is a low level change, if it has any issue it'll fail all DAP tests instead of just the one.

ko1:
I read your patch.

* we can introduce metadata or some more communication path between Session <-> TC
* Also we can accept to introduce getters instead of printers (show_... methods).
* Now your patch hard codes TUI mode into sources, and I don't think it is good idea for maintenance. So current patch is difficult to accept. Could you design extended general (extensible) mechanism at first?

Others:

* it is only for locals. Can we extend it on remote mode? I don't want to introduce difference between local and remote REPL as possible.
* Where the stdout/err outputs?

st0012:
> it is only for locals. Can we extend it on remote mode? I don't want to introduce difference between local and remote REPL as possible.

I can explore it and it should be possible. But because it requires redrawing the screen repeatedly, the result won't be good in remote consoles under current implementation (with bigger overhead to transfer data). I'll prefer to wait until optimizations like `Reduce screen refreshing` is implemented.

Another reason to wait would be to have a more complete test framework for TUI.

> Where the stdout/err outputs?

Do you mean we should preserve the stdout/err output after the debug session is finished?

ko1:
> Do you mean we should preserve the stdout/err output after the debug session is finished?

My question is, what happens when the program prints something to stdout/err?

st0012:
> My question is, what happens when the program prints something to stdout/err?

Currently, program's output to stdout/err won't be displayed because they don't trigger the screen refresh. I'm still finding the least impactful way to pipe program output to the console screen.

st0012:
@ko1 I've added stderr/stdout support. I've tested these scenarios in the demo:

- stdout from user program
- stdout from debugger console
- stderr from debugger console
- stdout resumed after TUI is turned off

https://user-images.githubusercontent.com/5079556/159140757-713d7724-1489-4dac-9df7-fd0fa399f024.mov

lib/debug/local.rb
```diff
@@ -105,5 +109,240 @@ def after_fork_parent
...
+      $stderr.define_singleton_method(:write) do |*args|
+        original_stderr.write(*args)
+        tui_console.puts(args.join)
+      end
+    end
+
+    def deactivate
+      super
+      clear_screen!
+      $stdout.reopen(@original_stdout)
```
st0012, line 142: This will reconnect the `$stdout` object to the original stdout object's stream. So both `STDOUT` and `$stdout` will be resumed.

lib/debug/local.rb
```diff
@@ -105,5 +109,240 @@ def after_fork_parent
...
+
+      tui_console = self
+      original_stdout = @original_stdout
+      original_stderr = @original_stderr
+      Reline.output = original_stdout if defined?(Reline)
+
+      $stdout.define_singleton_method(:write) do |*args|
+        original_stdout.write(*args)
+        tui_console.puts(args.join)
+      end
```
st0012, line 131: This is to listen to stdout's printing events and reflect the message on the TUI window as well. The patch will be cleared after `#deactivate` is called (see below).

lib/debug/local.rb
```diff
@@ -105,5 +109,240 @@ def after_fork_parent
...
+    def initialize(windows)
+      super()
+
+      @original_stdout = $stdout.dup
+      @original_stderr = $stderr.dup
+      @screen = Screen.new(width, height, windows, @original_stdout)
+
+      tui_console = self
+      original_stdout = @original_stdout
+      original_stderr = @original_stderr
```
st0012, line 125: Locals are for referencing from `define_singleton_method` (can't use ivar there), and ivars are for referencing from `#deactivate`.

lib/debug/local.rb
```diff
@@ -105,5 +109,240 @@ def after_fork_parent
...
+      else
+        w
+      end
+    end
+
+    def tui?
+      true
+    end
+
+    def store_prev_line(line)
```
st0012, line 158: This is mainly for storing the user input line.

lib/debug/session.rb
```diff
@@ -253,7 +253,8 @@ def process_event evt
         else
           tc << [:eval, :display, @displays]
         end
-
+      when :show
+        tc << [:show, :list, {dir: -1}]
```
ko1: who use it?

lib/debug/thread_client.rb
```diff
@@ -407,57 +407,68 @@ def frame_eval src, re_raise: false
...
+        []
       end
     rescue Exception => e
       p e
       pp e.backtrace
       exit!
     end
 
+    def show_src(frame_index: @current_frame_index, update_line: false, max_lines: CONFIG[:show_src_lines] || 10, **options)
+      return unless @target_frames
```
ko1, line 486: `if frame = get_frame(frame_index)` ?

lib/debug/session.rb
st0012: sorry I'm not sure where this line is. I can't find it in the latest changes.

lib/debug/thread_client.rb
st0012: 👍 will update

lib/debug/session.rb
st0012: this is now marked as outdated. I guess GH displayed some stale data when you left the comment? I think it's from an earlier implementation.

ko1:
* what the plan do you want to extend it with more information?
* do you plan to extend to accept key combination like gdb's TUI mode?
* Could you squash commits if the commits can be merged semantically? It will help me to review them.
* I couldn't review all code, so I need to more time for it.

Minor comments:

I tried with the script:

```ruby
# target.rb

def rec n
  rec n - 1 if n > 0
  a = b = c = d = e = f = g = h = i = j = k = l = m = n = o = p = q = r = s = t = u = nil
  bb
end

rec 100
```

* I can't see the src pane. any setting is needed? I can see the lvars pane and prompt
* When there are many local variables, it seems to show vars as possible.

image

* when the window size is changed, the output seems broken.

image

* I didn't use `get_xxx` terminology in `thread_client.rb` so if we have another good name, it seems better (but I couldn't find out another name).

st0012:
First of all, thank you for taking the time reviewing this PR 🙏 

> * what the plan do you want to extend it with more information?

at the very least, I hope we support these:

- `breakpoints` - the list of registered breakpoints
- `tracers` - the list of activated tracers
- `trace` - tracers' output will be directed here
- `threads`
- `frames` - nearby frames (it's unlikely we'll be able to show all frames)

(I've also listed them in the description)

and users should be able to use `tui on threads,src` to open TUI with the selected windows.

> * do you plan to extend to accept key combination like gdb's TUI mode?

yes. I'm not sure how to prioritize it with vertical window split though. I think we need some discussion around it.

> * Could you squash commits if the commits can be merged semantically? It will help me to review them.

of course. will let you know when done.

> I tried with the script:
> 
> ```ruby
> # target.rb
> 
> def rec n
>   rec n - 1 if n > 0
>   a = b = c = d = e = f = g = h = i = j = k = l = m = n = o = p = q = r = s = t = u = nil
>   bb
> end
> 
> rec 100
> ```
> 
> * I can't see the src pane. any setting is needed? I can see the lvars pane and prompt

you can use `tui on src` to switch to src pane, or `tui on locals,src` to open both panes.

> * When there are many local variables, it seems to show vars as possible.
> 
> image

do you mean it doesn't show as many locals as possible? that's because currently the window height is fixed.

I think the best way to address this issue is to have a key-binding for scrolling. I'll write a plan for that.

> * when the window size is changed, the output seems broken.

yes that's a known issue. 

> * I didn't use `get_xxx` terminology in `thread_client.rb` so if we have another good name, it seems better (but I couldn't find out another name).

how about `collect_xxx`?

st0012:
@ko1 I've reduced the number of commits to just 4. and the first 2 commits can actually be merged separately (that'll help reducing conflicts).

ko1:
> (I've also listed them in the description)

Sorry I missed it.

>> do you plan to extend to accept key combination like gdb's TUI mode?
> yes. I'm not sure how to prioritize it with vertical window split though. I think we need some discussion around it.
...
> I think the best way to address this issue is to have a key-binding for scrolling. I'll write a plan for that.

Yes I'm considering how to scroll the window to access more information like editors. If the window is small and information are many (for example ivar, frames, srcs) the importance of scroll is high.

st0012:
@ko1 I've checked Ruby TUI applications like [Jard](https://github.com/nguyenquangminh0711/ruby_jard) and [Ruco](https://github.com/grosser/ruco) (which is built on top of [dispel](https://github.com/grosser/dispel), a [curses](https://github.com/ruby/curses) replacement) to see how they handle keyboard input. And I think to support key-binding, we can't use `reline`'s `readline` method anymore and need to implement a lower level of input handling. This can affect how features like autocompletion are supported under different modes.

IMO, the principles should be

- We need to support the same set of input features (e.g. hints and autocompletion) in normal/TUI mode.
- They **can't** have on different implementations (`reline` vs a complete custom one) for the sake of maintenance.
- For whatever implementation we choose, it shouldn't rely on any 3rd party gem or C-library.
    - So both [dispel](https://github.com/grosser/dispel) and [curses](https://github.com/ruby/curses) are off of the table.

That leaves us with these options:

1. Use custom implementation to replace `reline` for both normal and TUI mode.
    - This clearly requires a lot of work.
2. Support key-binding detection in `reline` that also works with the `readline` method.
    - I think it's the ideal option but I'm not sure if that's in the scope of `reline`.

Do you have any thoughts on this?

ko1:
3. Use custom readline on TUI mode? (simple one)

Controlling terminal is very difficult task so I don't recommend you to dive into this area.
Also maybe I can't accept complex solution because I don't want to maintain it.

I think it is  more reasonable to make a CLI DAP client in another gem and connect with debug.gem.

st0012:
I agree that the debugger shouldn't maintain complicated terminal controlling mechanism. I'm not too keen to dive into that now either.

And I just found that GDB uses `ncurses` to implement TUI instead of building everything from scratch. So perhaps bringing the `curses` gem in would be a possible solution here.

But if that's not acceptable, I don't think building another UI on top of DAP is what I'd pursuit. My goal for this TUI are:

- Convenient to use: just a `tui` command can spin it up.
- Operations are still performed from REPL and can directly see stdout/stderr.

Which would be hard to achieve with DAP's architecture.

st0012:
@ko1 Since the chance of this PR getting merged is small, I won't proceed on it anymore. Do you think the first 2 refactoring commits worth merging? Otherwise I'll just close it.

ko1:
Could you make another PR for that?

st0012:
@ko1 Of course, here's the PR: https://github.com/ruby/debug/pull/628

PR #504, opened 2022-01-27
Proposal feature tilde instruction

kaiquekandykoga:
Hello guys,

Debugger has reserved words.

If the script being debugged has a variable called list it is not possible to visualize its value typing list.

The idea is to use a new ~ instruction to run code without considering debugger features.

If a variable list is initialized with value "Hello World" for instance.

After.

```
(rdbg) ~ list
"Hello World"

(rdbg) ~ list = "Hello debugger"
"Hello debugger"

(rdbg) list
     1| require 'debug'
     2| list = "Hello world"
=>   3| binding.b
```

Before.

```
(rdbg) ~ list
eval error: undefined method `~' for "Hello World":String
  (rdbg)/example.rb:1:in `<main>'
nil

(rdbg) ~ list = "Hello debugger"
eval error: undefined method `~' for "Hello debugger":String
  (rdbg)/example.rb:1:in `<main>'
nil

(rdbg) list
     1| require 'debug'
     2| list = "Hello World"
=>   3| binding.b
```

Thank you very much.

### DIFF
--- a/README.md
+++ b/README.md
@@ -521,6 +521,11 @@ You can use the following debug commands. Each command should be written in 1 li
 The `[...]` notation means this part can be eliminate. For example, `s[tep]` means `s` or `step` are valid command. `ste` is not valid.
 The `<...>` notation means the argument.
 
+### Instructions without reserved words
+
+* `~ <instructions>`
+  * Run instructions without debugger reserved words.
+
 ### Control flow
 
 * `s[tep]`
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -399,6 +399,14 @@ module DEBUGGER__
       # p cmd: [cmd, *arg]
 
       case cmd
+
+      ### Instructions without reserved words
+
+      # * `~ <instructions>`
+      #   * Run instructions without debugger reserved words.
+      when '~'
+        @tc << [:eval, :pp, arg || '']
+
       ### Control flow
 
       # * `s[tep]`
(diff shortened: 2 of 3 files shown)


ko1:
`p list` doesn't work?

kaiquekandykoga:
Hello @ko1 ,

Thank you for the fast response.

Printing the variable works, but there is also another scenario where I have not found a way to set a value.

Sorry, I forgot to mention this last one.

```
(rdbg) p list    # command
=> "Hello World"
(rdbg) list = ""    # command
Can not handle list argument: = ""
```

st0012:
I've also hit the same problem multiple times before. I think the debugger should at least allow `list = ""` without using `eval` or `p`. There is only one command `config <name> = <val>` that uses `=` so other cases should be able to use `=` to determine the input is Ruby code.

ko1:
we have `eval` or `e` command.

and also "# command" will help you to recognize it is debugger command or ruby expression.

st0012:
The `# command` hint helps. But since I've been using the debugger for a while, I don't always notice the change of it.

IMO, the core of this proposal is that users may expect the debugger to distinguish the command/expression better. If you say `list` or `list something` can be either an expression or a debugger command and we need to use `p` or `e` to avoid conflict, everybody can understand it. But at this moment, `list = something` is obviously an expression, so I'd expect the debugger can distinguish it.

As a user, I hope the debugger can allow me to type Ruby expressions freely whenever possible. And in the assignment case, the implementation effort should be small. So if that can allow users type more Ruby expressions without the need of `p` or `e`, I think it's worth doing it? Unless we're preserving this pattern for extensibility?

ko1:
I don't think so. Such special cases introduces confusion.

Recently I found `; expr` is evaluated as a Ruby expression, and it is similar to original proposal (`~ expr`).

Issue #502, opened 2022-01-25
`break` should respect current frame.


ko1:
**Your environment**

* `ruby -v`: 2.7
* `rdbg -v`: 1.4

**Describe the bug**

When the current frame is not a top-frame, but `b <line>` makes a breakpoint on the top frame.

**To Reproduce**

```ruby
# t1.rb

require_relative 't2'

foo

p 1
p 2
p 3

# t2.rb
def foo
  
end
```

```
$ exe/rdbg target.rb -e 'b t2.rb:3'
(rdbg) up    # command
=>   4| foo
(rdbg) b 7    # break command
#1  BP - Line (pending)  t2.rb:7     <--- it should be t1.rb:7
```

PR #505, opened 2022-01-27
Avoid manually setting ThreadClient's location

st0012:
The cause of #502 is frame operations don't set the `ThreadClient`'s location. So the `break` command gets the incorrect location when adding new breakpoints after the current frame is changed.

I think the best solution to this problem is to avoid manual location assignment altogether. Since there aren't any case that we want the TC's location be different from its current frame, delegating the location to its current frame's should be the simplest solution.


ko1:
it should be a problem on ractors, but I'll revisit it.

PR #507, opened 2022-01-28
Improve help

kaiquekandykoga:
Hello guys,

Colorize h[elp].

Write category with h[elp] using arguments the same way is already done with h[elp] without arguments.

Thank you very much.

### DIFF
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -81,6 +81,8 @@ module DEBUGGER__
   class Session
     attr_reader :intercepted_sigint_cmd, :process_group
 
+    include Color
+
     def initialize ui
       @ui = ui
       @sr = SourceRepository.new
@@ -1002,11 +1004,7 @@ module DEBUGGER__
       # * `h[elp] <command>`
       #   * Show help for the given command.
       when 'h', 'help', '?'
-        if arg
-          show_help arg
-        else
-          @ui.puts DEBUGGER__.help
-        end
+        show_help arg
         return :retry
 
       ### END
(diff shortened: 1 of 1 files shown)


lib/debug/color.rb
```diff
@@ -64,6 +64,16 @@ def color_pp obj, width
       end
     end
 
+    # colorize documentation of h[elp] and ? instructions
+    def colorize_help(desc)
```
st0012, line 68: The `Color` module is for generic coloring helpers. So I think this method shouldn't be added here.

lib/debug/session.rb
```diff
@@ -1159,10 +1159,15 @@ def cancel_auto_continue
     end
 
     def show_help arg
+      require_relative "color"
```
st0012, line 1162: This is not necessary as the files required by `session.rb` like `tracer` would also load it.

lib/debug/session.rb
```diff
@@ -401,27 +401,27 @@ def process_command line
       case cmd
       ### Control flow
 
-      # * `s[tep]`
+      # * => `s[tep]`
```
st0012, line 404: Even if you'd ultimately substitute it, require any existing and future command comment to have additional characters is not the best solution to help you colorize them. You should try other approaches like identifying an alternative pattern instead.

lib/debug/session.rb
```diff
@@ -1159,10 +1159,15 @@ def cancel_auto_continue
     end
 
     def show_help arg
+      require_relative "color"
+      extend Color
```
st0012, line 1163: This method can be called multiple times and putting this will result it to can `extend Color` repeatedly as well, which should be avoided.

lib/debug/config.rb
```diff
@@ -479,12 +479,15 @@ def self.commands
   end
 
   def self.help
```
st0012, line 481: This method is also used for generating `README.md` so its content shouldn't be colorized.

st0012 (commented):
I also think the help input should be colorized 👍 
However, the approaches taken in this PR is not the best solution IMO. I think the baseline for this PR is:
1. The comment's format, whether it's the way it's written or how it is printed, shouldn't be changed. If you do want to change them, that's out of the scope of this PR.
2. Requiring files and extending modules in method definitions should generally be avoided, especially when they'll be called multiple times.

lib/debug/color.rb
kaiquekandykoga: I removed the method from color.rb, and now it is using a proc inside show_help method in session.rb.

Thank you.

lib/debug/session.rb
kaiquekandykoga: require_relative deleted.

Thank you.

lib/debug/session.rb
kaiquekandykoga: All the changes made in the docs were reverted.

Thank you.

lib/debug/session.rb
kaiquekandykoga: I created an include Color above class Session this time, I think it is in a better place.

Thank you.

lib/debug/config.rb
kaiquekandykoga: Reverted changes.

Thank you.

kaiquekandykoga:
> I also think the help input should be colorized 👍 However, the approaches taken in this PR is not the best solution IMO. I think the baseline for this PR is:
> 
>     1. The comment's format, whether it's the way it's written or how it is printed, shouldn't be changed. If you do want to change them, that's out of the scope of this PR.
> 
>     2. Requiring files and extending modules in method definitions should generally be avoided, especially when they'll be called multiple times.

Hello @st0012 ,

I think I covered everything were asked.

This time the script is analyzing if the description starts with an instruction, and if it does colorizes the line.

There is just one workaround I had to make for the step back description. That is because the description starts with an instruction.

Thank you very much.

PR #511, opened 2022-02-01
Do not override an existed thread in the testing framework

ono-max:
A thread for storing debuggee's output is defined https://github.com/ruby/debug/blob/master/test/support/utils.rb#L333 as a `reader_thread` in remote_info. However, it's overridden in https://github.com/ruby/debug/pull/511/files#diff-a1bc4fb86b4a4e8cc3d94199a4d2b9152c2c8da2261be54d0529942f64a8105dL155. This PR fixes it.

### DIFF
--- a/test/support/dap_utils.rb
+++ b/test/support/dap_utils.rb
@@ -144,7 +144,7 @@ module DEBUGGER__
       }
     ]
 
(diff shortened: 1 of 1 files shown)


st0012:
Is this a cause of random CI failures?

ono-max:
Hopefully.
I can't have confidence in that, though.

PR #525, opened 2022-02-10
Add config option to skip confirmation on quit

hugovandevliert:
## Description

Hi there! I took the liberty to implement a feature that I would love to see added.

In short: this PR adds a config option that disables the 'Really quit?' prompt after hitting `CRTL+D` or `q[uit]`.

To give some context:
I have been using `pry` for a long time. Currently, I'm using `debug` for my personal projects, but professionally we haven't made the switch (yet). Meaning I often switch between the two.

I'm used to stopping debugging sessions by hitting `CTRL+D`, which instantly closes `pry`. In `debug` however, this gives the 'Really quit?' prompt, which requires a `y` and `↵`.

There is already `q!` which quits instantly, so this PR might be redundant. However, I find myself often forgetting about it and getting slightly frustrated during debugging sessions :).

This PR allows me to simply put `config set no_confirm_quit true` in my `~/.rdbgrc` and never worry about remembering to use the slightly awkward `q!` sequence again.

### DIFF
--- a/test/debug/quit_test.rb
+++ b/test/debug/quit_test.rb
@@ -27,6 +27,13 @@ module DEBUGGER__
       end
     end
 
+    def test_quit_quits_immediately_if_no_confirm_quit_is_set_to_true
+      debug_code(program) do
+        type 'config set no_confirm_quit true'
+        type 'q'
+      end
+    end
+
     def test_quit_with_exclamation_mark_quits_immediately_debugger_process
       debug_code(program) do
         type 'q!'
--- a/README.md
+++ b/README.md
@@ -470,6 +470,7 @@ config set no_color true
   * `RUBY_DEBUG_POSTMORTEM` (`postmortem`): Enable postmortem debug (default: false)
   * `RUBY_DEBUG_FORK_MODE` (`fork_mode`): Control which process activates a debugger after fork (both/parent/child) (default: both)
   * `RUBY_DEBUG_SIGDUMP_SIG` (`sigdump_sig`): Sigdump signal (default: disabled)
+  * `RUBY_DEBUG_NO_CONFIRM_QUIT` (`no_confirm_quit`): Do not ask for confirmation on q[uit] or Crtl-D (default: false)
 
 * BOOT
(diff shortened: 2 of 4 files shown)


test/debug/quit_test.rb
```diff
@@ -27,6 +27,13 @@ def test_quit_does_not_quit_debugger_process_if_not_confirmed
       end
     end
 
+    def test_quit_quits_immediately_if_no_confirm_quit_is_set_to_true
+      debug_code(program) do
+        type 'config set no_confirm_quit true'
+        type 'q'
+      end
+    end
```
hugovandevliert, line 35: This spec doesn't pass. I suspect it's because of `test/tool/test_builder.rb:121`, but I'm not sure. If the maintainers are interested in this PR I'll happily spend some time trying to make this spec pass.

st0012:
@hugovandevliert As an ex-pry guy, I get the motivation of this proposal. It also bothered me a while for adjusting to the new commands.
But I feel adding and maintaining a configuration option just for this is unjustified. I hope we can approach the solution from the other direction: what can the debugger do better so you won't need to switch between `pry` and it? Like is it because of the lack of commands like `show-source`? Or it's for other reasons? Please tell us more :-)

hugovandevliert:
> @hugovandevliert As an ex-pry guy, I get the motivation of this proposal. It also bothered me a while for adjusting to the new commands.
> But I feel adding and maintaining a configuration option just for this is unjustified. I hope we can approach the solution from the other direction: what can the debugger do better so you won't need to switch between `pry` and it? Like is it because of the lack of commands like `show-source`? Or it's for other reasons? Please tell us more :-)

Fair enough. The only fact that's keeping me from using this debugger full-time is the fact that we're still using `pry` at my workplace. Sounds like I should lobby some more at my coworkers to switch over so I can get used to this debugger some more :-)

Thanks for your time.

hugovandevliert:
TUI support looks amazing. Will definitely keep an eye out for that. Thanks for the examples :-)

ko1:
Can we close it? Like `apt ... -y` we can introduce "do not ask anything" option is acceptable.

Issue #408, opened 2021-11-22
The console prevents Zeitwerk from autoloading certain constants


fxn:
(Originally reported in https://github.com/rails/rails/issues/43691).

The `debug` console prevents Zeitwerk from autoloading constants that are children of namespaces defined in a file. This script reproduces the issue:

```ruby
require 'tmpdir'

Dir.mktmpdir do |dir|
  Dir.chdir(dir)

  FileUtils.mkdir_p('lib/git/ref')
  File.write('lib/git/ref.rb', 'module Git::Ref; end')
  File.write('lib/git/ref/collection.rb', 'module Git::Ref::Collection; end')

  File.write('test.rb', <<~EOS)
    require 'zeitwerk'

    loader = Zeitwerk::Loader.new
    loader.push_dir('lib')
    loader.setup

    p Git::Ref::Collection
  EOS

  system 'ruby test.rb'
  system 'rdbg test.rb'
end
```

1. `ruby` is able to autoload `Git::Ref::Collection` just fine.
2. If you type `continue` in the `debug` console, `test.rb` is also executed correctly.
3. If, instead, you `next 4` and type `eval Git::Ref::Collection` you'll get a `NameError`.
4. Same if you use the `irb` debugger command and try to evaluate `Git::Ref::Collection` in that session.

In case it matters, when Zeitwerk finds a file like `ref.rb` that defines a namespace, it sets a trace point on the `:class` event, because if the file had this content:

```ruby
module Git
  module Ref
    Collection
  end
end
```

it should be able to autoload `Git::Ref::Collection` on line 3. In order to provide that feature, Zeitwerk listens to `:class` events, and as soon as the one for `Git::Ref` is triggered, Zeitwerk lists the contents of all `git/ref` directories in the root paths, and sets autoloads for what it finds. In this case, it would set an autoload for `Collection` on the module object stored in `Git::Ref`.

Problem with `byebug` was that Ruby does not support nested trace points. Are we in the same situation perhaps?

st0012:
> Problem with byebug was that Ruby does not support nested trace points.

Yeah I think it's the same problem: when using stepping commands (e.g. `step` or `next`), the debugger will operate inside a tracepoint callback. And unless the callback leaves, other callbacks (like Zeitwerk's) would be blocked.

In this case, we can see that Zeitwerk's `tracepoint_class_callback` is disabled after `next` command is called. I used a slightly tweaked script for this:

```rb
require 'tmpdir'

[...]

PR #540, opened 2022-02-16
Use `TracePoint.allow_reentry` when available

casperisfine:
Fix: https://github.com/ruby/debug/issues/408

This fixes compatibility with TracePoint users such as Zeitwerk

Test script:

```ruby
require 'fileutils'
FileUtils.mkdir_p('/tmp/lib/foo')
File.write('/tmp/lib/foo.rb', 'module Foo; end')
File.write('/tmp/lib/foo/bar.rb', 'Foo::Bar = 1')

require 'zeitwerk'

loader = Zeitwerk::Loader.new
loader.push_dir('/tmp/lib')
loader.setup

require 'debug'
binding.break
p Foo::Bar
```

before:
```
[8, 16] in /tmp/debug-zeitwerk.rb
     8| loader = Zeitwerk::Loader.new
     9| loader.push_dir('/tmp/lib')
    10| loader.setup
    11|
    12| require 'debug'
=>  13| binding.break
    14| p Foo::Bar
    15|
    16| p :done
=>#0    <main> at /tmp/debug-zeitwerk.rb:13
(rdbg) p Foo::Bar    # command
eval error: uninitialized constant Foo::Bar
  (rdbg)//tmp/debug-zeitwerk.rb:1:in `<main>'
=> nil
```

after:
```
[8, 16] in /tmp/debug-zeitwerk.rb
     8| loader = Zeitwerk::Loader.new
     9| loader.push_dir('/tmp/lib')
    10| loader.setup
    11|
    12| require 'debug'
=>  13| binding.break
    14| p Foo::Bar
    15|
    16| p :done
=>#0    <main> at /tmp/debug-zeitwerk.rb:13
(rdbg) p Foo::Bar    # command
=> 1
```

Note that I'm fairly unfamiliar with this codebase, so I'll try to see how to add tests, but it may take me a while. I figured it was worth starting a PR anyway.

### DIFF
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -356,6 +356,16 @@ module DEBUGGER__
       [:return_value,     "_return"],
     ]
 
+    def eval_with_binding(binding, src)
+      if binding
+        f, _l = binding.source_location
+        binding.eval(src, "(rdbg)/#{f}")
+      else
+        frame_self = current_frame.self
+        instance_eval_for_cmethod(frame_self, src)
+      end
+    end
+
     def frame_eval src, re_raise: false
       @success_last_eval = false
 
@@ -365,13 +375,12 @@ module DEBUGGER__
         b.local_variable_set(name, var) if /\%/ !~ name
       end
 
-      result = if b
-                  f, _l = b.source_location
-                  b.eval(src, "(rdbg)/#{f}")
-                else
-                  frame_self = current_frame.self
-                  instance_eval_for_cmethod(frame_self, src)
-                end
+      result = if TracePoint.respond_to?(:allow_reentry)
+        TracePoint.allow_reentry { eval_with_binding(b, src) }
+      else
+        eval_with_binding(b, src)
+      end
+
       @success_last_eval = true
       result
 


lib/debug/thread_client.rb
```diff
@@ -406,9 +410,9 @@ def show_src(frame_index: @current_frame_index,
           unless start_line
             if frame.show_line
               if dir > 0
-                start_line = frame.show_line
+                start_line = frame.show_line.lineno
```
st0012, line 413: `frame.show_line` should already be an integer so this change is not need.

lib/debug/thread_client.rb
```diff
@@ -367,7 +367,11 @@ def frame_eval src, re_raise: false
 
       result = if b
```
st0012, line 368: I think `TracePoint.allow_reentry` should wrap the entire evaluation part. That is:

```rb
def eval_with_binding(b, src)
  if b
    f, _l = b.source_location
    b.eval(src, "(rdbg)/#{f}")
  else
    frame_self = current_frame.self
  end
end

# in frame_eval
result = if TracePoint.respond_to?(:allowallow_reentry)
           TracePoint.allowallow_reentry do
             eval_with_binding(b, src)
           end
         else
           eval_with_binding(b, src)
         end
```

lib/debug/thread_client.rb
casperisfine: Yeah, something was wrong with my environment. Removed.

lib/debug/thread_client.rb
casperisfine: Done.

st0012:
@casperisfine I'm not sure why the tests are failing and I can't reproduce them locally 😕
I'll also investigate this when I after time this week.

st0012:
There is an issue that breakpoints can now be triggered inside another breakpoint's subsession. 

The debugger hasn't supported nested suspension yet. So currently, having multiple subsessions can only be caused by bugs and the debugger raises an exception when that happens. But with `allow_reentry`, it's possible to trigger nested subsessions with normal usages (see example). And this crashing behavior becomes a problem.

### Example

#### Script

```rb
def foo
  10
end

binding.b
```

#### Commands

```
$ exe/rdbg -e "break Object#foo ;; c ;; foo" test_540.rb
```

1. `break Object#foo` - set breakpoint on foo
2. `c` - continue the program and enter the `binding.b` breakpoint
3. `foo` - call foo and trigger the breakpoint we set earlier


#### Result

(debug log are enabled by merging https://github.com/ruby/debug/pull/469 locally)

```
❯ exe/rdbg -e "break Object#foo ;; c ;; foo ;; c" test_540.rb

# not important

DEBUGGER (DEBUG): #<DBG:TC 1:running@test_540.rb:5:in `<main>'> changes mode (running -> waiting)
DEBUGGER (DEBUG): #<DBG:TC 1:waiting@test_540.rb:5:in `<main>'> sends Event { type: :suspend, args: [":breakpoint", "[\"/Users/st0012/projects/debug/lib/debug/session.rb\", 2227]"] } to Session

# enters binding.b subsession
DEBUGGER (INFO): enter_subsession
[1, 5] in test_540.rb
     1| def foo
     2|   10
     3| end
     4|
=>   5| binding.b
=>#0    <main> at test_540.rb:5
(rdbg:commands) foo
DEBUGGER (DEBUG): #<DBG:TC 1:waiting@test_540.rb:5:in `<main>'> receives Cmd { type: :eval, args: [":pp", "\"foo\""] } from Session
DEBUGGER (DEBUG): #<DBG:TC 1:waiting@test_540.rb:5:in `<main>'> changes mode (waiting -> running)
DEBUGGER (DEBUG): #<DBG:TC 1:running@test_540.rb:5:in `<main>'> changes mode (running -> waiting)
DEBUGGER (DEBUG): #<DBG:TC 1:waiting@test_540.rb:5:in `<main>'> sends Event { type: :load, args: ["<RubyVM::InstructionSequence:<main>@(rdbg)/test_540.rb:1>", "\"foo\""] } to Session
DEBUGGER (INFO): Load (rdbg)/test_540.rb
DEBUGGER (DEBUG): #<DBG:TC 1:waiting@test_540.rb:5:in `<main>'> receives Cmd { type: :continue, args: [] } from Session
DEBUGGER (DEBUG): #<DBG:TC 1:waiting@test_540.rb:5:in `<main>'> changes mode (waiting -> running)
DEBUGGER (DEBUG): #<DBG:TC 1:running@test_540.rb:5:in `<main>'> is suspended for :call
DEBUGGER (DEBUG): #<DBG:TC 1:running@test_540.rb:5:in `<main>'> changes mode (running -> waiting)

# tries to enter another subsession because of the breakpoint we set
DEBUGGER (DEBUG): #<DBG:TC 1:waiting@test_540.rb:5:in `<main>'> sends Event { type: :suspend, args: [":breakpoint", "\"Object#foo\""] } to Session
#<Thread:0x000000010cbd6f98@DEBUGGER__::SESSION@server /Users/st0012/projects/debug/lib/debug/session.rb:149 run> terminated with exception (report_on_exception is true):
/Users/st0012/projects/debug/lib/debug/session.rb:1537:in `enter_subsession': already in subsession (RuntimeError)
        from /Users/st0012/projects/debug/lib/debug/session.rb:235:in `process_event'
        from /Users/st0012/projects/debug/lib/debug/session.rb:199:in `session_server_main'
        from /Users/st0012/projects/debug/lib/debug/session.rb:169:in `block in activate'
DEBUGGER (DEBUG): #<DBG:TC 1:waiting@test_540.rb:5:in `<main>'> changes mode (waiting -> running)
["DEBUGGER Exception: /Users/st0012/projects/debug/lib/debug/thread_client.rb:1011",
 #<RuntimeError: already in subsession>,
 ["/Users/st0012/projects/debug/lib/debug/session.rb:1537:in `enter_subsession'", "/Users/st0012/projects/debug/lib/debug/session.rb:235:in `process_event'", "/Users/st0012/projects/debug/lib/debug/session.rb:199:in `session_server_main'", "/Users/st0012/projects/debug/lib/debug/session.rb:169:in `block in activate'"]]

# not important

```


I think to solve #408 we probably need more preparation, like

- Supporting nested subsessions with `allow_reentry`, which is a great feature but will only work for Ruby 3.1+. and the inconsistent behavior may confuse users.
- Or, adjust the current subsession check to ignore the later subsessions instead of raising an error.

But this is @ko1's call.

ko1:
Yes, supporting nested subsession should be supported with this change.

> Supporting nested subsessions with allow_reentry, which is a great feature but will only work for Ruby 3.1+. and the inconsistent behavior may confuse users.

We can say "please update to 3.1".

st0012:
> Inconsistently working is better than consistently broken.

To be clear, I didn't mean keep the broken behavior. I only mean triggering breakpoint inside another breakpoint will be ignored (no nested subsessions) but this conflict with Zeitwerk (and other TracePoint using gems) will be fixed.

ko1:
I think nested subsession (nested breakpoint) should be supported, if it is possible.
Anyway, I'll rewrite it, so I close this PR.

PR #547, opened 2022-02-22
Add the new test framework for DAP and CDP

ono-max:
This test framework is designed to test both DAP and CDP for a single method.
# Example
```ruby
# frozen_string_literal: true

module DEBUGGER__

  class BreakTest < TestCase
    PROGRAM = <<~RUBY
      1| module Foo
      2|   class Bar
      3|     def self.a
      4|       "hello"
      5|     end
      6|   end
      7|   Bar.a
      8|   bar = Bar.new
      9| end
    RUBY

    def test_break1
      run_protocol_scenario PROGRAM do
        req_set_breakpoints 5
        req_set_breakpoints 8
        req_continue
        assert_line_num 5
        req_continue
        assert_line_num 8
        req_terminate_debuggee
      end
    end
  end
end
```

# API

- run_protocol_scenario program, dap: true, cdp: true, &scenario

Execute debugging `program` with `&scenario`. If you want to test it only for DAP, you can write as follows:

`run_protocol_scenario program, cdp: false ...`

- req_add_breakpoint(lineno, path: temp_file_path, cond: nil)

Sends request to rdbg to add a breakpoint.

- req_delete_breakpoint bpnum

Sends request to rdbg to delete a breakpoint.

- req_set_exception_breakpoints

Sends request to rdbg to set exception breakpoints.

- req_continue

Sends request to rdbg to resume the program.

- req_step

Sends request to rdbg to step into next method.

- req_next

Sends request to rdbg to step over next method.

- req_finish

Sends request to rdbg to step out of current method.

- req_step_back

Sends request to rdbg to step back from current method.

- req_terminate_debuggee

Sends request to rdbg to terminate the debuggee.

- assert_reattach

Passes if reattaching to rdbg is successful.

- assert_hover_result(expected, expression: nil)

Passes if result of `expression` is equal to `expected`.

- assert_repl_result(expected, expression: nil)

Passes if result of `expression` is equal to `expected`.

- assert_watch_result(expected, expression: nil)

Passes if result of `expression` is equal to `expected`.

- assert_line_num(expected)

Passes if `expected` is equal to the location where debugger stops.


st0012:
@ono-max the new APIs look amazing! Thanks for the great work 🎉

test/support/protocol_utils.rb
```diff
@@ -0,0 +1,343 @@
...
+        | All Protocol Messages |
+        -------------------------
+
+        #{@backlog.join("\n")}
+      DEBUGGER_MSG
+
+      last_msg = @backlog.reverse[0..3].map{|m|
+        h = m.sub(/(D|V)(<|>)(D|V)\s/, '')
+        JSON.pretty_generate(JSON.parse h)
+      }.reverse.join("\n")
```
st0012, line 259: I feel that this part may need a comment for explanation.

test/support/protocol_utils.rb
```diff
@@ -0,0 +1,343 @@
...
+        result = result_res.dig(*keys)
+        assert_kind_of expected, result, failure_msg
+      }
+    end
+
+    def fetch_dap_json
+      return if File.exist? DAP_JSON_PATH
+
+      json = Net::HTTP.get(URI.parse('https://microsoft.github.io/debug-adapter-protocol/debugAdapterProtocol.json'))
+      File.write DAP_JSON_PATH, json
```
st0012, line 228: Shouldn't we store this file as a source of truth? From this method, it looks like each developer and the CI could have different versions of the protocol if it's changed?
If we want to keep up with the latest protocol, we can have a CI action comparing the project's version and the latest version.

ono-max: > If we want to keep up with the latest protocol, we can have a CI action comparing the project's version and the latest version.

Yes, I want to keep up with the latest version, but I don't know how to do it without too much requesting.

test/support/protocol_utils.rb
ono-max: Thanks, this part is complex.
I'll refactor the logic.

PR #555, opened 2022-03-10
Fix some parts in test/support/protocol_utils.rb to run tests correctly

ono-max:


test/support/protocol_utils.rb
```diff
@@ -606,6 +604,35 @@ def recv_response
     end
 
     class WebSocketClient
+      class Frame
+        attr_reader :b
+
+        def initialize
+          @b = ''.b
```
st0012, line 611: I didn't know there's a `String#b` method until today 🤯

test/support/protocol_utils.rb
```diff
@@ -606,6 +604,35 @@ def recv_response
     end
 
     class WebSocketClient
+      class Frame
+        attr_reader :b
```
st0012, line 608: Do you think `bytes` will be  a more expressive name?

ono-max: I'm not sure.
Because `String#b` method exists, I think it's enough, though.

st0012: So it's intentional to make it feel like `String#b`?

ono-max: Yes. Also, I'm ok with `binary` if you want a longer name. I don't think `bytes` is a good name because`String#bytes` returns an array of bytes.

st0012: Yeah I think `binary` will be better. That's actually why I asked at the first place: I didn't know what `b` stands for.

st0012:
@ono-max thanks for the update!
@ko1 I think this can be merged 🙂

PR #559, opened 2022-03-12
Add assert locals to protocol tests

st0012:


st0012:
cc @ono-max

st0012:
And since we now also cover local variables, I think the "raw" tests can be deleted?

ono-max:
I prefer "assert_assert_local_result expected, variable: " to align with others.

ono-max:
I prefer "assert_assert_local_result expected, variable: " to align with others.

st0012:
@ono-max how about having both? `assert_local_result value: val, name: name, type: type` and `assert_local_results(array)`

ono-max:
1. In the testing APIs, I want to minimize the number of APIs because it's simple for users to write tests.

2. As I said before, I prefer `assert_local_result expected, variable: nil` or `assert_local_result expected, name: nil`.  I don't want to write types of values because it can be inferred from `expected`.

st0012:
> We don't need to write types of values because they can be inferred from expected

That's interesting, how does the framework now if we expect `nil` or `"nil"` from CDP response? I think it'll return `"nil"` for both cases with different types?

ono-max:
You might want to read https://github.com/ruby/debug/blob/master/test/support/protocol_utils.rb#L414.

st0012:
The current value assertion imo isn't practical in the long term. Not passing `type` information means we need to replicate the actual value in the test framework, like

https://github.com/ruby/debug/blob/12c5ac3770f59d71aef72c6b62915c6faa2bf165/test/protocol/hover_test.rb#L82-L88

This means that different test cases are not isolated and that'll make future tests harder to write (because we need to know what constants defined by other tests) and harder to debug.

ono-max:
Right. I agree with your opinion.
Let me rethink about `assert_hover_result` first. Then, I'll be back.

st0012:
@ono-max I don't like the idea that you always own the CDP/DAP test framework and everything to be added needs to be approved by you. Why can't we have this helper first and then you can refactor it when you have a clear idea? 

You didn't get the framework right the first time with #547, right? It actually broke, which is normal in software development. And you later updated it and it works for all of us now. So why can't I implement the assertion helper this way first and need to wait for your implementation? It's not public API and can always be changed.

st0012:
@ono-max Simply put: You can and should rethink about `assert_hover_result`. But unless there are obvious flaws in the `assert_locals_result` helper, I don't know why we need to block its merge when you doing that.

If you later want to change it to `assert_local_result expected, name:, type:`, we can discuss that.

ono-max:
> I don't like the idea that you always own the CDP/DAP test framework and everything to be added needs to be approved by you.

In CDP/DAP tests, @ko1-san doesn't review it, and when PRs are created about changes to the framework, he asks me to review them. That's why I think that I am responsible for them and need to approve them.

----

Although changes to the framework don't affect debugger users, it affects contributors who write tests. Because I want to minimize the impact, I think that I need to decide `assert_hover_result` first. Then, `assert_locals_result` can align with them.

st0012:
> Although changes to the framework don't affect debugger users, it affects contributors who write tests.

There could be contributors affected by the changes, which is an assumption. But what we surely know now is that we can use `assert_locals_result` helper right now, in some tests. And plus with some other new test helpers, we can remove all `raw` tests. Why can't we prioritize that?

Another thing is that, **we as software developers** don't make things right at the first, 2nd, or 3rd try. Trying to make things perfect and align the later implementation is an illusion. And more importantly, this shouldn't be a reason to block other progress. You used the whole #547 PR to implement the framework and helpers, but it didn't workout at all after it's merged. And you didn't even spotted it. No other contributors spotted it either. I spotted it. Because I'm the one who use this framework. You also didn't spot the flaw in the current design, I spotted it. Because again, I'm **the contributor** impacted by those issues.

So can you stop pretending there's other contributors that may be affected by this, while I, who does write these tests, telling you I'm already affected by the lack of type information and `assert_locals_result` helper.

I know and understand that you want to make this framework great. And you did make it great over the time, with mistakes and discussions. But your responsibility to the framework shouldn't become an excuse to hold it and not letting other people touching it, or ask them to develop with your own pace. This is not your personal project.

ono-max:
Thank you for explaining to me.
Ok, I accept this PR.

st0012:
Thank you.
Now, my suggestion to methods like `assert_hover_result` is to let all of them take `type:` argument and make the tests isolated again. 

And perhaps later we can set rules to infer some value types automatically, like:

- `1` for integer
- `"\"foobar\""` for string
- `Foo` for class/module (in most cases we don't need to distinguish them)

So `assert_hover_result 1, expression: "i"` doesn't need `type:` argument.

But for matching Ruby objects (e.g. instance of `Post`), we can match them with `type`: `assert_hover_result expression: "p", type: "Post"`

test/support/protocol_utils.rb
```diff
@@ -203,6 +203,53 @@ def assert_reattach
...
+        actual_locals = res.dig(:body, :variables).map { |loc| { name: loc[:name], value: loc[:value], type: loc[:type] } }
+      when 'chrome'
+        current_frame = @crt_frames.first
+        locals_scope = current_frame[:scopeChain].find { |f| f[:type] == "local" }
+        object_id = locals_scope.dig(:object, :objectId)
+        send_request "Runtime.getProperties", objectId: object_id
+        res = find_crt_cdp_response
+
+        actual_locals = res.dig(:result, :result).map do |loc|
+          type = loc.dig(:value, :className) || loc.dig(:value, :type).capitalize # TODO: sync this with get_ruby_type
```
ono-max, line 242: Oh, I noticed that this is incorrect. We can't use `:type` as Ruby class.
I'll fix it later.

test/support/protocol_utils.rb
```diff
@@ -203,6 +203,53 @@ def assert_reattach
...
+
+          { name: loc[:name], value: loc.dig(:value, :description), type: type }
+        end
+      end
+
+      failure_msg = FailureMessage.new{create_protocol_message "result:\n#{JSON.pretty_generate res}"}
+
+      assert_equal(expected, actual_locals, failure_msg)
```
ko1, line 250: Is it intentional to check the lvar's order?
I'm not sure it should be checked or not.

test/support/protocol_utils.rb
st0012: That'd be great, thank you.

test/support/protocol_utils.rb
st0012: no it's not intentional. I've updated it.

ono-max:
Could you add `assert_locals_result` to CONTRIBUTING.md?

CONTRIBUTING.md
```diff
@@ -364,6 +364,12 @@ Passes if result of `expression` is equal to `expected`.
 
 Passes if `expected` is equal to the location where debugger stops.
 
+- assert_locals_result(expected)
+
+Passes if all of `expected` local variable entries match the ones returned by debugger.
+
+An variable entry looks like this: `{ name: "bar", value: "nil", type: "NilClass" }`.
+
```
ono-max, line 372: Could you mention that value and type must be String?

st0012: 👍 done

CONTRIBUTING.md
```diff
@@ -364,6 +364,14 @@ Passes if result of `expression` is equal to `expected`.
...
+Please note that both `value` and `type` needs to be strings.
```
ono-max, line 373: nit
```suggestion
Please note that both `value` and `type` need to be strings.
```

test/support/protocol_utils.rb
```diff
@@ -203,6 +203,53 @@ def assert_reattach
...
+        send_request 'stackTrace',
+                      threadId: 1,
+                      startFrame: 0,
+                      levels: 20
+        res = find_crt_dap_response
+        f_id = res.dig(:body, :stackFrames, frame_idx, :id)
+
+        # get variablesReference
+        send_request 'scopes', frameId: f_id
+        res = find_crt_dap_response
```
ono-max, line 219:
```suggestion
        res = find_crt_dap_response
        assert_dap_response :ScopesResponse
```

test/support/protocol_utils.rb
```diff
@@ -203,6 +203,53 @@ def assert_reattach
...
+        locals_scope = res.dig(:body, :scopes).find { |d| d[:presentationHint] == "locals" }
+        locals_reference = locals_scope[:variablesReference]
+
+        # get variables
+        send_request 'variables', variablesReference: locals_reference
+        res = find_crt_dap_response
```
ono-max, line 225:
```suggestion
        res = find_crt_dap_response
        assert_dap_response :VariablesResponse
```

test/support/protocol_utils.rb
```diff
@@ -203,6 +203,53 @@ def assert_reattach
...
+          end
+        end
+
+        actual_locals = res.dig(:body, :variables).map { |loc| { name: loc[:name], value: loc[:value], type: loc[:type] } }
+      when 'chrome'
+        current_frame = @crt_frames.first
+        locals_scope = current_frame[:scopeChain].find { |f| f[:type] == "local" }
+        object_id = locals_scope.dig(:object, :objectId)
+        send_request "Runtime.getProperties", objectId: object_id
+        res = find_crt_cdp_response
```
ono-max, line 239:
```suggestion
        res = find_crt_cdp_response
        assert_cdp_response 'Runtime.getProperties'
```

ono-max:
I think that we should support Regexp because we can't use `assert_locals_result` in the following case:
```
class Foo
  def initialize a, b
    @a = a
    @b = b
  end
end

a = Foo.new 1, 2
p a # Set a breakpoint here and type `info locals`.
```
a will be something like `a = #<Foo:0x0000000104b37608 @a=1, @b=2>`.

st0012:
How about `{ name: "a", type: "Foo", value: { "@a" => 1, "@b" => 2 }  }`?
And in the cases we don't need to validate the ivars, just `{ name: "a", type: "Foo"}`

ono-max:
For me, it's easier to write a regular expression and read it.

st0012:
Sure, so `{ name: "a", type: "Foo", value: /@a=1, @b=2/ }`?

test/support/protocol_utils.rb
```diff
@@ -203,6 +203,66 @@ def assert_reattach
...
+
+      actual_locals = actual_locals.sort_by { |h| h[:name] }
+      expected = expected.sort_by { |h| h[:name] }
+
+      expected.each_with_index do |expect, index|
+        actual = actual_locals[index]
+        assert_equal(expect[:name], actual[:name], failure_msg)
+        assert_equal(expect[:type], actual[:type], failure_msg)
+
+        if expect[:value].is_a?(Regexp)
```
ono-max, line 258: Do you have a reason why we need to check if the value is Regexp or not?
assert_match can check Regexp and String, right?

st0012: Because if it's given a string, I expect that the values to be equivalent (`assert_equal`)? I don't want `value: "f"` to match `"foobarbaz"`.

ono-max: Oh, you're right. I misunderstand it. I apologize for it.

st0012:
@ono-max I think this is good to merge?

ono-max:
Almost.

You forgot to change my commit suggestions like https://github.com/ruby/debug/pull/559#discussion_r834857499 or https://github.com/ruby/debug/pull/559#discussion_r835736078. Maybe, when you rebase this PR, they are removed.

st0012:
Ah right, they must be left during the force push. I've re-added them.

ono-max:
Great!
LGTM

PR #566, opened 2022-03-14
Close the socket inside of WebSocket class for CDP tests

ono-max:


st0012 (commented):
👍 thank you for the fix

Issue #538, opened 2022-02-15
Allow auto-injected comment in console command to be disabled


sriedel:
**Your proposal**
Currently, when entering a command on the debug console, a '# ruby' comment is auto injected to the right of my input cursor. This is  distracting as I always have movement in the corner of my eye, trying to grab my attention, while I'm trying to concentrate on debugging.

As this comment is not static, but comes and goes as I type, it produces a kind of flickering effect, making the distraction even worse.

Having a way to optionally disable this auto-injected comment would be much appreciated.

PR #574, opened 2022-03-20
change prompt to `ruby` for ruby statements

ko1:
Instead of adding `#ruby` on the right of input Ruby expressions,
this patch changes the prompt to the `ruby`.

This patch can relax issue about #538

### DIFF
--- a/lib/debug/console.rb
+++ b/lib/debug/console.rb
@@ -30,6 +30,18 @@ module DEBUGGER__
         prepend m
       end if SIGWINCH_SUPPORTED
 
+      def parse_input buff, commands
+        c, rest = get_command buff
+        case
+        when commands.keys.include?(c)
+          :command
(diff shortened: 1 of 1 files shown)


st0012:
It looks like the prompt needs an extra space to change?

Like typing `foo`'s prompt is still `(rdbg)`. And it'll require `foo ` for the prompt to be `(ruby)`

ko1:
Now it is intentional because can can't understand `bre` is ruby or debug command.

st0012:
Why? Since the debugger doesn't autocomplete it, it'll only be a Ruby expression.

st0012:
The extra space requirement already exists in master, so it's not introduced by the change. But I'm just curious about the distinguishing limitation.

ko1:
because debugger can't understand the typing `bre` is ruby or debug command (`break`).

PR #576, opened 2022-03-22
Improve protocol tests error handling

st0012:
There are 2 fixes in this PR:

1. `@sock` or `@web_sock` may not always present. For example, I was getting:

```
Error: test_break_stops_at_correct_place(DEBUGGER__::BreakTest1):
  NoMethodError: undefined method `cleanup' for nil:NilClass

        @web_sock.cleanup
                 ^^^^^^^^
/Users/st0012/projects/debug/test/support/protocol_utils.rb:275:in `ensure in execute_cdp_scenario'
/Users/st0012/projects/debug/test/support/protocol_utils.rb:278:in `execute_cdp_scenario'
/Users/st0012/projects/debug/test/support/protocol_utils.rb:47:in `run_protocol_scenario'
test/protocol/break_test.rb:20:in `test_break_stops_at_correct_place'
     17:     RUBY
     18:
     19:     def test_break_stops_at_correct_place
  => 20:       run_protocol_scenario PROGRAM do
     21:         req_add_breakpoint 5
     22:         req_continue
     23:         assert_line_num 5
```

While the actual error was

```
Error: test_break_stops_at_correct_place(DEBUGGER__::BreakTest1): Errno::ECONNREFUSED: Connection refused - connect(2) for 127.0.0.1:62352
/Users/st0012/.rbenv/versions/3.1.0/lib/ruby/3.1.0/socket.rb:64:in `connect'
/Users/st0012/.rbenv/versions/3.1.0/lib/ruby/3.1.0/socket.rb:64:in `connect_internal'
/Users/st0012/.rbenv/versions/3.1.0/lib/ruby/3.1.0/socket.rb:137:in `connect'
/Users/st0012/.rbenv/versions/3.1.0/lib/ruby/3.1.0/socket.rb:642:in `block in tcp'
/Users/st0012/.rbenv/versions/3.1.0/lib/ruby/3.1.0/socket.rb:227:in `each'
/Users/st0012/.rbenv/versions/3.1.0/lib/ruby/3.1.0/socket.rb:227:in `foreach'
/Users/st0012/.rbenv/versions/3.1.0/lib/ruby/3.1.0/socket.rb:632:in `tcp'
/Users/st0012/projects/debug/test/support/protocol_utils.rb:350:in `attach_to_cdp_server'
/Users/st0012/projects/debug/test/support/protocol_utils.rb:269:in `execute_cdp_scenario'
/Users/st0012/projects/debug/test/support/protocol_utils.rb:47:in `run_protocol_scenario'
test/protocol/break_test.rb:20:in `test_break_stops_at_correct_place'
     17:     RUBY
     18:
     19:     def test_break_stops_at_correct_place
  => 20:       run_protocol_scenario PROGRAM do
     21:         req_add_breakpoint 5
     22:         req_continue
     23:         assert_line_num 5
```

In that case, the connection is never established and the test failed before `@web_sock` is even assigned.
But because `@web_sock.close` is always called, the `NoMethodError` it causes override the actual error and makes it hard to debug.

2. A waiting logic should also be placed prior to `Socket.tcp` to make sure remote TCP debuggee is established before the call.

### DIFF
--- a/test/support/protocol_utils.rb
+++ b/test/support/protocol_utils.rb
@@ -251,7 +251,7 @@ module DEBUGGER__
       flunk create_protocol_message "Expected the debuggee program to finish" unless wait_pid @remote_info.pid, TIMEOUT_SEC
     ensure
       @reader_thread.kill
-      @sock.close
+      @sock.close if @sock
       @remote_info.reader_thread.kill
       @remote_info.r.close
       @remote_info.w.close
@@ -272,7 +272,7 @@ module DEBUGGER__
       flunk create_protocol_message "Expected the debuggee program to finish" unless wait_pid @remote_info.pid, TIMEOUT_SEC
     ensure
       @reader_thread.kill
-      @web_sock.cleanup
+      @web_sock.cleanup if @web_sock
       @remote_info.reader_thread.kill
       @remote_info.r.close
       @remote_info.w.close
--- a/test/support/protocol_utils.rb
+++ b/test/support/protocol_utils.rb
@@ -261,6 +261,10 @@ module DEBUGGER__
       ENV['RUBY_DEBUG_TEST_UI'] = 'chrome'
 
       @remote_info = setup_tcpip_remote_debuggee
+      Timeout.timeout(TIMEOUT_SEC) do
+        sleep 0.001 until @remote_info.debuggee_backlog.join.include? @remote_info.port.to_s
+      end
+
       @res_backlog = []
       @bps = [] # [b_id, ...]
       @queue = Queue.new
@@ -348,9 +352,11 @@ module DEBUGGER__
 
     def attach_to_cdp_server
       sock = Socket.tcp HOST, @remote_info.port
+
       Timeout.timeout(TIMEOUT_SEC) do
         sleep 0.001 until @remote_info.debuggee_backlog.join.include? 'Connected'
       end
+
       @web_sock = WebSocketClient.new sock
       @web_sock.handshake @remote_info.port, '/'
       @id = 1
--- a/test/support/utils.rb
+++ b/test/support/utils.rb
@@ -274,7 +274,7 @@ module DEBUGGER__
 
     def kill_safely pid, name, test_info
       return if wait_pid pid, 3
-      
+
       test_info.failed_process = name
 
       Process.kill :TERM, pid


ono-max:
How can we reproduce this error?

st0012:
It happens to me whenever I run `bundle exec ruby TEST_FILES`. For example:

```
❯ bundle exec ruby test/protocol/break_test.rb
Calling `DidYouMean::SPELL_CHECKERS.merge!(error_name => spell_checker)' has been deprecated. Please call `DidYouMean.correct_error(error_name, spell_checker)' instead.
Calling `DidYouMean::SPELL_CHECKERS.merge!(error_name => spell_checker)' has been deprecated. Please call `DidYouMean.correct_error(error_name, spell_checker)' instead.
Loaded suite test/protocol/break_test
Started
E
=====================================================================================================================================================================================================================================================================================================================================================
Error: test_break_stops_at_correct_place(DEBUGGER__::BreakTest1):
  NoMethodError: undefined method `cleanup' for nil:NilClass

        @web_sock.cleanup
                 ^^^^^^^^
/Users/st0012/projects/debug/test/support/protocol_utils.rb:275:in `ensure in execute_cdp_scenario'
/Users/st0012/projects/debug/test/support/protocol_utils.rb:278:in `execute_cdp_scenario'
/Users/st0012/projects/debug/test/support/protocol_utils.rb:47:in `run_protocol_scenario'
test/protocol/break_test.rb:20:in `test_break_stops_at_correct_place'
     17:     RUBY
     18:
     19:     def test_break_stops_at_correct_place
  => 20:       run_protocol_scenario PROGRAM do
     21:         req_add_breakpoint 5
     22:         req_continue
     23:         assert_line_num 5
=====================================================================================================================================================================================================================================================================================================================================================
```

And also `bundle exec rake test_protocol`.

Interestingly, it doesn't happen when I run the same scripts without `bundle exec`.

But regardless what actually causes the error, I think the risk of having a failed connection is real. And in that case, the actual error shouldn't be overridden and having a proper waiting logic can reduce the chance for that to happen.

ono-max:
Hmm, I can't reproduce that.

```shell
% RUBY_DEBUG_PROTOCOL_TEST=1 bundle exec ruby test/protocol/break_test.rb
Loaded suite test/protocol/break_test
Started
..
Finished in 1.799843 seconds.
2 tests, 23 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
1.11 tests/s, 12.78 assertions/s
```

ono-max:
In my opinion, we should consider why this error happens because the case that socket is nil is unusual.

st0012:
@ono-max The error happens because when running with `bundle exec`, the debuggee started with a small delay (probably < 100ms). So when the framework calls `Socket.tcp` directly without waiting, it raises the connection error.
And in that case, because the Socket connection is what failed, the socket ivars remain `nil`.

I'm not sure why that's the case but having a delay between the debuggee and Socket connection is completely expectable imo. That's why we also having waiting logic in `setup_unix_domain_socket_remote_debuggee`. So I think the changes are legit.

ono-max:
Thanks for explaining to me. I understand it.

test/support/protocol_utils.rb
```diff
@@ -347,10 +347,16 @@ def attach_to_dap_server
     HOST = '127.0.0.1'
 
     def attach_to_cdp_server
+      Timeout.timeout(TIMEOUT_SEC) do
```
ono-max, line 350: Could you align this waiting logic with DAP part?

ono-max: Here: https://github.com/ruby/debug/blob/master/test/support/protocol_utils.rb#L239

ono-max: Here: https://github.com/ruby/debug/blob/master/test/support/protocol_utils.rb#L239

st0012: Sure, I've updated it.

ono-max:
Nice work!
LGTM!

Issue #554, opened 2022-03-08
VS Code Breakpoints do not stop depending on the case of Windows drive name


golirev:
**Your environment**

* `ruby -v`: ruby 3.1.1p18 (2022-02-18 revision 53f5fc4236) [x64-mingw-ucrt]
* `rdbg -v`: rdbg 1.4.0

**Describe the bug**
When the drive letter start with "c:"(Lowercase) in a terminal, VS Code Line Breakpoints works fine.
But  the drive letter start with "C:"(Uppercase), it doesn't stop at the breakpoints.

![Windows_terminal](https://user-images.githubusercontent.com/17066700/157270731-90f597de-4809-43e5-93e5-8568b2314b54.png)

The drive letter uppercase/lowercase can be changed in a terminal like below:
```
cd C:\
 or
cd c:\
```

**To Reproduce**

rdbg 1.4.0 has SIGURG issue for Windows. Change a source like below. (I don't know if it is appropriate.)
```
$ git diff lib/debug/server_dap.rb
diff --git a/lib/debug/server_dap.rb b/lib/debug/server_dap.rb
index be4b86d..2f11f0f 100644
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -214,7 +214,7 @@ module DEBUGGER__
           @is_attach = false
         when 'attach'
           send_response req
-          Process.kill(:SIGURG, Process.pid)
+          Process.kill(:SIGURG, Process.pid) unless RUBY_PLATFORM =~ /mswin|mingw/
           @is_attach = true
         when 'setBreakpoints'
           path = args.dig('source', 'path')
@@ -313,7 +313,7 @@ module DEBUGGER__
           exit
         when 'pause'
           send_response req
-          Process.kill(:SIGURG, Process.pid)
+          Process.kill(:SIGURG, Process.pid) unless RUBY_PLATFORM =~ /mswin|mingw/
         when 'reverseContinue'
           send_response req,
                         success: false, message: 'cancelled',
```

Add debugPort parameter to the launch.json.

```
<snip>
        {
            "type": "rdbg",
            "name": "Attach with rdbg",
            "request": "attach",
            "debugPort": "1234"
        }
<snip>
```

**Expected behavior**
Line Breakpoints works, whether the drive letter is uppercase or lowercase.

ko1:
I'll fix issues on Windows soon.

PR #577, opened 2022-03-23
Support case insensitive filename

ko1:
On Windows and some other file systems, they support case insensitve
filenames.  This patch almost supports case insensitve file name
on break command. Maybe we need more features to support case insensitve
file names, so reports are welcome.

fix https://github.com/ruby/debug/issues/554

### DIFF
--- a/lib/debug/breakpoint.rb
+++ b/lib/debug/breakpoint.rb
@@ -220,7 +220,7 @@ module DEBUGGER__
       nearest = nil # NearestISeq
 
       ObjectSpace.each_iseq{|iseq|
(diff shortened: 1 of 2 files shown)


st0012:
There are other path comparisons like:

https://github.com/ruby/debug/blob/3ef203d108237f65fac19d8d6956828377f7288e/lib/debug/thread_client.rb#L797
https://github.com/ruby/debug/blob/3ef203d108237f65fac19d8d6956828377f7288e/lib/debug/thread_client.rb#L410

ko1:
They are not needed because they are not user input path.

PR #582, opened 2022-03-24
Align servers' evaluation approach with debugger core

st0012:
`frame_eval_core` enabled nested TracePoint callbacks, so I think we should use that in our server implementations too.

### DIFF
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -699,8 +699,7 @@ module DEBUGGER__
         frame = get_frame(fid)
         message = nil
 
-        if frame && (b = frame.binding)
-          b = b.dup
+        if frame && (b = frame.eval_binding)
           special_local_variables current_frame do |name, var|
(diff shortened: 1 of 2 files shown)


lib/debug/server_dap.rb
```diff
@@ -699,16 +699,15 @@ def process_dap args
         frame = get_frame(fid)
         message = nil
 
-        if frame && (b = frame.binding)
-          b = b.dup
+        if frame && (b = frame.eval_binding)
           special_local_variables current_frame do |name, var|
```
st0012, line 703: @ko1 perhaps this should be `frame` instead?

ko1: it should be.

ko1:
I need to try it on vscode.

st0012:
Hmm in my tests it doesn't seem to enable nested breakpoints yet.

st0012:
We may need more planning to get servers support nested breakpoints (it could also be impossible). So closing this for now.

PR #587, opened 2022-03-26
Fix formats in evaluation methods for protocol tests

ono-max:


test/protocol/eval_test.rb
```diff
@@ -17,9 +17,9 @@ def test_eval_evaluates_arithmetic_expressions
       run_protocol_scenario PROGRAM do
         req_add_breakpoint 5
         req_continue
-        assert_repl_result 2, expression: 'a'
-        assert_repl_result 4, expression: 'd'
-        assert_repl_result 3, expression: '1+2'
+        assert_repl_result({value: '2', type: 'Integer'}, expression: 'a')
```
st0012, line 20: Why don't we make `value` and `type` keyword arguments too? This 2 level hash will make it harder to write and seems unnecessary.

ono-max: Because assertion methods are usually `assert_something expected, result`.

st0012: I don't think that's a strict rule that we must follow. And in this case we don't even pass the `result` in anyway, we just tell the helper how to get the result. 
So I feel that following that rule doesn't get us much benefit in terms of readability improvement but instead making things more complicated than it needs to be.

ono-max: I want to follow that rule because it's easy for users to configure the expected result as a first argument.
I think that ` assert_repl_result({value: '2', type: 'Integer'}, 'a')` is the best. We can remove 2 level hash in this case.

st0012: ok let's give it a try 👍

st0012:
@ono-max this probably deserves another PR, but can we also support checking error messages? like

```rb
assert_hover_result({ error: "Not defined global variable: \"$bar\"" }, "$bar")
```

currently the test would just fail if it received an error response, even though that's exactly what I want to test.

an example would be, the debugger console [skips a list of global vars](https://github.com/ruby/debug/blob/master/lib/debug/thread_client.rb#L553) but the servers don't. if we want to align servers' behavior with the console, we need to be able to test the error conditions to make sure those vars are skipped.

ono-max:
Nice catch. We should support it, too.

st0012:
@ono-max I think this can be merged but there seems to be conflicts?

Issue #591, opened 2022-03-28
`test_break_only_stops_when_path_matches` failed


ono-max:
**Your environment**

* `ruby -v`: 
ruby 3.1.0p0 (2021-12-25 revision fb4df44d16) [arm64-darwin21]
* `rdbg -v`:

**Describe the bug**
```shell
$ ruby test/debug/break_test.rb
Tests on local and remote. You can disable remote tests with RUBY_DEBUG_TEST_NO_REMOTE=1.
Loaded suite test/debug/break_test
Started
...............F
================================================================================================================================================================================================================================================================================
Failure: test_break_only_stops_when_path_matches(DEBUGGER__::BreakAtInstanceMethodsTest::PathOptionTest):
  #<Test::Unit::AssertionFailedError:
  --------------------
  | Debugger Session |
  --------------------

  > [1, 7] in /var/folders/5j/z2c9zm7124q81f_py4xmd3scp7w9j7/T/debug-20220328-77077-q24op2.rb
  > =>   1| class Foo
  >      2|   def bar; end
  >      3| end
  >      4|
  >      5| Foo.new.bar
  >      6|
  >      7| load "/var/folders/5j/z2c9zm7124q81f_py4xmd3scp7w9j7/T/503f6b1dbafoopy20220328-77077-34msvr.rb"
  > =>#0	<main> at /var/folders/5j/z2c9zm7124q81f_py4xmd3scp7w9j7/T/debug-20220328-77077-q24op2.rb:1
  > break Foo#bar path: /.py/
  > (rdbg) break Foo#bar path: /.py/
uninitialized constant Foo
  >
  > Foo
  > ^^^
  > #0  BP - Method (pending)  Foo#bar
  > (rdbg) c
DEBUGGER:  BP - Method  Foo#bar at /var/folders/5j/z2c9zm7124q81f_py4xmd3scp7w9j7/T/debug-20220328-77077-q24op2.rb:2 is activated.
  > [1, 7] in /var/folders/5j/z2c9zm7124q81f_py4xmd3scp7w9j7/T/debug-20220328-77077-q24op2.rb
  >      1| class Foo
  > =>   2|   def bar; end
  >      3| end
  >      4|
  >      5| Foo.new.bar
  >      6|
  >      7| load "/var/folders/5j/z2c9zm7124q81f_py4xmd3scp7w9j7/T/503f6b1dbafoopy20220328-77077-34msvr.rb"
[...]

PR #592, opened 2022-03-28
Use longer extension name to reduce name collision chance

st0012:
In #591, the test failed because the temp folder's name (`/../z2c9zm7124q81f_py4xmd3scp7w9j7/..`) also matches the path pattern `.py`, so it stopped the breakpoint unexpectedly. Using a longer name like `.html` should greatly reduce the chance of a similar issue from happening again.


ono-max:
Thank you

PR #599, opened 2022-03-29
Verify CAP/DAP requests automatically

st0012:
Since all protocol responses should be verified except for a few exceptions (e.g. DAP's `stepBack`), the verification should be performed automatically. This can prevent situations like https://github.com/ruby/debug/pull/559#discussion_r835736078 and also make the API easier to use (getting response from the same method call).

There are 2 exceptions though. Both of them are DAP requests:
- `stepBack` doesn't return response "yet".
- `stackTrace`'s response doesn't match the schema. I think this is something to be fixed in another PR.

This refactor can actually centralize those exceptions and make them easier to manage. And sooner or later they should go anyway.

### DIFF
--- a/test/support/protocol_utils.rb
+++ b/test/support/protocol_utils.rb
@@ -151,10 +151,11 @@ module DEBUGGER__
       case ENV['RUBY_DEBUG_TEST_UI']
       when 'vscode'
         send_dap_request 'terminate'
-        cleanup_reader
       when 'chrome'
         send_cdp_request 'Runtime.terminateExecution'
       end
+
+      cleanup_reader
     end
 
     def assert_reattach
@@ -314,11 +315,11 @@ module DEBUGGER__
         send_dap_request 'disconnect',
                       restart: false,
                       terminateDebuggee: false
-        cleanup_reader
       when 'chrome'
         @web_sock.send_close_connection
-        cleanup_reader
       end
+
+      cleanup_reader
     end
 
     def req_set_breakpoints_on_dap
--- a/test/support/protocol_utils.rb
+++ b/test/support/protocol_utils.rb
@@ -61,14 +61,12 @@ module DEBUGGER__
       when 'chrome'
         escaped = Regexp.escape File.realpath path
         regexp = "#{escaped}|file://#{escaped}"
-        send_request 'Debugger.setBreakpointByUrl',
+        res = send_cdp_request 'Debugger.setBreakpointByUrl',
                       lineNumber: lineno - 1,
                       urlRegex: regexp,
                       columnNumber: 0,
                       condition: cond
-        res = find_crt_cdp_response
         @bps << res.dig(:result, :breakpointId)
-        assert_cdp_response 'Debugger.setBreakpointByUrl', res
       end
     end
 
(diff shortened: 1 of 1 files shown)


ono-max:
I like this idea 👍

ono-max:
> stackTrace's response doesn't match the schema. 

Could you be more specific?

st0012:
If you verify the `stackTrace` request, you'll get

```
  -------------------
  | Failure Message |
  -------------------

  result:
  {
    "type": "response",
    "command": "stackTrace",
    "request_seq": 10,
    "success": true,
    "message": "Success",
    "body": {
      "stackFrames": [
        {
          "name": "Foo::Bar.a",
          "line": 5,
          "column": 1,
          "source": {
            "name": "debug-20220330-24493-xudwvq.rb",
            "path": "/private/var/folders/yg/hnbymwxd5pn7v94_clc59y6r0000gn/T/debug-20220330-24493-xudwvq.rb",
            "sourceReference": null
          },
          "id": 4
        },
        {
          "name": "<module:Foo>",
          "line": 7,
          "column": 1,
          "source": {
            "name": "debug-20220330-24493-xudwvq.rb",
            "path": "/private/var/folders/yg/hnbymwxd5pn7v94_clc59y6r0000gn/T/debug-20220330-24493-xudwvq.rb",
            "sourceReference": null
          },
          "id": 5
        },
        {
          "name": "<main>",
          "line": 1,
          "column": 1,
          "source": {
            "name": "debug-20220330-24493-xudwvq.rb",
            "path": "/private/var/folders/yg/hnbymwxd5pn7v94_clc59y6r0000gn/T/debug-20220330-24493-xudwvq.rb",
            "sourceReference": null
          },
          "id": 6
        }
      ]
    },
    "seq": 12
  }

  The property '#/body/stackFrames/0/source/sourceReference' of type null did not match the following type: integer
```

[Related code](https://github.com/ruby/debug/blob/264f8c56b3c953eff75d21f3b603b1815a18b168/lib/debug/server_dap.rb#L533-L539)

Even if the fix itself may be simple, fixing this will also affect all the recorded raw tests. So I think it's better to do it in another PR.

test/support/protocol_utils.rb
```diff
@@ -381,16 +338,12 @@ def req_set_breakpoints_on_dap
                           condition: condition
                         }
                       }
-        res = find_crt_dap_response
-        assert_dap_response :SetBreakpointsResponse, res
       }
     end
 
     def assert_disconnect_result
```
ono-max, line 344: I think that we can change this method name to something like `cleanup_reader` because it doesn't assert the result.

st0012: done

test/support/protocol_utils.rb
```diff
@@ -381,20 +338,16 @@ def req_set_breakpoints_on_dap
...
                         }
                       }
-        res = find_crt_dap_response
-        assert_dap_response :SetBreakpointsResponse, res
       }
     end
 
-    def assert_disconnect_result
+    def cleanup_reader
+      @reader_thread.raise Detach
```
ono-max, line 345: Nice refactoring.

test/support/protocol_utils.rb
```diff
@@ -149,43 +124,32 @@ def req_finish
...
       when 'vscode'
-        send_request 'terminate'
-        assert_disconnect_result
+        send_dap_request 'terminate'
+        cleanup_reader
       when 'chrome'
-        send_request 'Runtime.terminateExecution'
-        res = find_crt_cdp_response
-        assert_cdp_response 'Runtime.terminateExecution', res
+        send_cdp_request 'Runtime.terminateExecution'
```
ono-max, line 152: Maybe, I forgot to add `cleanup_reader` method. Could you add it here, too?

st0012: Updated 👍

st0012:
@ono-max can you give this another look? thx

ono-max:
Great! LGTM

PR #602, opened 2022-03-30
Avoid passing nil path to File.basename

st0012:
Some frames may not aways have `path` value.

```
irb((run)@bin/rails:in `require' (not under control)):003:0> frame.location
=> "bin/rails:in `require'"
irb((run)@bin/rails:in `require' (not under control)):004:0> frame.path
=> nil
```

And when that happens, `File.basename(nil)` would cause exception and crash the debugger.

I can reproduce this with: `rdbg --command --open --stop-at-load  -- bundle exec rails s` on any Rails apps.

(I accidentally added the unnecessary `--stop-at-load` and found the issue)

Since in that condition, there's no way to retrieve the correct path as source, I think using the `pretty_path`'s `#<none>` is acceptable.

截圖 2022-03-30 15 11 35

But of course, there may be a better solution than this patch.


st0012:
Actually, I think `frame.location.to_s` is better. Will update.

PR #607, opened 2022-03-30
Add `assert_threads_result` helper and threads test

st0012:
Some notes:
- Ids don't always start from `0` and increment by `1`, as shown in the below example. So asserting ids is pointless imo.
- Threads request is only supported in DAP server.
- Example failure message:

```
result:
{
  "type": "response",
  "command": "threads",
  "request_seq": 8,
  "success": true,
  "message": "Success",
  "body": {
    "threads": [
      {
        "id": 1,
        "name": "#1 /var/folders/yg/hnbymwxd5pn7v94_clc59y6r0000gn/T/debug-20220411-21868-x0t9a2.rb:6:in `<main>'"
      },
      {
        "id": 4,
        "name": "#4 /var/folders/yg/hnbymwxd5pn7v94_clc59y6r0000gn/T/debug-20220411-21868-x0t9a2.rb:2:in `block in foo'"
      }
    ]
  },
  "seq": 9
}.
Expect all thread names to be matched. Unmatched threads:
<[]> expected but was
<["#1 /var/folders/yg/hnbymwxd5pn7v94_clc59y6r0000gn/T/debug-20220411-21868-x0t9a2.rb:6:in `<main>'"]>

diff:
? ["#1 /var/folders/yg/hnbymwxd5pn7v94_clc59y6r0000gn/T/debug-20220411-21868-x0t9a2.rb:6:in `<main>'"]
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -372,6 +372,23 @@ An variable entry looks like this: `{ name: "bar", value: "nil", type: "NilClass
 
 Please note that both `value` and `type` need to be strings.
 
+- assert_threads_result(expected)
+
+Passes if both conditions are true:
+
+1. The number of expected pattern matches the number of threads.
+2. The expected patterns match the thread names in the given order.
+
+Example:
+
+```
+assert_threads_result(
+  [
+    /\.rb:\d:in `<main>'/,
+    /\.rb:\d:in `block in foo'/
+  ]
+)
+```
 
 ## To Update README
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -376,8 +376,8 @@ Please note that both `value` and `type` need to be strings.
 
 Passes if both conditions are true:
 
-1. The number of expected pattern matches the number of threads.
-2. The expected patterns match the thread names in the given order.
+1. The number of expected patterns matches the number of threads.
+2. Every pattern matches a thread name. Notice that the order of threads info is not guaranteed.
 
 Example:
 
--- a/test/protocol/threads_test.rb
+++ b/test/protocol/threads_test.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+require_relative '../support/test_case'
+
+module DEBUGGER__
+  class ThreadsTest < TestCase
+    PROGRAM = <<~RUBY
+       1| def foo
+       2|   Thread.new { sleep 30 }
+       3| end
+       4|
+       5| foo
+       6| sleep 0.1 # make sure the thread stops
+       7| binding.b
+    RUBY
+
+    def test_reponse_returns_correct_threads_info
+      run_protocol_scenario PROGRAM do
+        req_continue
+
+        assert_threads_result(
+          [
+            /\.rb:\d:in `<main>'/,
+            /\.rb:\d:in `block in foo'/
+          ]
+        )
+
+        req_terminate_debuggee
+      end
+    end
+  end
+end
+
--- a/test/protocol/threads_test.rb
+++ b/test/protocol/threads_test.rb
@@ -15,7 +15,7 @@ module DEBUGGER__
     RUBY
 
     def test_reponse_returns_correct_threads_info
-      run_protocol_scenario PROGRAM do
+      run_protocol_scenario PROGRAM, cdp: false do
         req_continue
 
         assert_threads_result(
--- a/test/support/protocol_utils.rb
+++ b/test/support/protocol_utils.rb
@@ -234,15 +234,22 @@ module DEBUGGER__
       case ENV['RUBY_DEBUG_TEST_UI']
       when 'vscode'
         res = send_dap_request 'threads'
+        failure_msg = FailureMessage.new{create_protocol_message "result:\n#{JSON.pretty_generate res}."}
 
         threads = res.dig(:body, :threads)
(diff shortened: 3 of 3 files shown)


test/protocol/threads_test.rb
```diff
@@ -0,0 +1,33 @@
...
+       2|   Thread.new { sleep 30 }
+       3| end
+       4|
+       5| foo
+       6| sleep 0.1 # make sure the thread stops
+       7| binding.b
+    RUBY
+
+    def test_reponse_returns_correct_threads_info
+      run_protocol_scenario PROGRAM do
```
ono-max, line 18:
```suggestion
      run_protocol_scenario PROGRAM, cdp: false do
```

ono-max:
Could you update CONTRIBUTING.md?

st0012:
Sure. But speaking of documentation, do you think we can do that in Ruby comments? Like [test-unit does](https://github.com/test-unit/test-unit/blob/master/lib/test/unit/assertions.rb). And in the contribution guideline we just link to the file.

I think it fits the convention better and it's easier for writers of methods (like me) to remember too 😛

ono-max:
We can write how to use API in the comments and CONTRIBUTING.md. I think that writing in CONTRIBUTING.md is good because they are well organized.

st0012:
I don't think having documentation in 2 places is a good idea. It's the same reason we only write command intro in `session.rb` and generate it in `README.md`, instead of having it in both places.

ono-max:
Well, I see. Or, we can create rdoc page from the comment?

st0012:
Yes. In that case, we'll need to write the comments follow [yard](https://github.com/lsegal/yard)'s specification.

st0012:
@ono-max I've updated the contribution guideline for now. can you check it? thx

CONTRIBUTING.md
```diff
@@ -372,6 +372,23 @@ An variable entry looks like this: `{ name: "bar", value: "nil", type: "NilClass
 
 Please note that both `value` and `type` need to be strings.
 
+- assert_threads_result(expected)
+
+Passes if both conditions are true:
+
+1. The number of expected pattern matches the number of threads.
```
ono-max, line 379:
```suggestion
1. The number of expected patterns matches the number of threads.
```

ono-max:
Nice work! LGTM

CONTRIBUTING.md
```diff
@@ -372,6 +372,23 @@ An variable entry looks like this: `{ name: "bar", value: "nil", type: "NilClass
 
 Please note that both `value` and `type` need to be strings.
 
+- assert_threads_result(expected)
+
+Passes if both conditions are true:
+
+1. The number of expected patterns matches the number of threads.
+2. The expected patterns match the thread names in the given order.
```
ko1, line 380: what the "order" is?

ono-max:
You missed https://github.com/ruby/debug/pull/607/files#r843338555.

test/support/protocol_utils.rb
```diff
@@ -230,6 +230,29 @@ def assert_locals_result expected, frame_idx: 0
...
+        thread_names = threads.map { |t| t[:name] }
+
+        expected_names.each do |expected|
+          thread_names.reject! do |name|
+            name.match?(expected)
+          end
+        end
+
+        failure_msg = FailureMessage.new{create_protocol_message "result:\n#{JSON.pretty_generate res}.\nExpect all thread names to be matched. Unmatched threads:"}
+        assert_equal [], thread_names, failure_msg.to_s
```
ono-max, line 252: Why do you add `to_s` method?

test/support/protocol_utils.rb
```diff
@@ -230,6 +230,29 @@ def assert_locals_result expected, frame_idx: 0
...
+
+        threads = res.dig(:body, :threads)
+
+        assert_equal expected_names.count, threads.count, failure_msg
+
+        thread_names = threads.map { |t| t[:name] }
+
+        expected_names.each do |expected|
+          thread_names.reject! do |name|
+            name.match?(expected)
```
ono-max, line 247: We can use `assert_match` method, could you tell me the reason why you implemented so?

st0012: Why do you think we can use it here?

ono-max: Ah, that's right. My bad 🙏

st0012: No worries 😉 
Do other changes look good to you?

test/support/protocol_utils.rb
```diff
@@ -230,6 +230,29 @@ def assert_locals_result expected, frame_idx: 0
...
+          end
+        end
+
+        failure_msg = FailureMessage.new{create_protocol_message "result:\n#{JSON.pretty_generate res}.\nExpect all thread names to be matched. Unmatched threads:"}
+        assert_equal [], thread_names, failure_msg
```
ono-max, line 252: Can we use `assert_empty`?

st0012: I decided to use `assert_equal` because it provides better highlighting in the failure messages:

截圖 2022-04-17 10 25 01


**`assert_empty`**

截圖 2022-04-17 10 25 15

I guess it's because `assert_empty` doesn't care about individual items in the array, but `assert_equal` does. So it provides a more detailed information.

PR #614, opened 2022-04-02
Use tcpip connection for DAP protocol tests too

st0012:
2 benefits:

1. It can avoid/reduce random broken pipe error:
    - https://github.com/ruby/debug/runs/5750847945?check_suite_focus=true
    - https://github.com/ruby/debug/runs/5729199865?check_suite_focus=true
    - https://github.com/ruby/debug/runs/5785039496?check_suite_focus=true
    - All of them came from DAP's socket connection.
2. We can unify the scenario execution logic.

### DIFF
--- a/test/support/protocol_utils.rb
+++ b/test/support/protocol_utils.rb
@@ -297,10 +297,9 @@ module DEBUGGER__
 
     # Not API
 
-    def execute_dap_scenario scenario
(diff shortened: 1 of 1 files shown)


ko1:
Now UNIX Domain socket is default debug port to connect with VSCode so it is not acceptable.

> It can avoid/reduce random broken pipe error:

It should be fixed respectively.

PR #616, opened 2022-04-06
CheckBreakpoint should only stop at the first condition fulfillment line

### LINKED ISSUE #406: Stop at CheckBreakpoint only for once
Currently, we can add a CheckBreakpoint with `break if: <expression>`. I personally haven't used it before but it looks like it's not practical in most cases because: once the `expression` is fulfilled, the breakpoint will stop at every single line afterward.

```
❯ ruby -Ilib -rdebug check_bp_issue.rb
[1, 6] in check_bp_issue.rb
=>   1| binding.b(do: "break if: a == 1")
     2|
     3| a = 1
     4| b = 2
     5| c = 3
     6| d = 4
=>#0    <main> at check_bp_issue.rb:1
(rdbg:binding.break) break if: a == 1
#0  BP - Check  a == 1
[1, 6] in check_bp_issue.rb
     1| binding.b(do: "break if: a == 1")
     2|
     3| a = 1
=>   4| b = 2
     5| c = 3
     6| d = 4
=>#0    <main> at check_bp_issue.rb:4

Stop by #0  BP - Check  a == 1
(rdbg) c    # continue command
[1, 6] in check_bp_issue.rb
     1| binding.b(do: "break if: a == 1")
     2|
     3| a = 1
     4| b = 2
=>   5| c = 3
     6| d = 4
=>#0    <main> at check_bp_issue.rb:5

Stop by #0  BP - Check  a == 1
(rdbg) c    # continue command
[1, 6] in check_bp_issue.rb
     1| binding.b(do: "break if: a == 1")
     2|
     3| a = 1
     4| b = 2
     5| c = 3
=>   6| d = 4
=>#0    <main> at check_bp_issue.rb:6

Stop by #0  BP - Check  a == 1
(rdbg) c    # continue command
```

So perhaps we should make this a one-off breakpoint?

st0012:
As described in #406

When a condition is fulfilled (e.g. `a >= 1`), it usually remains true for at least a while. However, under the current implementation, it'll stops at every single line after the condition becomes true, which is annoying.

So this PR introduces an extra state to make sure the BP doesn't stop repeatedly after the condition is fulfilled.

Closes #406

### DIFF
--- a/lib/debug/breakpoint.rb
+++ b/lib/debug/breakpoint.rb
@@ -326,8 +326,13 @@ module DEBUGGER__
         next if ThreadClient.current.management?
         next if skip_path?(tp.path)
 
-        if safe_eval tp.binding, @cond
-          suspend
+        if safe_eval(tp.binding, @cond)
+          unless fulfilled_on_current_thread?
+            set_fulfilled_on_current_thread
+            suspend
+          end
+        else
+          set_unfulfilled_on_current_thread
         end
       }
     end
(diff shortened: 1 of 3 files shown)


ko1:
I use `private def ...` to make it clear which methods are private.

Issue #523, opened 2022-02-07
Terminated Event is not fired before disconnect


oats4goats:
**Your environment**

* `ruby -v`: ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-linux]
* `rdbg -v`: rdbg 1.4.0

**Describe the bug**
It looks like in Remote Debug mode (DAP client `nvim-dap`) after the last program step, the debugger disconnects but doesn't fire 'Terminated Event'. In such situation, DAP client doesn't know the [debugging of the debuggee has terminated](https://microsoft.github.io/debug-adapter-protocol/specification#Events_Terminated) and keeps the debug session alive forever. Thus requiring manual treatment.

**To Reproduce**
1. Start a debugger with `rdbg --open --port 38698 target.rb` in a terminal
2. Attach to the adapter with `nvim-dap`
3. Continue or step over till the very end of the program.
4. The debugger in the terminal disconnects.

Example of console output:
```
#14867:[>] {"type":"request","seq":52,"arguments":{"granularity":"statement","threadId":1},"command":"next"}
#14867:[<] {"type":"response","command":"next","request_seq":52,"success":true,"message":"Success","seq":62}
12
DEBUGGER: Disconnected.
```

**Expected behavior**
Before the debugger disconnects 'Terminated Event' is issued so the client could close the debug session properly.

**Additional context**
Relevant discussion in [mfussenegger/nvim-dap](https://github.com/mfussenegger/nvim-dap/issues/429)

ko1:
https://github.com/mfussenegger/nvim-dap/pull/436 solves the issue?

It's a bit tough to observe the debuggee's liveness (because the debugger runs in same process).
But not impossible.

If it should be supported, I'll work it.

mfussenegger:
The discussion in https://github.com/mfussenegger/nvim-dap/issues/469 and the change that resulted out of the conversion (https://github.com/mfussenegger/nvim-dap/pull/481) may be relevant to this as well.

The TL;DR version is that if the debugger process runs to completion, the debug adapter *must* send a terminated event. If the client sends a disconnect request before the debugger process finishes, the debug adapters behavior depends on whether a launch/attach configuration was used and on the (optional) `terminateDebugee` flag of the request.

ko1:
> the debug adapter must send a terminated event.

Could you give me a reference?

st0012:
[...]

PR #618, opened 2022-04-09
Update remote debugging overview with table

st0012:
I think using a table can make this section even more easier to understand while providing more information.

**Before**

截圖 2022-04-09 23 32 05

**After**

截圖 2022-04-09 23 30 40

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,11 +9,15 @@ New debug.rb has several advantages:
 
 * Fast: No performance penalty on non-stepping mode and non-breakpoints.
 * [Remote debugging](#remote-debugging): Support remote debugging natively.
-  * UNIX domain socket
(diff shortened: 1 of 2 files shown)


ko1:
I didn't check nvim-dap-ruby, but is it work enough?

st0012:
@ko1 Yes I think it already works for basic features (I didn't test it intensively though). But based on related issue reports like https://github.com/ruby/debug/issues/523, I think there are users actively using it already.

ko1:
I'll test before merge it.

ko1:
I couldn't try nvim-dap-ruby because of my lack of nvim knowledge...

st0012:
I have removed NeoVim from the table.

PR #621, opened 2022-04-12
Enhance DAP's `setExceptionBreakpoints` command

st0012:
**Main Changes**

1. `Session#add_bp` should return either `nil` (duplicated bp being rejected) or the `Breakpoint` object. Currently it returns `true` for duplicated breakpoints (from `bp.delete`), which makes `setExceptionBreakpoints` returns `verify: true` incorrectly. (see the 3rd commit)
2. The `setExceptionBreakpoints` command should clear all breakpoints before creating the assigned ones. Otherwise, it's not possible to deactivate exception breakpoints. After the changes, its behavior would follow `setBreakpoints` command and [align with debugpy's implementation](https://github.com/microsoft/debugpy/blob/328d4026b16815807e43a26f60988da3de8c05a4/src/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_process_net_command_json.py#L839).
3. It supports setting conditions with exception breakpoints.

**Other Changes**

1. Test framework's `send_request` helper should also provide protocol messages when seeing exceptions (e.g. when the debugger exists expectedly). Currently it doesn't print those messages and makes it hard to debug.
2. To support the improved `setExceptionBreakpoints` command, I also updated `req_set_exception_breakpoints` helper.

### DIFF
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -107,14 +107,14 @@ module DEBUGGER__
                {
                  filter: 'any',
                  label: 'rescue any exception',
-                 #supportsCondition: true,
+                 supportsCondition: true,
                  #conditionDescription: '',
                },
                {
                  filter: 'RuntimeError',
                  label: 'rescue RuntimeError',
                  default: true,
-                 #supportsCondition: true,
+                 supportsCondition: true,
                  #conditionDescription: '',
                },
              ],
@@ -252,27 +252,30 @@ module DEBUGGER__
         when 'setFunctionBreakpoints'
           send_response req
         when 'setExceptionBreakpoints'
-          process_filter = ->(filter_id) {
-            case filter_id
-            when 'any'
-              bp = SESSION.add_catch_breakpoint 'Exception'
-            when 'RuntimeError'
-              bp = SESSION.add_catch_breakpoint 'RuntimeError'
-            else
-              bp = nil
-            end
+          process_filter = ->(filter_id, cond = nil) {
+            bp =
+              case filter_id
+              when 'any'
+                SESSION.add_catch_breakpoint 'Exception', cond: cond
+              when 'RuntimeError'
+                SESSION.add_catch_breakpoint 'RuntimeError', cond: cond
+              else
+                nil
+              end
             {
-              verified: bp ? true : false,
+              verified: !bp.nil?,
               message: bp.inspect,
             }
           }
+
+          SESSION.clear_catch_breakpoints 'Exception', 'RuntimeError'
 
           filters = args.fetch('filters').map {|filter_id|
             process_filter.call(filter_id)
           }
 
           filters += args.fetch('filterOptions', {}).map{|bp_info|
-            process_filter.call(bp_info.dig('filterId'))
+            process_filter.call(bp_info['filterId'], bp_info['condition'])
           }
 
           send_response req, breakpoints: filters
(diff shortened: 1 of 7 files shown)


st0012:
@ono-max can you also give this a look? thx

lib/debug/server_dap.rb
```diff
@@ -252,27 +252,30 @@ def process
...
+              when 'any'
+                SESSION.add_catch_breakpoint 'Exception', cond: cond
+              when 'RuntimeError'
+                SESSION.add_catch_breakpoint 'RuntimeError', cond: cond
+              else
+                nil
+              end
             {
-              verified: bp ? true : false,
+              verified: !bp.nil?,
```
ko1, line 266: bp can be false?

ko1 (approved):

lib/debug/server_dap.rb
st0012: I don't think it should be `false` in any case. If it's not added, it should just be `nil`. Why do you expect it to be `false` too?

ko1: I want to know the reason why the change is needed.

st0012: It's not a necessary change so if you don't want it I can revert it.

test/protocol/catch_test.rb
```diff
@@ -5,22 +5,41 @@
...
+        assert_line_num 3
+        req_terminate_debuggee
+      end
+    end
+
+    def test_set_exception_breakpoints_unsets_exception_breakpoints
+      run_protocol_scenario PROGRAM, cdp: false do
+        req_set_exception_breakpoints([{ name: "RuntimeError" }])
+        req_set_exception_breakpoints([])
+        req_continue
```
ono-max, line 29: Can we finish the program with `req_terminate_debuggee`?

I know that there is no problem in the only DAP case, but it may confuse other users when they write test cases for both protocol.

st0012: This won't work because the debuggee would already have exited and the terminate request will cause borken pipe error:

```
  Failed to send request because of #<Errno::EPIPE: Broken pipe>.
/Users/st0012/projects/debug/test/support/assertions.rb:97:in `assert_block'
/Users/st0012/projects/debug/test/support/protocol_utils.rb:498:in `rescue in send_request'
/Users/st0012/projects/debug/test/support/protocol_utils.rb:487:in `send_request'
/Users/st0012/projects/debug/test/support/protocol_utils.rb:502:in `send_dap_request'
/Users/st0012/projects/debug/test/support/protocol_utils.rb:153:in `req_terminate_debuggee'
test/protocol/catch_test.rb:30:in `block in test_set_exception_breakpoints_unsets_exception_breakpoints'
     27:         req_set_exception_breakpoints([{ name: "RuntimeError" }])
     28:         req_set_exception_breakpoints([])
     29:         req_continue
  => 30:         req_terminate_debuggee
     31:       end
     32:     end
     33:
/Users/st0012/projects/debug/test/support/protocol_utils.rb:299:in `execute_dap_scenario'
/Users/st0012/projects/debug/test/support/protocol_utils.rb:50:in `run_protocol_scenario'
test/protocol/catch_test.rb:26:in `test_set_exception_breakpoints_unsets_exception_breakpoints'
```

ono-max: Oh, that's right. Thank you for explaining to me.

ono-max (commented):
Sorry for my late reviews.

Nice. LGTM

PR #628, opened 2022-04-21
Refactor Session and ThreadClient

st0012:
These 2 commits were cherry-picked from #503, which I think are beneficial for future chagnes regardless if TUI is supported or not.

### DIFF
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -415,55 +415,64 @@ module DEBUGGER__
       raise if re_raise
     end
 
-    def show_src(frame_index: @current_frame_index,
-                 update_line: false,
-                 max_lines: CONFIG[:show_src_lines] || 10,
+    def get_src(frame:,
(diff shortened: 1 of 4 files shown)


lib/debug/thread_client.rb
```diff
@@ -415,55 +415,66 @@ def frame_eval src, re_raise: false
...
+          end_line = [start_line + max_lines, lines.size].min
+        end
 
+        if start_line != end_line && max_lines
+          [start_line, end_line, lines]
+        else
+          []
+        end
+      else # no file lines
+        []
```
ko1, line 455: nil?

ko1:
```ruby
def foo
  nil
end

a, b, c = *foo()
p [a, b, c] #=> [nil, nil, nil]
```

st0012: 👍  updated

PR #636, opened 2022-05-16
Move block and method args to their own methods

### LINKED ISSUE #635: Slow backtraces due to inspect of large local variables
**Your environment**
* `rdbg -v`: rdbg 1.5.0

**Describe the bug**
Backtraces are very slow at some points.  I added logging and was able to pin the slowness to `frame_info.rb` in `parameters_info`:
```
          local_var = local_variable_get(var)
          t1 = Time.now
          value = DEBUGGER__.safe_inspect(local_var, short: true)
          t2 = Time.now - t1
          if t2 > 0.1
            ::DEBUGGER__.info  "Slow parameters info, safe inspect... length: #{local_var.inspect.length}, type: #{local_var.class}, var: #{var}, value: #{value}, time#{t2}"
          end
```

This resulted in some output for slow parameters like the following:
```
DEBUGGER (INFO): Slow parameters info, safe inspect... length: 11729953, type: Hash, var: env, value: {"rack.version"=>[1, 6], "rack.errors"=>..., time0.321832
DEBUGGER (INFO): Slow parameters info, safe inspect... length: 11729955, type: Array, var: args, value: [{"rack.version"=>[1, 6], "rack.errors"=..., time0.436716
```
I believe the slowness is attributed to the `obj.inspect` in the safe_inspect.  I have a stack with 100+ depth which is resulting in about 25 seconds for the backtrace to return due to slowness on a bunch of frames (most seem to be with the env variable).  With the dap_client it doesn't even look like the parameters are used, so it feels a bit pointless retrieving these.  However the thread_client however does use this to output the shortened info.  Here's a sample output on the rdbg session for one of the slow entries:
```
  #35	Rails::Engine#call(env={"rack.version"=>[1, 6], "rack.errors"=>...) at ~/.gem/ruby/3.1.2/gems/railties-7.0.2.4/lib/rails/engine.rb:530
```

**To Reproduce**
I'm not super clear what's unique about the environment, but it's something to do with having a very large environment variable that causes this type of behavior.  I was able to reproduce in the dap client (VS Code) and with the regular thread client (using rdbg -A).

**Expected behavior**
Backtraces should not take really long times to execute.

dpeacock:
The DAP server doesn't use args and in some instances retrieving these is very slow.  Extracted them into their own methods to prevent this form slowing down DAP requests as the data is not used.
- Relates to #635

### DIFF
--- a/lib/debug/frame_info.rb
+++ b/lib/debug/frame_info.rb
@@ -49,11 +49,10 @@ module DEBUGGER__
       # p frame_type: frame_type, self: self
       case frame_type
       when :block
-        level, block_loc, _args = block_identifier
+        level, block_loc = block_identifier
         "block in #{block_loc}#{level}"
       when :method
-        ci, _args = method_identifier
-        "#{ci}"
+        method_identifier
       when :c
         c_identifier
       when :other
@@ -83,16 +82,13 @@ module DEBUGGER__
 
     def block_identifier
       return unless frame_type == :block
-      args = parameters_info
       _, level, block_loc = location.label.match(BLOCK_LABL_REGEXP).to_a
-      [level || "", block_loc, args]
+      [level || "", block_loc]
     end
 
     def method_identifier
       return unless frame_type == :method
-      args = parameters_info
-      ci = "#{klass_sig}#{callee}"
-      [ci, args]
+      "#{klass_sig}#{callee}"
     end
 
     def c_identifier
@@ -138,6 +134,17 @@ module DEBUGGER__
       end
     end
 
+    def parameters_info
+      vars = iseq.parameters_symbols
+      vars.map{|var|
+        begin
+          { name: var, value: DEBUGGER__.safe_inspect(local_variable_get(var), short: true) }
+        rescue NameError, TypeError
+          nil
+        end
+      }.compact
+    end
+
     private
 
     def get_singleton_class obj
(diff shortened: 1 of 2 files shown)


lib/debug/frame_info.rb
```diff
@@ -83,16 +82,23 @@ def frame_type
...
       return unless frame_type == :block
-      args = parameters_info
       _, level, block_loc = location.label.match(BLOCK_LABL_REGEXP).to_a
-      [level || "", block_loc, args]
+      [level || "", block_loc]
+    end
+
+    def block_args
+      return unless frame_type == :block
+      parameters_info
```
st0012, line 91: so this and `method_args` actually just calls `parameters_info` 🤔 
I think the frame type check isn't very necessary here because they're not exposed to end users. how about we remove them and just call `parameters_info` from ThreadClient?

st0012:
@dpeacock I don't think it fixes #635. It just excludes DAP users from the problem. But console users will still face it as far as I understand?

dpeacock:
> @dpeacock I don't think it fixes https://github.com/ruby/debug/issues/635. It just excludes DAP users from the problem. But console users will still face it as far as I understand?

Good call-out it just prevents the issue from happening for DAP clients.  Console users will experience it, but at least can specific the depth of the stacktrace from the commandline which the DAP clients currently cannot do.

st0012:
@dpeacock Thanks for the PR 🎉 
@ko1 I think this one looks good and can be merged.

st0012:
@peterzhu2118 Could you give this a look? Thx

peterzhu2118 (approved):
Looks good, thank you!

PR #638, opened 2022-05-16
Use levels and startFrame in DAP server backtrace

### LINKED ISSUE #637: DAP Server stack trace should use startFrame and levels parameters
**Your environment**
* `rdbg -v`: rdbg 1.5.0

**Describe the bug**
The DAP server does not use the parameters on the [StackTrace](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_StackTrace) request.  In some instances this can result in a very large stack trace being returned, which can take extra time.  VS code for example requests a depth of 20 at a time and has a 'more' button to retrieve additional data.  Currently these are ignored, and the more button just resends the same data every time.

**To Reproduce**
Set a breakpoint in VS Code where the stack depth is > 20.  VS Code will request 20 levels, but the dap server will ignore this and send the entire stack trace.

**Expected behavior**
Only send the requested levels.

dpeacock:
## Description
Added functionality to the DAP server to use the levels & startFrame parameters as described in the [StackTrace request documentation](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_StackTrace). 

- Fixes #637

### DIFF
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -621,8 +621,11 @@ module DEBUGGER__
 
       case type
       when :backtrace
+        levels = req.dig('arguments', 'levels') || @target_frames.length
+        startFrame = req.dig('arguments', 'startFrame') || 0
+
         event! :dap_result, :backtrace, req, {
-          stackFrames: @target_frames.map{|frame|
+          stackFrames: @target_frames[startFrame, levels].map{|frame|
             path = frame.realpath || frame.path
             source_name = path ? File.basename(path) : frame.location.to_s
 


lib/debug/server_dap.rb
```diff
@@ -621,8 +621,11 @@ def process_dap args
 
       case type
       when :backtrace
+        levels = req.dig('arguments', 'levels') || @target_frames.length
+        startFrame = req.dig('arguments', 'startFrame') || 0
```
st0012, line 625: Given the usage below, I think this part should be:

```rb
start_frame = req.dig('arguments', 'startFrame') || 0
levels = req.dig('arguments', 'levels') 

if levels && levels != 0
  end_frame = start_frame + levels
  end_frame = -1 if end_frame > @target_frames.length
else
  end_frame = -1
end
```

Because `levels` isn't an absolute location, just a number for the next page.

PR #640, opened 2022-05-19
Don't stop at DAP's connection

### LINKED ISSUE #629: Unexpected stoppage on remote debugging connection
Recently I mostly use remote debugging and this trap (https://github.com/ruby/debug/blob/master/lib/debug/server.rb#L229) means the program (usually Rails server) will be stopped upon connection, even with the `nonstop` flag.

And I'll need to manually skip the first few frames every time the connection is established, which is annoying. One example:

```
❯ ~/projects/debug/exe/rdbg -A
# No sourcefile available for /Users/st0012/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/puma-5.6.2/lib/puma/single.rb
=>#0    [C] Thread#join at ~/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/puma-5.6.2/lib/puma/single.rb:61
  #1    Puma::Single#run at ~/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/puma-5.6.2/lib/puma/single.rb:61
  # and 16 frames (use `bt' command for all frames)

Stop by SIGURG
(rdbg:remote)
```

So my questions are:

1. ~~Why does the debugger need to stop for this signal?~~
2. ~~If it's needed in some cases, can we skip this behavior when `nonstop` flag is used?~~

1. Why should the debuggee be paused with signal by itself upon activation? The stoppage is [triggered by the debugger server itself](https://github.com/ruby/debug/blob/0319744b2a73f158170a599c8781d091b084be97/lib/debug/server.rb#L71) ([pause source](https://github.com/ruby/debug/blob/0319744b2a73f158170a599c8781d091b084be97/lib/debug/server.rb#L332-L335))

st0012:
Based on the discussion in #629, DAP's connection stoppage can be removed.

Closes #629


ko1:
I found two lines comment outs are needed to "non-stop" with VSCode attach mode.
Is it intentional or the following patch your favorite in the scenario on https://github.com/ruby/debug/issues/629 (I guess so)?

```patch
diff --git a/lib/debug/server.rb b/lib/debug/server.rb
index 9abb187..abd3752 100644
--- a/lib/debug/server.rb
+++ b/lib/debug/server.rb
@@ -68,7 +68,7 @@ module DEBUGGER__
           } unless already_connected

           setup_interrupt do
-            pause unless already_connected
+            # pause unless already_connected
             process
           end

diff --git a/lib/debug/server_dap.rb b/lib/debug/server_dap.rb
index 84f954d..0842784 100644
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -256,7 +256,7 @@ module DEBUGGER__
           UI_DAP.local_fs_map_set req.dig('arguments', 'localfs') || req.dig('arguments', 'localfsMap')
         when 'attach'
           send_response req
-          Process.kill(UI_ServerBase::TRAP_SIGNAL, Process.pid)
+          # Process.kill(UI_ServerBase::TRAP_SIGNAL, Process.pid)
           @is_attach = true
           UI_DAP.local_fs_map_set req.dig('arguments', 'localfs') || req.dig('arguments', 'localfsMap')
         when 'setBreakpoints'
```

st0012:
I think the current code works: DAP connection will skip the initial connection stop, but CUI connection will still stop.
However, after skipping the initial stop, the terminating disconnect request doesn't work anymore:

**With Stop (master)**

```
#53409:[>] {"command":"disconnect","arguments":{"restart":false,"terminateDebuggee":true},"type":"request","seq":15}
#53409:[<] {"type":"response","command":"disconnect","request_seq":15,"success":true,"message":"Success","seq":18}
# Debuggee exits
```

**Without Stop (this branch)**

```
#52510:[>] {"command":"disconnect","arguments":{"restart":false,"terminateDebuggee":true},"type":"request","seq":10}
#52510:[<] {"type":"response","command":"disconnect","request_seq":10,"success":true,"message":"Success","seq":11}
#52510:[<] {"type":"event","event":"terminated","seq":12}
DEBUGGER: Disconnected.
# Debuggee doesn't exit
```

It looks like without the initial stop, this `kill!` command won't be processed:

https://github.com/ruby/debug/blob/49ac0a15e12523e248f68bf497e0be44942abfbf/lib/debug/server_dap.rb#L318

To reproduce:

1. Start a Rails server with `rdbg -O -n -c -- bundle exec rails`
2. Attach to it from VSCode
3. Click the disconnect dropdown and click `Stop`

ko1:
fixed by #688

PR #652, opened 2022-06-09
Make tests `Exception#inspect` independent

st0012:
So the tests won't fail because of changes like https://github.com/ruby/ruby/pull/4857

### DIFF
--- a/test/console/debugger_local_test.rb
+++ b/test/console/debugger_local_test.rb
@@ -34,7 +34,7 @@ module DEBUGGER__
             type "catch Exception"
             type "c"
             type "_raised"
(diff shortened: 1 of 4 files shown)


test/console/debugger_local_test.rb
```diff
@@ -34,7 +34,7 @@ def test_raised_is_accessible_from_repl
             type "catch Exception"
             type "c"
             type "_raised"
-            assert_line_text(/#<NameError: undefined local variable or method `foo' for main:Object/)
```
st0012, line 37: Note that the exception here is printed with `#inspect` because that's how debugger prints objects from console (pp + width limit). So it's not produced by the test framework.

st0012:
cc @mame

PR #654, opened 2022-06-11
Protocol test framework improvements

st0012:
I made these changes when debugging Ruby head's hanging issue and they did make the build failures easier to catch.

1. Both debuggee connection and workflows should have timeout
2. Rename `WebSocketClient#cleanup` to `#close` because that's all it does
3. Only perform cleanup logic when the socket/threads are present, which may not be the case when the error happens during test setup.

### DIFF
--- a/.github/workflows/protocol.yml
+++ b/.github/workflows/protocol.yml
@@ -17,6 +17,7 @@ jobs:
   test:
 
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     strategy:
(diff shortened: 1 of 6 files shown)


st0012:
@peterzhu2118 I could also use your help on this test improvement. You can run protocol tests with `bundle exec rake test_protocol`, which is **not** included in `rake test`.

peterzhu2118 (approved):
Looks good, thank you!

PR #658, opened 2022-06-17
Extract remote debugging guide

st0012:
## Motivation

1. For users that don't need remote debugging, the large part of remote debugging intro only delays their time to find the info they're interested in (e.g. configuration).
2. Because the remote debugging part already takes a lot of space in readme, adding more detailed info becomes harder and harder.

## Description

Leave simple remote debugging overview in readme and move most of the rest into a dedicated file.


st0012:
@peterzhu2118 This is not ready for final review yet, but I could use your feedback on the structure 🙂

peterzhu2118 (commented):
:+1: Structure looks good to me.

st0012:
@ko1 What do you think about the structure?

ko1:
It is intentional to show large part of remote debugging because users don't know the remote debugging culture.
If we make separate pages, other parts (configuration, debug command and so on) should be separated too and README.md should contain simple tutorial (and the remote debugging example should be contained).

Another point: Your proposal (and current README.md) contains VSCode/Chrome frontend but they should not be contained in *Remote debugging* because users should not care they are connected remotely or not. Current README.md lists all and maybe it is not emphasize, but making separate page, it should be emphasize and it is confusing.

st0012:
> If we make separate pages, other parts (configuration, debug command and so on) should be separated too and README.md should contain simple tutorial (and the remote debugging example should be contained).

That's exactly what I want to do. But I don't feel like doing it anymore. Thanks for the recognition and closing without even discussion 🙂

PR #660, opened 2022-06-25
Refactor test framework structure

st0012:
Because we test console and protocol servers differently, we have dedicated test helpers for each of them. But currently all test helpers are loaded in `TestCase`, which is inherited in both types of tests. This means we always have redundant methods loaded when running tests. For example, protocol tests don't need the `debug_code` helper, but they still have it defined.

This can cause some problems:

1. It creates name conflict and make naming harder.
2. It could make debugging harder because it's hard to tell which helpers/constants are relevant to the current tests.

So in this PR, I made these changes:

1. Create `ConsoleTestCase` and `ProtocolTestCase`, which only contains relevant helpers for that type of tests. And a test class should inherit either one of them.
2. Move mixings, methods, and classes from top-level `TestCase` to the above classes when suited.
3. Since we now have different test case classes, we don't need `Utils` modules anymore. So I removed them and moved their content to those classes.
4. Categorize top-level tests (the ones on `/test`) and move them to better locations.

### Before

**[Console](https://github.com/ruby/debug/runs/7048420722?check_suite_focus=true)**

```
Finished in 194.17270551 seconds.
-------------------------------------------------------------------------------
246 tests, 389 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
-------------------------------------------------------------------------------
```

**[Protocol](https://github.com/ruby/debug/runs/7048420931?check_suite_focus=true)**

```
Finished in 54.990209757 seconds.
-------------------------------------------------------------------------------
42 tests, 7603 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
-------------------------------------------------------------------------------
```

### After

**[Console](https://github.com/ruby/debug/runs/7053673087?check_suite_focus=true)**

```
Finished in 187.907855868 seconds.
-------------------------------------------------------------------------------
246 tests, 389 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
-------------------------------------------------------------------------------
```

**[Protocol](https://github.com/ruby/debug/runs/7053673065?check_suite_focus=true)**

```
Finished in 56.320539272 seconds.
-------------------------------------------------------------------------------
42 tests, 7603 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
-------------------------------------------------------------------------------
```


st0012:
cc @peterzhu2118

test/support/console_test_case.rb
```diff
@@ -306,54 +181,62 @@ def manual_debug_code(program)
       write_temp_file(strip_line_num(program))
       remote_info = setup_unix_domain_socket_remote_debuggee
 
-      Timeout.timeout(TIMEOUT_SEC) do
```
peterzhu2118, line 309: Can we remove this timeout? This looks like it could cause an infinite loop.

test/support/console_test_case.rb
```diff
@@ -50,133 +47,38 @@ def create_message fail_msg, test_info
...
-              new_thread { debug_code_on_local },
-              new_thread { debug_code_on_unix_domain_socket },
-              new_thread { debug_code_on_tcpip },
-            ]
-
-            th.each do |t|
-              if fail_msg = t.join.value
-                th.each(&:kill)
-                flunk fail_msg
+      Timeout.timeout(60) do
```
peterzhu2118, line 51: Should we respect `TIMEOUT_SEC` for this timeout?

test/support/protocol_test_case.rb
```diff
@@ -37,11 +41,13 @@ class Detach < StandardError
     # API
 
     def run_protocol_scenario program, dap: true, cdp: true, &scenario
-      write_temp_file(strip_line_num(program))
-      execute_dap_scenario scenario if dap
-      execute_cdp_scenario scenario if cdp
+      Timeout.timeout(60) do
```
peterzhu2118, line 44: Should we also use `TIMEOUT_SEC` here too?

test/support/test_case.rb
```diff
@@ -75,334 +79,108 @@ def program_with_line_numbers(program)
...
-        end
-      end
+    private
 
-      def char bytes
-        @binary << bytes
-      end
+    def wait_pid pid, sec
+      total_sec = 0.0
+      wait_sec = 0.001 # 0.1ms
```
peterzhu2118, line 105: This is 1ms right? Not 0.1ms.

test/support/console_test_case.rb
st0012: Yeah good idea. There shouldn't be any test taking more than 10 seconds to run atm.

test/support/test_case.rb
st0012: Yeah. But in order to preserve the behavior, I'll update the comment instead.

test/support/console_test_case.rb
st0012: Won't the timeout prevent that from happening instead? 

Or did you mean **not** removing it? I now see it's removed unintentionally.

peterzhu2118: Yeah I meant whether or not it's ok to remove the timeout.

st0012: Ah right. I didn't realize it was a removal so I thought you were asking me to remove it 😬

test/support/protocol_test_case.rb
```diff
@@ -37,11 +41,13 @@ class Detach < StandardError
     # API
 
     def run_protocol_scenario program, dap: true, cdp: true, &scenario
-      write_temp_file(strip_line_num(program))
-      execute_dap_scenario scenario if dap
-      execute_cdp_scenario scenario if cdp
+      Timeout.timeout(30) do
```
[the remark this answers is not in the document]
peterzhu2118: Why don't we bump the global timeout to 30s? I'm not a fan of the fact that this is hard coded and not configurable.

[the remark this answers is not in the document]
st0012: The current `TIMEOUT_SEC` is for individual tests, while the timeout here could have for multiple tests. Forcing them to use the same value will cause problems imo. 
For example, when debugging a hanging test I usually shorten the timeout to 3 to get quicker feedback. But then if I run the entire suite, many of the protocol tests will now fail because of it. So I'd need to change that back to 30 again.

peterzhu2118 (approved):
Looks good, thanks for addressing the comments!

PR #664, opened 2022-06-27
return when `line` is `nil`

### LINKED ISSUE #622: Remote Debuggee crashes on Debugger's "quit" command
**Your environment**
Windows 11: Using Docker Desktop container (with docker compose) running Alpine, using foreman to run Procfile on start

* `ruby -v`: `ruby 3.1.1p18 (2022-02-18 revision 53f5fc4236) [x86_64-linux-musl]`
* `rdbg -v`: `rdbg 1.5.0`

**Describe the bug**
Running `quit` command on attached debugger causes the debuggee to exit (logs provided below).

**To Reproduce**
1. Set up a rails application on a docker container (I can provide my configurations if needed).
2. Add a `require 'debug/open_nonstop'` in the `bin/rails` file at the top of the file.
3. Add a line in Procfile to start `bin/rails server -p 3000 -b 0.0.0.0` if it's not already there
4. Set up your docker-compose.yml to run the `bin/dev` command on start to run foreman using the Procfile
5. Start docker container `docker compose up -d`
6. Attach to debuggee using `docker compose exec <container name> rdbg --attach`
7. Add `debugger` line in the application and trigger the breakpoint (by visiting a web page or any method you prefer)
8. In rdbg console you attached to (in step 6) type `quit`
9. Debugging process quits but the container also breaks saying:
-- ```unsupported:  \  exited with code 1```

**Expected behavior**
Debugger should gracefully detach and debuggee should be open to accept a new debugger attachment.

**Additional context**
Dockerfile uses `FROM ruby:3.1.1-alpine3.15` base image.

Whole docker container terminal output when following above steps to reproduce:
```
20:26:35 web.1  | started with pid 10
20:26:35 web.1  | DEBUGGER: Debugger can attach via UNIX domain socket (/tmp/ruby-debug-sock-0/ruby-debug-ruby-debug-10)
20:26:37 web.1  | DEBUGGER: Connected.
20:26:43 web.1  | => Booting Puma
20:26:43 web.1  | => Rails 7.0.2.3 application starting in development 
20:26:43 web.1  | => Run `bin/rails server --help` for more startup options
20:26:44 web.1  | Puma starting in single mode...
20:26:44 web.1  | * Puma version: 5.6.4 (ruby 3.1.1-p18) ("Birdie's Version")
20:26:44 web.1  | *  Min threads: 5
20:26:44 web.1  | *  Max threads: 5
20:26:44 web.1  | *  Environment: development
20:26:44 web.1  | *          PID: 10
20:26:44 web.1  | * Listening on http://0.0.0.0:3000
20:26:44 web.1  | Use Ctrl-C to stop
20:26:47 web.1  | Started GET "/" for 192.168.32.1 at 2022-04-13 20:26:47 +0000
20:26:47 web.1  | Cannot render console from 192.168.32.1! Allowed networks: 127.0.0.0/127.255.255.255, ::1
20:26:47 web.1  |   ActiveRecord::SchemaMigration Pluck (1.0ms)  SELECT "schema_migrations"."version" FROM "schema_migrations" ORDER BY "schema_migrations"."version" ASC
20:26:47 web.1  | Processing by ApplicationController#index as HTML
20:26:50 web.1  | unsupported: 
20:26:51 web.1  | exited with code 1
20:26:51 system | sending SIGTERM to all processes
```

ko1:
`@sock.gets` can return `nil` if `@sock` is closed.
Return this method and cleanup the server connection.

fix #622

### DIFF
--- a/lib/debug/server.rb
+++ b/lib/debug/server.rb
@@ -164,6 +164,7 @@ module DEBUGGER__
           }
         end
 
+        return unless line
         next if line == :can_not_read
 
         case line
@@ -187,7 +188,7 @@ module DEBUGGER__
             raise "pid:#{Process.pid} but get #{line}"
           end
         else
-          STDERR.puts "unsupported: #{line}"
+          STDERR.puts "unsupported: #{line.inspect}"
           exit!
         end
       end


lib/debug/server.rb
```diff
@@ -164,6 +164,7 @@ def process
           }
         end
 
+        return unless line
```
st0012, line 167: If the `nil` is expected after the Socket is closed, I think it'd be clearer if we dear with it around line 162. For example:

```rb
         if l = @sock.gets
           DEBUGGER__.info "UI_Server received: #{l.chomp}"
         else
           # socket is closed
           return
         end
```

ko1: Now the lines are already complicated so I left it.

PR #669, opened 2022-07-01
`PP` trick for long `inspect` code

### LINKED ISSUE #635: Slow backtraces due to inspect of large local variables

ko1:
* Stop `pp` when the output exceeds the maximum length
* Use `PP.singleline_pp` to disable breaking features

fix #635

This does not solve the problem if the `pretty_print` consumes
time for example using network connections.

### DIFF
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -441,6 +441,7 @@ module DEBUGGER__
 
       when 'stackTrace'
         tid = req.dig('arguments', 'threadId')
+
         if tc = find_waiting_tc(tid)
           request_tc [:dap, :backtrace, req]
         else
@@ -551,9 +552,10 @@ module DEBUGGER__
 
       case type
       when :backtrace
-        result[:stackFrames].each.with_index{|fi, i|
+        result[:stackFrames].each{|fi|
+          frame_depth = fi[:id]
           fi[:id] = id = @frame_map.size + 1
-          @frame_map[id] = [req.dig('arguments', 'threadId'), i]
+          @frame_map[id] = [req.dig('arguments', 'threadId'), frame_depth]
           if fi[:source]
             if src = fi[:source][:sourceReference]
               src_id = @src_map.size + 1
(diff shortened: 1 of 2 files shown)


ko1:
```ruby
class C
  def initialize
    @a = ['x'] * 10_000
    @b = ['y'] * 10_000
    @c = ['z'] * 10_000
  end
end

$obj = C.new

def rec0 o = $obj
  a = 0
  binding.b
end


(1..1000).each{|i|
  eval("def rec#{i}(o = $obj) = (a = #{i}; rec#{i-1})\n")
}

rec1000
```


```
$ time exe/rdbg  -n target.rb -e 'bt;;c

# before
real    0m7.428s
user    0m7.055s
sys     0m0.154s

# after
real    0m0.799s
user    0m0.504s
sys     0m0.111s
```

PR #671, opened 2022-07-01
skip if `ThreadClient.current.management?`

### LINKED ISSUE #609: Weird behavior on catch command
**Your environment**

* `ruby -v`: `ruby 3.1.0p0 (2021-12-25 revision fb4df44d16) [x86_64-darwin21]`
* `rdbg -v`: `1.5.0`

### Description

I spotted a weird behavior when using `catch Exception` or `catch StandardError` and then causing an exception in the console. Example:

```
❯ exe/rdbg -e "catch Exception do: 'debugger spotted an error' ;; raise 'foo' ;; " target.rb
[1, 1] in target.rb
=>   1| puts(123)
=>#0    <main> at target.rb:1
(rdbg:commands) catch Exception do: 'debugger spotted an error'
#0  BP - Catch  "Exception"
(rdbg:commands) raise 'foo'
[1, 1] in (rdbg)/target.rb
=>   1| raise 'foo'
=>#0    <main> at (rdbg)/target.rb:1
  #1    [C] Binding#eval at ~/projects/debug/lib/debug/thread_client.rb:369
  # and 2 frames (use `bt' command for all frames)

Stop by #0  BP - Catch  "Exception"
(rdbg:catch) 'debugger spotted an error'
"debugger spotted an error"
eval error: foo
nil

# catch breakpoint is triggered and performed command as expected, so far so good.
# let's run some other commands
(rdbg) ls    # outline command
Object.methods: inspect  to_s

# for some reason, the catch bp's command is triggered again
(rdbg:catch) 'debugger spotted an error'
"debugger spotted an error"

# and it exits the current bp for no apparent reason
123
```

### Reproduction

**Scripit**

```rb
puts(123)
```

**Command and output**

1. `$ exe/rdbg -e "catch Exception do: 'debugger spotted an error' ;; raise 'foo' ;; " target.rb`
2. Run arbitrary debugger command, like `ls`

### Expected Behavior

1. The `'debugger spotted an error'` message is only printed once.
2. The arbitrary command doesn't exit the debugging session.

### Actual Behavior

1. The `do:` command is executed for the second time when no new exception is raised.
2. When 1 happens, it also tries to leave the current debugging session without being told to do so.

However, if we use a more specific exception name, like `catch RuntimeError`, neither the problem occurs.

Therefore, I suspect that debugger has some internal exceptions happening between step 1 and 2, which triggered the catch breakpoint unexpectedly.

ko1:
fix #609

### DIFF
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -17,7 +17,9 @@ module DEBUGGER__
 
   module SkipPathHelper
     def skip_path?(path)
-      CONFIG.skip? || !path ||
+      !path ||
+      CONFIG.skip? ||
+      ThreadClient.current.management? ||
       skip_internal_path?(path) ||
       (skip_paths = CONFIG[:skip_path]) && skip_paths.any?{|skip_path| path.match?(skip_path)}
     end


st0012:
I think it's great to centralize skipping logic but I don't think `skip_path?` is the place for it. Maybe we can have rename `SkipPathHelper` to `SkipHelper` and add something like `skip?` to it? So it'll be

```rb
skip?(path)

module SkipHelper
  def skip?(path)
    ThreadClient.current.management? || skip_path?(path)
  end
end
```

ko1:
Good idea. Another PR.

PR #672, opened 2022-07-02
Don't fail fast on CI

peterzhu2118:
fail-fast defaults to true, meaning that when a CI job fails, all the
other ones are cancelled. CI is flaky, so we shouldn't cancel other CI
jobs when one fails.


st0012:
Yeah this makes sense 👍

PR #642, opened 2022-05-22
Centralize configuration defaults

st0012:
Currently, we use `CONFIG[:key] || default` to apply default values at the places we use the config. This has 2 problems:

1. The config descriptions' default and the actual default needs to be synced manually. And as the time goes by, some of them are likely to be out-of-sync gradually.
2. Config defaults are scattered around the codebase and you need to know where a config is used to see its default. With code searching this isn't a big issue but it's still nicer to see all of them in one place.

So in this PR, I moved all configs' default to the `config.rb`, just right after the config type info. And now the default is synced between `README <-> config command <-> actual logic`.

### Changes on the `config` command output

**Before**

```
(rdbg) config    # command
log_level = (default)          # UI: Log level same as Logger (default: WARN)
show_src_lines = (default)     # UI: Show n lines source code on breakpoint (default: 10 lines)
show_frames = (default)        # UI: Show n frames on breakpoint (default: 2 frames)
use_short_path = (default)     # UI: Show shorten PATH (like $(Gem)/foo.rb)
no_color = (default)           # UI: Do not use colorize (default: false)
no_sigint_hook = (default)     # UI: Do not suspend on SIGINT (default: false)
no_reline = (default)          # UI: Do not use Reline library (default: false)
skip_path = (default)          # CONTROL: Skip showing/entering frames for given paths (default: [])
skip_nosrc = (default)         # CONTROL: Skip on no source code lines (default: false)
keep_alloc_site = (default)    # CONTROL: Keep allocation site and p, pp shows it (default: false)
postmortem = (default)         # CONTROL: Enable postmortem debug (default: false)
fork_mode = (default)          # CONTROL: Control which process activates a debugger after fork (both/parent/child) (default: both)
sigdump_sig = (default)        # CONTROL: Sigdump signal (default: disabled)
nonstop = (default)            # BOOT: Nonstop mode
stop_at_load = (default)       # BOOT: Stop at just loading location
init_script = (default)        # BOOT: debug command script path loaded at first stop
commands = (default)           # BOOT: debug commands invoked at first stop. commands should be separated by ';;'
no_rc = (default)              # BOOT: ignore loading ~/.rdbgrc(.rb)
history_file = (default)       # BOOT: history file (default: ~/.rdbg_history)
save_history = (default)       # BOOT: maximum save history lines (default: 10,000)
port = (default)               # REMOTE: TCP/IP remote debugging: port
host = (default)               # REMOTE: TCP/IP remote debugging: host (localhost if not given)
[...]

PR #675, opened 2022-07-04
Fix remote client configuration

st0012:
**This is a leftover issue caused by https://github.com/ruby/debug/pull/642**

When a remote debugger client is started, it doesn't have configurations filled like the debuggee does, which can cause some issues. Like:

```
disconnected (undefined method `empty?' for nil:NilClass

      if !history_file.empty?
                      ^^^^^^^)
```

This is because `CONFIG.update` replaces the entire config hash and removes all the default values `CONFIG` already has. So if we start a client with `rdbg -A`, this will happen:

1. `config = DEBUGGER__::Config::parse_argv(ARGV)` returns a hash like `{ mode: :attach }`.
    - The `mode` isn't a real configuration option. It's only used to pass the mode information from the `-A` flag.
2. The `CONFIG` will be initialized with all the default configuration values.
    - `{:mode=>:start, :log_level=>:WARN, :show_src_lines=>10, :show_frames=>2, :use_short_path=>false, .....}`
3. `CONFIG.update` will replace the hash in 2 with the `{ mode: :attach }`.
4. The client loses all the default configurations.

### DIFF
--- a/exe/rdbg
+++ b/exe/rdbg
@@ -20,7 +20,7 @@ when :start
 
 when :attach
   require_relative "../lib/debug/client"
-  ::DEBUGGER__::CONFIG.update config
+  ::DEBUGGER__::CONFIG.set_config **config
 
   begin
     if ARGV.empty? && config[:port]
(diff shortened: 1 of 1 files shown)


st0012:
cc @peterzhu2118

peterzhu2118 (approved):
Looks good!

PR #682, opened 2022-07-05
Support per-project rdbgrc(.rb) files

st0012:
Each project may have its own project-specific debugger configuration, like different paths to skip. So the debugger should also read `.rdbgrc` files at the project root, similar to what `irb` does with `irbrc`.

### DIFF
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -2032,11 +2032,13 @@ module DEBUGGER__
   end
 
   def self.load_rc
-    [[File.expand_path('~/.rdbgrc'), true],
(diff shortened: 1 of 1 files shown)


ko1:
BTW, config/*.rb is not enough to write per-workspace (per-application) setting?

st0012:
It's possible but putting dev tool specific configs into the Rails config isn't a good practice. It's why we still have `.irbrc` or `.pryrc` at project root instead of having those code inside `config/`.

ko1:
> It's possible but putting dev tool specific configs into the Rails config isn't a good practice.

If you know, could you tell me why?

PR #684, opened 2022-07-06
Add warning about legacy debug gem

### LINKED ISSUE #676: rdbg + Bundler -> numerous gem version conflicts
**Your environment**

* `ruby -v`: `ruby 3.1.1p18 (2022-02-18 revision 53f5fc4236) [arm64-darwin21]`
* `rdbg -v`: `rdbg 1.4.0`

**Describe the bug**
My goal is to debug RSpec test cases in a large project that uses Bundler. (Integrating with VS Code, but that is not relevant to this bug report.) The command that I invoke is:
```
rdbg --command --open --stop-at-load --sock-path=<foo> -- bundle exec rspec <bar>/spec/zzz_im_spec.rb

# tried this variant too
bundle exec rdbg --command --open --stop-at-load --sock-path=<foo> -- bundle exec rspec <bar>/spec/zzz_im_spec.rb
```

However, my bundle includes conflicting versions of numerous gems that `ruby/debug` requires by itself: `io-console`, `irb`, `json`, `reline` and `timeout`. It's not clear to me whether all of these are default gems -- I don't _think_ so, but maybe I'm wrong and the default gems have exploded in Ruby 3! Regardless, I end up with conflicts because Bundler tries to activate the version in my Gemfile, but `debug` has already activated a different version. It raises:

```
You have already activated io-console 0.5.11, but your Gemfile requires io-console 0.5.6. Since io-console is a default gem, you can either remove your dependency on it or try updating to a newer version of bundler that supports io-console as a default gem. (Gem::LoadError)
```

This is very mysterious considering that this happens even when `debug` is explicitly installed in the bundle and `rdbg` is invoked via `bundle exec`.

**To Reproduce**
Add _any_ of the gems mentioned above to the Gemfile; `io-console 0.5.6` is a good test case with Ruby 3.1.1 and rdbg 1.4.0. Then just invoke `rdbg` and ask it to run `bundle exec` on any script file that activates the bundle. (I can create a self-contained reproduction as a gist, if you are interested.)

**Expected behavior**
- There should be _some_ way to use rdbg with Bundler, because both of them are default gems
- I have a workaround (align Gemfile perfectly with default gems) but wish I didn't need to do this

**Additional context**
I realize this isnt' necessarily a bug in `debug`; I'm hoping for guidance. Happy to work on a fix, if this is indeed a bug.

xeger:
This closes #676 whose root cause was "user did not follow README instructions closely enough." ;-)

## Description
- reinforced the need for debug 1.x when debugging in a Bundler project

### DIFF
--- a/misc/README.md.erb
+++ b/misc/README.md.erb
@@ -38,6 +38,9 @@ If you use Bundler, write the following line to your Gemfile.
 gem "debug", ">= 1.0.0"
 ```
 
+(The version constraint is important; `debug < 1.0.0` is an older,
+abandoned gem that is completely different from this product.)
+
 # HOW TO USE
 
 To use a debugger, roughly you will do the following steps:
@@ -134,7 +137,7 @@ d => 4
 ### Invoke the program from the debugger as a traditional debuggers
 
 If you don't want to modify the source code, you can set breakpoints with a debug command `break` (`b` for short).
-Using `rdbg` command to launch the program without any modifications, you can run the program with the debugger.
+Using `rdbg` command (or `bundle exec rdbg`) to launch the program without any modifications, you can run the program with the debugger.
 
 ```shell
 $ cat target.rb                        # Sample program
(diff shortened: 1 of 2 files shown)


misc/README.md.erb
```diff
@@ -285,7 +288,7 @@ There are multiple ways to run your program as a debuggee:
 Stop at program start | [`rdbg` option](https://github.com/ruby/debug#rdbg---open-or-rdbg--o-for-short) | [require](https://github.com/ruby/debug#require-debugopen-in-a-program) | [debugger API](https://github.com/ruby/debug#start-by-method)
 ---|---|---|---|
 Yes | `rdbg --open` | `require "debug/open"` | `DEBUGGER__.open`
-No | `rdbg --open --nonstop` | `require "debug/open_nonstop"` | `DEBUGGER__.open(nonstop: true)` 
+No | `rdbg --open --nonstop` | `require "debug/open_nonstop"` | `DEBUGGER__.open(nonstop: true)`
```
ko1, line 291: what is different?

ko1: ah, spacing at the end of line.

PR #703, opened 2022-07-11
Improve method breakpoint performance.

ko1:
With pending method breakpoints, track all method invocation to track
`method_added` and `singleton_method_added`.
Of course, this approach slows down the debuggee's performance.

This patch tracks all `method_added` and `singleton_method_added` methods
and further defined such methods for the classes/modules. This approach
doesn't have performance penalty any more.

This patch also fix the message at registering the pending method
breakpoint.

```ruby

binding.b do: 'b C#foo' unless ENV['NO_PENDING_BP']

def fib n
  if n<2
    n
  else
    fib(n-1)+fib(n-2)
  end
end

require 'benchmark'

Benchmark.bm{|x|
  x.report{ fib(35) }
}
```

Results:

```
Without pending breakpoints:

$ NO_PENDING_BP=1 exe/rdbg -n target.rb
       user     system      total        real
   0.929580   0.000000   0.929580 (  0.929682)

With pending breakpoints - Before this patch:

$ exe/rdbg -n target.rb
[1, 10] in target.rb
=>   1| binding.b do: <<~DBG
     2|   b C#foo
     3| DBG
     4|
     5| def fib n
     6|   if n<2
     7|     n
     8|   else
     9|     fib(n-1)+fib(n-2)
    10|   end
=>#0    <main> at target.rb:1
(rdbg:binding.break) b C#foo
uninitialized constant C

C
^
#0  BP - Method (pending)  C#foo
       user     system      total        real
   3.276217   0.000000   3.276217 (  3.276299)

With pending breakpoints - After this patch:

[master]$ exe/rdbg -n target.rb
[1, 10] in target.rb
=>   1| binding.b do: <<~DBG
     2|   b C#foo
     3| DBG
     4|
     5| def fib n
     6|   if n<2
     7|     n
     8|   else
     9|     fib(n-1)+fib(n-2)
    10|   end
=>#0    <main> at target.rb:1
(rdbg:binding.break) b C#foo
Unknown name `C` for `Object`
#0  BP - Method (pending)  C#foo
       user     system      total        real
   0.933402   0.002353   0.935755 (  0.935757)
```

### DIFF
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -1643,7 +1643,7 @@ module DEBUGGER__
       b = tp.binding
       if var_name = b.local_variables.first
         mid = b.local_variable_get(var_name)
-        unresolved = false
+        resolved = true
 
         @bps.each{|k, bp|
           case bp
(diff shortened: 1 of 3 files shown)


lib/debug/session.rb
```diff
@@ -1654,15 +1654,53 @@ def method_added tp
...
       end
     end
 
+    class ::Module
+      undef method_added
+      def method_added mid; end
+      def singleton_method_added mid; end
+    end
+
+    def self.method_added_tracekr_create mod, method_added_id, method_accessor = :method
```
st0012, line 1679: Should this be `method_added_tracker_create` instead?

lib/debug/session.rb
```diff
@@ -1654,15 +1654,53 @@ def method_added tp
...
+      m = mod.__send__(method_accessor, method_added_id)
+      METHOD_ADDED_TRACKERS[m] = TracePoint.new(:call) do |tp|
+        SESSION.method_added tp
+      end
+    end
+
+    def self.method_added_tracekr_activate
```
st0012, line 1686: `method_added_tracker_activate`?
I'd also recommend putting the verb `activate` in front of the method. So it'd be `activate_method_added_tracker`.

ko1:
Thank you for the typo report. I fixed it.

> I'd also recommend putting the verb activate in front of the method. So it'd be activate_method_added_tracker

The name `foo_create` `foo_activate` I intended that the receiver is `foo`, but it seems confusing. fixed.

st0012:
> The name foo_create foo_activate I intended that the receiver is foo, but it seems confusing. fixed.

It's not confusing. But in English we usually say `activate the breakpoint` or `install the gem`. So naming methods as `verb_the_receiver` like `active_method_added_tracker` reads more similar to that.

PR #720, opened 2022-07-16
Update protocol tests

st0012:
1. Add termination tests
1. Test CDP and DAP's disconnection separately because DAP has more options to cover
1. Remove obsolete detaching tests because of ^
1. Remove unnecessary restart tests (they actually doesn't cover the feature)

This PR doesn't fix the flaky tests because `send_dap_request` still waits for the response of `disconnect` request. But that request is not deterministic as the debugger can execute the `kill!` command (and exit) before returning the response. I hope #721 can address the issue.


test/support/protocol_test_case.rb
```diff
@@ -412,6 +399,9 @@ def attach_to_cdp_server
       sleep 0.001 while @reader_thread.status != 'sleep'
       @reader_thread.run
       INITIALIZE_CDP_MSGS.each{|msg| send(**msg)}
+    end
+
+    def assert_cdp_connected
```
ono-max, line 404: I don't think that it's not a good idea to divide this process into a method because it could cause bugs when using frames after calling `attach_to_cdp_server`.

test/support/protocol_test_case.rb
```diff
@@ -155,20 +155,6 @@ def req_terminate_debuggee
       close_reader
     end
 
-    def assert_reattach
```
ono-max, line 158: If you want to use `req_disconnect` and `attach_to_cdp_server`, please move them into API block and update document.

https://github.com/ruby/debug/blob/master/test/support/protocol_test_case.rb#L41

test/protocol/disconnect_dap_test.rb
```diff
@@ -0,0 +1,57 @@
...
+       9|   end
+      10|   Bar.a
+      11|   bar = Bar.new
+      12| end
+    RUBY
+
+    def test_disconnect_without_terminateDebuggee_keeps_debuggee_alive
+      run_protocol_scenario PROGRAM, cdp: false do
+        req_dap_disconnect(terminate_debuggee: false)
+        attach_to_dap_server
```
ono-max, line 26: If you want to use `attach_to_dap_server`, please move please move it into API block and update document.

https://github.com/ruby/debug/blob/master/test/support/protocol_test_case.rb#L41

test/protocol/disconnect_dap_test.rb
```diff
@@ -0,0 +1,57 @@
...
+      send_dap_request "pause", threadId: 1
+    end
+
+    def assert_reattached
+      res = find_crt_dap_response
+      result_cmd = res.dig(:command)
+      assert_equal 'configurationDone', result_cmd
+    end
+  end
+end
```
ono-max, line 57: Can we move these method into test/support/protocol_test_case.rb?

These methods use internal methods, and I don't want to use them in the test cases.

ono-max (commented):
I left some comments.

test/support/protocol_test_case.rb
st0012: Sorry I don't understand this. Can you give me an example scenario?

test/support/protocol_test_case.rb
st0012: 👍 done

test/protocol/disconnect_dap_test.rb
st0012: 👍 done

test/support/protocol_test_case.rb
ono-max: Here is a scenario:
1. call `attach_to_cdp_server`.
2. call something that uses `@crt_frames`
3. `@crt_frames` is not the latest frame because it's always changed every time 'Debugger.paused' is called.

CONTRIBUTING.md
```diff
@@ -320,6 +320,18 @@ Execute debugging `program` with `&scenario`. If you want to test it only for DA
 
 `run_protocol_scenario program, cdp: false ...`
 
+- attach_to_dap_server
```
ono-max, line 323: Thank you

test/protocol/disconnect_dap_test.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    def test_disconnect_with_terminateDebuggee_kills_debuggee
+      run_protocol_scenario PROGRAM, cdp: false do
+        req_dap_disconnect(terminate_debuggee: true)
+      end
+    end
+
+    private
+
+    def req_dap_disconnect(terminate_debuggee:)
```
ono-max, line 42: How about this one?

* Divide `req_disconnect` to `req_disconnect_dap_server` and `req_disconnect_cdp_server`.
* Configure `terminate_debuggee: false` in `req_disconnect_dap_server` as a param

test/protocol/disconnect_dap_test.rb
ono-max: Thank you for explaining the reason to me.

test/protocol/disconnect_dap_test.rb
```diff
@@ -0,0 +1,57 @@
...
+      send_dap_request 'disconnect', terminateDebuggee: terminate_debuggee
+      close_reader
+    end
+
+    def suspend_debugee
+      send_dap_request "pause", threadId: 1
+    end
+
+    def assert_reattached
```
ono-max, line 51: Since we introduce `attach_to_cdp_server` and `attach_to_dap_server` as public API, we can remove them from `assert_reattach`. Therefore, here is my suggestion.

* Remove `attach_to_cdp_server` and `attach_to_dap_server` from `assert_reattach`.
https://github.com/ruby/debug/blob/master/test/support/protocol_test_case.rb#L158
* Then, we don't have to define `assert_reattached` here.

test/protocol/disconnect_dap_test.rb
ono-max: I agree with `suspend_debugee`, but I'd like to discuss the others. Could you take a look at the threads? Thx.

test/protocol/disconnect_dap_test.rb
st0012: I can do that. But I didn't define `assert_reattached` because `ProtocolTestCase` used to have it. I use it to abstract the underlying code, so the `test_disconnect_without_terminateDebuggee_keeps_debuggee_alive` test case would be easier to understand. 

I think this

```rb
attach_to_dap_server
assert_reattached
suspend_debugee
req_terminate_debuggee
```

is better than.

```rb
attach_to_dap_server
res = find_crt_dap_response
result_cmd = res.dig(:command)
assert_equal 'configurationDone', result_cmd
suspend_debugee
req_terminate_debuggee
```

Wdyt?

test/support/protocol_test_case.rb
st0012: I've removed it 👍

ono-max: Thank you.

test/support/protocol_test_case.rb
```diff
@@ -50,6 +50,51 @@ def run_protocol_scenario program, dap: true, cdp: true, &scenario
...
+      @id = 1
+      @reader_thread = Thread.new do
+        while res = @web_sock.extract_data
+          @queue.push res
+        end
+      rescue Detach
+      end
+      sleep 0.001 while @reader_thread.status != 'sleep'
+      @reader_thread.run
+      INITIALIZE_CDP_MSGS.each{|msg| send(**msg)}
```
ono-max, line 85: Please add the following lines here.
https://github.com/ruby/debug/blob/master/test/support/protocol_test_case.rb#L415-L416

ono-max: The reason for adding this here is as follows:

1. call `attach_to_cdp_server` **directory**.
2. call something that uses @crt_frames
3. @crt_frames is not the latest frame because it may be changed every time 'Debugger.paused' is called.

test/protocol/disconnect_dap_test.rb
ono-max: I agree with that. Thank you 👍

test/support/protocol_test_case.rb
st0012: I think `Debugger.paused` is only returned at the first connection, not the second?

```
Loaded suite test/protocol/disconnect_cdp_test
Started
start initial attach
{:id=>2, :result=>{}}
{:method=>"Runtime.executionContextCreated", :params=>{:context=>{:id=>"eae712537be02fa944b64faad339a2ab", :origin=>"http://127.0.0.1:63585", :name=>""}}}
{:id=>3, :result=>{}}
{:id=>4, :result=>{}}
{:id=>8, :result=>{}}
{:method=>"Debugger.scriptParsed", :params=>{:scriptId=>"1", :url=>"/private/var/folders/yg/hnbymwxd5pn7v94_clc59y6r0000gn/T/debug-20221106-84811-rdzgps.rb", :startLine=>0, :startColumn=>0, :endLine=>12, :endColumn=>0, :executionContextId=>1, :hash=>"2195685978487654025"}}
{:method=>"Debugger.paused", :params=>{:reason=>"other", :callFrames=>[{:callFrameId=>"0fb3aefabd50b783487268956071f58c", :functionName=>"<main>", :functionLocation=>{:lineNumber=>0, :scriptId=>"1"}, :location=>{:lineNumber=>0, :scriptId=>"1"}, :url=>"/private/var/folders/yg/hnbymwxd5pn7v94_clc59y6r0000gn/T/debug-20221106-84811-rdz
gps.rb", :scopeChain=>[{:type=>"local", :object=>{:type=>"object", :objectId=>"0.449851775437477"}}, {:type=>"script", :object=>{:type=>"object", :objectId=>"0.07042992169684925"}}, {:type=>"global", :object=>{:type=>"object", :objectId=>"0.7769977982489024"}}], :this=>{:type=>"object"}}]}}
end initial attach
disconnected
start second attach
{:id=>2, :result=>{}}
{:method=>"Runtime.executionContextCreated", :params=>{:context=>{:id=>"bfb3534d760e52e0db4d88a84ea031a4", :origin=>"http://127.0.0.1:63585", :name=>""}}}
{:id=>3, :result=>{}}
{:id=>4, :result=>{}}
{:id=>8, :result=>{}}
F
=============================================================================================================================================================================================================================================================================================================================================
Failure: test_closing_cdp_connection_doesnt_kill_the_debuggee(DEBUGGER__::DisconnectCDPTest):
```

That's why I didn't move those 2 lines as it'll break reconnected cases like `test/protocol/disconnect_cdp_test`

ono-max: Oh, seriously? Thank you for checking it. I'll also check it later.

ono-max: I just checked the behavior, and you are right. Also, because I noticed that this behavior is a bug, I need to fix it, too.

ono-max: If it's OK with you, could you write comments somewhere as follows?

```
TODO: Chrome DevTools doesn't support reattaching. We need to fix it.
```

ono-max:
I think that this PR looks good to me 👍 

Thank you for your great work!

test/support/protocol_test_case.rb
st0012: Sure. But should I also remove the reattaching part from `test/protocol/disconnect_cdp_test.rb` as it's actually not working?

ono-max: > But should I also remove the reattaching part from test/protocol/disconnect_cdp_test.rb as it's actually not working?

If you can, yes.

st0012: I found that we still need to reattach to the CDP server in disconnection test because we need to terminate it. But I added the comment right above the attaching call so others can understand that's not how actual reattaching works.

ono-max: Thank you!

ko1:
> Remove unnecessary restart tests (feature not supported)

Please remain this test because user can push the *restart* button on the VSCode.
And I'll merge this PR.

ono-max:
Sorry, using `req_terminate_debuggee` is my mistake. Here is my suggestion to test restart meaningfully:

1. Define the following method such as follows:
```ruby
def req_restart
      case get_target_ui
      when 'vscode'
        send_dap_request 'terminate', restart: true
      end
    end
```
2. Use it in restart test.

WDYT?

st0012:
@ono-max If we currently only support `vscode` for restart, I'd prefer not adding `req_restart` yet. Otherwise other devs need to check its definition to find out CDP actually is not supported by it.

ono-max:
Ok, so how about this one? we define `req_restart` method as you did here.

https://github.com/ruby/debug/pull/720/files#diff-172faac32ddd51787eb1039d2109dc1a698cae3d3753c8140c36a3ad9aeb9940R42

st0012:
> Please remain this test because user can push the restart button on the VSCode.

After some testing, I think DAP's restart actually doesn't work.

https://user-images.githubusercontent.com/5079556/201910153-43c812d4-190b-435c-9663-b85702c1c82d.mov

The last DAP request is:

```
#11078:[>] {"command":"disconnect","arguments":{"restart":true,"terminateDebuggee":true},"type":"request","seq":31}
#11078:[<] {"type":"response","command":"disconnect","request_seq":31,"success":true,"message":"Success","seq":37}
```

Based on the [specification](https://microsoft.github.io/debug-adapter-protocol/specification#leftwards_arrow_with_hook-restart-request), if we're not supporting the restart request yet ([commented out](https://github.com/ruby/debug/blob/1ecb014d44fb0d851827a1272e86673eb28e918b/lib/debug/server_dap.rb#L184)), we need to emulate the restarting behavior by starting a new debuggee manually in the server.

And we can see that DAP server currently doesn't handle the `restart` argument:

https://github.com/ruby/debug/blob/1ecb014d44fb0d851827a1272e86673eb28e918b/lib/debug/server_dap.rb#L354-L371

So it appears that we actually don't support restart yet?

ko1:
Now "launch" mode "Restart Debug" command (from menu) seems working.
On "attach" mode it doesn't (DAP support).

The spec https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Disconnect doesn't say anything I think.

```
  /**
   * A value of true indicates that this `disconnect` request is part of a
   * restart sequence.
   */
  restart?: boolean;
```

st0012:
> Now "launch" mode "Restart Debug" command (from menu) seems working.

Yes but we can't cover this case atm because we can only test attach mode in protocol tests. This means the current test tries to test a feature we're not supporting yet. And it is passing because it actually just test disconnection, which we already covered in other tests. That's why I want to delete that test.

ko1:
ok I merge it.

PR #722, opened 2022-07-17
Activate debug gem

nobu:
When loading `debug/start.rb` from `rdbg`, the debug gem is not activated properly, and cannot find the `debug/debug` extension library under the gem `extensions` directory.
By this activation, the path is added to `$LOAD_PATH`.


ko1:
We discussed this issue with @nobu and concluded it is not needed.

Issue #526, opened 2022-02-12
Proposal - show_source-ish command


st0012:
## Naming Options
- `show-source`
  - I prefer this because:
    1. I used to use `pry` a lot so I'm used to it.
    2. Basically impossible to conflict with user's variable/method name.
- `show_source`
- something else
  - If we go for this option, a one-word name would probably be better.
  - But to get people onboard more easily, I think an alias to `show-source` or `show_source` may still be needed.

## Interface

- `show-source` - same as `show-source self`
- `show-source obj` - show the object's class
- `show-source obj.method` - show the method's definition
- `show-source class` - show class definition
- `show-source class#method` - show the method's definition
- `show-source class.method` - show the method's definition

## Output Format

**pry**

```
[2] pry(main)> show-source Bar#foo

From: foo.rb:6:
Owner: Bar
Visibility: public
Signature: foo()
Number of lines: 3

def foo
  super
end
```

**irb**

```
irb(main):001:0> show_source "Bar#foo"

From: foo.rb:6

  def foo
    super
  end

=> nil
```

You can see that `pry` provides a nicer output. But we can probably start from something simpler, like `irb`'s.

### Support for  `super`

`pry`'s `show-source`  also supports using `-s` option to access the method's super definition. 

```rb
class Foo
  def foo; end
end

class Bar < Foo
  def foo
    super
  end
end
```

```
[2] pry(main)> show-source Bar#foo


Owner: Bar
Visibility: public
Signature: foo()
Number of lines: 3

def foo
  super
end
[3] pry(main)> show-source Bar#foo -[]()s

From: foo.rb:2:
Owner: Foo
Visibility: public
Signature: foo()
Number of lines: 1

def foo; end
```

You can even use multiple `s` to access farther definitions. And it'd print a warning when there's no more superclasses.

```
[4] pry(main)> show-source Bar#foo -ss
Error: No superclass found for #<UnboundMethod: Bar#foo() foo.rb:6>
```

(I think `irb`'s `show_source` currently doesn't support this feature.)

I believe this is something we should support from the beginning. But currently the debugger dosn't take command options like  `-opt` and I don't think we'll support that in the near future. So we need to come up with another way to do it. 

Perhaps `show-source Bar#foo level:[num]` ?

- `show-source Bar#foo level:0` - shows `Bar#foo`
- `show-source Bar#foo level:1` - shows `Foo#foo`

## Looking for Feedback

- Name?
[...]

PR #724, opened 2022-07-20
[POC] Command Registration

st0012:
### Extensible

Users or gems will be able to register new commands with:

```rb
DEBUGGER__::Session.register_command("foo", aliases: ["fo"]) do |arg|
  # my command
end.document category: "My Gem", content: <<~MD
  * foo does blablabla
MD
```

And then the command document will appear in the help 

```
❯ ruby -Ilib -rdebug target.rb
[4, 9] in target.rb
     4|   # my command
     5| end.document category: "My Gem", content: <<~MD
     6|   * foo does blablabla
     7| MD
     8|
=>   9| binding.b
=>#0    <main> at target.rb:9
(rdbg) help foo    # command

### My Gem

* foo does blablabla
(rdbg)
```

And because commands are stored together, it can support autocompletion in the future too.

This extensibility will allow:

1. Integration with popular libraries, like Rails integration demonstrated in the comments below.
2. 3rd-party command implementations. We can have commands like [show-source](https://github.com/ruby/debug/issues/526) implemented in a 3rd-party gem instead of implementing everything in the debugger core. (not saying that `show-source` should be implemented externally, but just an example).

### Easier to maintain

Currently, command documents need to be written in Ruby comments with a strict format. And the parsing logic is quite adhoc for that reason. With this approach, we can simplify the parsing logic and make the document naturally a part of the command definition.

## Additional thoughts

- The `CommandSet` logic is a replication of the `Config` class, which prevents sharing mutable hash while allowing modifying values. I hope this will make it Ractor-safe?

### DIFF
--- a/exe/rdbg
+++ b/exe/rdbg
@@ -21,6 +21,7 @@ when :start
   exec(env, cmd, *ARGV)
 
 when :attach
+  require_relative "../lib/debug/session"
   require_relative "../lib/debug/client"
   ::DEBUGGER__::CONFIG.set_config(**config)
 
--- a/lib/debug/config.rb
+++ b/lib/debug/config.rb
@@ -471,32 +471,15 @@ module DEBUGGER__
 
   def self.parse_help
     helps = Hash.new{|h, k| h[k] = []}
-    desc = cat = nil
     cmds = Hash.new
 
-    File.read(File.join(__dir__, 'session.rb'), encoding: Encoding::UTF_8).each_line do |line|
-      case line
-      when /\A\s*### (.+)/
-        cat = $1
-        break if $1 == 'END'
-      when /\A\s*register_command (.+)/
-        next unless cat
-        next unless desc
-        ws = $1.split(/,\s*/).map{|e| e.gsub('\'', '')}
-        helps[cat] << [ws, desc]
-        desc = nil
-        max_w = ws.max_by{|w| w.length}
-        ws.each{|w|
-          cmds[w] = max_w
-        }
-      when /\A\s+# (\s*\*.+)/
-        if desc
-          desc << "\n" + $1
-        else
-          desc = $1
-        end
+    DEBUGGER__::Session::COMMAND_SET.values.uniq.each do |command|
+      command.names.each do |name|
+        cmds[name] = command.name
       end
+      helps[command.category] << [command.names, command.doc]
     end
+
     @commands = cmds
     @helps = helps
   end
--- a/lib/debug/config.rb
+++ b/lib/debug/config.rb
@@ -479,7 +479,7 @@ module DEBUGGER__
       when /\A\s*### (.+)/
         cat = $1
         break if $1 == 'END'
-      when /\A      when (.+)/
+      when /\A\s*register_command (.+)/
         next unless cat
         next unless desc
         ws = $1.split(/,\s*/).map{|e| e.gsub('\'', '')}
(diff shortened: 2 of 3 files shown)


ko1:
Do you want to invoke some Ruby code on the session thread?

Could you list the example usages?
This is a kind of language design and examples are very important.

st0012:
### Example - fetching Rails configurations via `rails config`

**Command Code**

```rb
COMMAND_OPTIONS_SPLITTING_REGEXP = /([^\s]+)(?:\s+(.+))?/

DEBUGGER__::Session.register_command("rails") do |arg|
  COMMAND_OPTIONS_SPLITTING_REGEXP =~ arg
  sub_command, options = $1, $2

  case sub_command
  when "config"
    config = ::Rails.application.config

    configs =
      if options
        if config.respond_to?(options)
          config.send(options)
        else
          @ui.puts("unknown Rails component #{options}")
          []
        end
      else
        config.instance_variables.each_with_object({}) do |ivar, hash|
          key = ivar.to_s.sub("@", "")
          value = config.instance_variable_get(ivar)
          hash[key] = value
        end
      end

    configs.each do |key, value|
      key = colorize_cyan(key)
      @ui.puts("  #{key} => #{DEBUGGER__.safe_inspect(value)}")
    end
  else
    @ui.puts("unknown Rails debuggger command: #{arg}")
  end

  :retry
end.document category: "Rails", content: <<~MD
  * `rails config`
    * List top-level Rails configurations
  * `rails config <component>`
    * List <component>'s configurations
MD

```

**Result**

截圖 2022-07-31 13 15 10


Other useful command extensions for Rails could be:

- `rails breadcrumbs` - See all the recorded [`ActiveSupport::Notifications`](https://api.rubyonrails.org/classes/ActiveSupport/Notifications.html) events.
    - Of course this needs extra work to store/access the events but it's relatively easy to achieve without debugger's support.
    - The breadcrumbs concept exists in many error monitoring services like [Sentry](https://docs.sentry.io/product/issues/issue-details/breadcrumbs/). It'll help users track major application events happened before the program reaches the current state
- `rails callbacks Model ` - List a model's callbacks by running `Model._save_callbacks` and filter/display the result nicely
- `rails routes /regexp/` - List the application's routes

@rafaelfranca is there other Rails specific debugger commands you'd like to see as well?

ko1:
So they are global information.

```ruby
# define SOME_CLASS.some_method before

eval SOME_CLASS.some_method
```

doesn't work?

ko1:
I didn't mean "use eval" on this but providing alias feature like:

```
alias rails "eval SOME_CLASS.some_method(args)"
```

with documentation.

Ah, `@ui.puts` is not work well. adding eval option?
Anyway, I don't think it is good idea to expose `@ui.puts` for extensions.

st0012:
> alias rails "eval SOME_CLASS.some_method(args)"

That could work too. But I think `alias` is confusing and something like `command` should be clearer, along with an argument for documentation

```rb
command :rails, "eval RailsDebugger.execute(args)", doc: <<~DOC
  * `rails config`
    * List top-level Rails configurations
  * `rails config <component>`
    * List <component>'s configurations
DOC
```

> Anyway, I don't think it is good idea to expose @ui.puts for extensions.

Yeah it's not ideal to expose instance variables. But we still need to access the UI from the extension code. Do you think calling `DEBUGGER__::SESSION.print(msg)` will be acceptable?

PR #727, opened 2022-07-21
Avoid setting @tc in thread_begin event

### LINKED ISSUE #725: Deadlock on executing commands from binding.b in Rails app
**Your environment**

```bash
❯ ruby -v
ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-darwin20]
❯ rdbg -v
rdbg 1.6.1
```

**Describe the bug**

When using `binding.b` in our Rails app — in various places: code run from test suites, code run from serving web requests — the debug REPL will hang.

Here is an example session entering the debugger from unit tests:

```
.[5, 14] in ~/src/debug_hang/app/models/concerns/debug_hang/sample_concern.rb
     5|     def respond_to_missing?(method, include_private = false)
     6|       method.to_s.start_with?("test_") || super
     7|     end
     8|
     9|     def method_missing(method, *args, &block)
=>  10|       binding.b
    11|
    12|       if method.to_s =~ /^test_/
    13|         puts "Running test: #{method}"
    14|         super
=>#0	DebugHang::SampleConcern#method_missing(method=:hang_the_debugger, args=[], block=nil) at ~/src/debug_hang/app/models/concerns/debug_hang/sample_concern.rb:10
  #1	block in <class:ExampleTest> at ~/src/debug_hang/test/models/example_test.rb:5
  # and 25 frames (use `bt' command for all frames)
(ruby) false == true
false
(ruby) false == true
["DEBUGGER Exception: /Users/ian/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/debug-1.6.1/lib/debug/thread_client.rb:1108",
 #<fatal: No live threads left. Deadlock?
20 threads, 20 sleeps current:0x00007f8590c458b0 main thread:0x00007f85f6c06d30
* #<Thread:0x0000000105c30ce8 sleep_forever>
   rb_thread_t:0x00007f85f6c06d30 native:0x0000000106beae00 int:0
```

[full logfile](https://github.com/ruby/debug/files/9160304/hang.log)

**To Reproduce**

```bash
git clone git@github.com:ianterrell/debug_hang.git
cd debug_hang
bundle install
rails db:create db:migrate
rails t
```

From debugger, repeatedly enter `false == true`.

**Expected behavior**

No deadlock.

**Additional context**

- This looks like it could be related to #248

st0012:
This event will be triggered before the new thread even has a `ThreadClient`. So in this case, the `@tc` will be overridden to just `nil`,
which is wrong.

However, we also don't want to just set `@tc` to the new thread's newly created `ThreadClient`. This is because a thread's creation can happen during a subsession, and promptly switching `@tc` to another thread's `ThreadClient` will misfire thread suspension.

Fixes #725

### DIFF
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -213,18 +213,24 @@ module DEBUGGER__
 
     def process_event evt
       # variable `@internal_info` is only used for test
-      @tc, output, ev, @internal_info, *ev_args = evt
+      tc, output, ev, @internal_info, *ev_args = evt
 
       output.each{|str| @ui.puts str} if ev != :suspend
 
-      case ev
-
-      when :thread_begin # special event, tc is nil
+      # special event, tc is nil
+      # and we don't want to set @tc to the newly created thread's ThreadClient
+      if ev == :thread_begin
         th = ev_args.shift
         q = ev_args.shift
         on_thread_begin th
         q << true
 
+        return
+      end
+
+      @tc = tc
+
+      case ev
       when :init
         enter_subsession
         wait_command_loop
--- a/test/console/thread_test.rb
+++ b/test/console/thread_test.rb
@@ -1,0 +1,24 @@
+require_relative '../support/console_test_case'
+
+module DEBUGGER__
+  class ThreadControlTest < ConsoleTestCase
+    def program
+      <<~RUBY
+       1| def foo
+       2|   Thread.new { sleep 5 }
+       3| end
+       4|
+       5| 5.times do
+       6|   foo
+       7|   binding.b(do: "1 == 2") # eval Ruby code in debugger
+       8| end
+      RUBY
+    end
+
+    def test_debugger_isnt_hung_by_new_threads
+      debug_code(program) do
+        type "c"
+      end
+    end
+  end
+end


ko1:
https://github.com/ruby/debug/commit/00ec229cb46e5c9059c0285a4a4de5f6ac322292
It's my fault can not checking this bug.

ko1:
Could you add a test program or repro- small code?

ko1:
I'm not sure why `@tc` should not be rewritten yet.

1. Stop at TC1: `@tc = TC1`
2. eval `Thread.new{}` on TC1
3. receive `thread_begin` and `@tc = nil`
4. receive `result` event and `@tc = T1` (maybe)

In this case, there is no issue. Could you describe what's happens?

ko1:
ah, https://github.com/ruby/debug/blob/master/lib/debug/session.rb#L1537 this comparison requires `@tc`.

ko1:
Another threads can resume by some events so that current code is unstable for that even if the patch is merged.
Anyway I'll merge it (and modify them soon) if you wrote can write a test code.

st0012:
I've added a test that'd hang and timeout on master.

ianterrell:
@ko1 @st0012 — Thanks so much for this fix, it really helps our workflows. I expect this is impacting a lot of Rails apps in the wild. Is there anything else I can do to help get this merged in and released?

PR #732, opened 2022-08-02
Check if HOME can change or not instead of ENV["CI"]

znz:
Tests of snapshot tarball are failed, because the CIs run without write permissions to HOME.
So I want to check the desired feature instead of `ENV["CI"]`.

https://github.com/ruby/actions/runs/7617816773?check_suite_focus=true#step:17:2135
```
  Failure: test_frame_filtering_works_with_unexpanded_path_and_expanded_skip_path(DEBUGGER__::BasicBacktraceTest):
    Test::Unit::AssertionFailedError: 
    --------------------
    | Debugger Session |
    --------------------
    
    > [1, 2] in /tmp/debug-20220801-16442-89w16l.rb
    > =>   1| load "~/foo.rb"
    >      2| Foo.new.bar
    > =>#0	<main> at /tmp/debug-20220801-16442-89w16l.rb:1
    > file = File.open('/home/runner/foo.rb', 'w+') { |f| f.write('class Foo; def bar; debugger; end; end') }
    > [?2004h(rdbg) file = File.open('/home/runner/foo.rb', 'w+') { |f| f.write('class Foo; def bar; debugger; end; end') }
    > [?2004l
  eval error: Permission denied @ rb_sysopen - /home/runner/foo.rb
    >   (rdbg)//tmp/debug-20220801-16442-89w16l.rb:1:in `initialize'
    >   (rdbg)//tmp/debug-20220801-16442-89w16l.rb:1:in `open'
    >   (rdbg)//tmp/debug-20220801-16442-89w16l.rb:1:in `<main>'
    > nil
    > c
    > [?2004h(rdbg) c
    > [?2004l
  /tmp/debug-20220801-16442-89w16l.rb:1:in `load': cannot load such file -- ~/foo.rb (LoadError)
    > 	from /tmp/debug-20220801-16442-89w16l.rb:1:in `<main>'
```

### DIFF
--- a/test/support/console_test_case.rb
+++ b/test/support/console_test_case.rb
@@ -9,12 +9,23 @@ module DEBUGGER__
       warn "Tests on local and remote. You can disable remote tests with RUBY_DEBUG_TEST_NO_REMOTE=1."
     end
 
+    HOME_CANNOT_CHANGE =
+      begin
+        pwd = Dir.pwd
+        ruby = ENV['RUBY'] || RbConfig.ruby
+        home_cannot_change = false
+        PTY.spawn({ "HOME" => pwd }, ruby, '-e', 'puts ENV["HOME"]') do |r,|
+          home_cannot_change = r.gets.chomp != pwd
+        end
+        home_cannot_change
+      end
+
     class << self
       attr_reader :pty_home_dir
 
       def startup
         @pty_home_dir =
-          if ENV["CI"]
+          if HOME_CANNOT_CHANGE
             # CIs usually doesn't allow overriding the HOME path
             # we also don't need to worry about adding or being affected by ~/.rdbgrc on CI
             # so we can just use the original home page there
@@ -25,7 +36,7 @@ module DEBUGGER__
       end
 
       def shutdown
-        unless ENV["CI"]
+        unless HOME_CANNOT_CHANGE
           FileUtils.remove_entry @pty_home_dir
         end
       end
--- a/test/support/console_test_case.rb
+++ b/test/support/console_test_case.rb
@@ -9,7 +9,11 @@ module DEBUGGER__
       warn "Tests on local and remote. You can disable remote tests with RUBY_DEBUG_TEST_NO_REMOTE=1."
     end
 
-    HOME_CANNOT_CHANGE =
+    # CIs usually doesn't allow overriding the HOME path
+    # we also don't need to worry about adding or being affected by ~/.rdbgrc on CI
+    # so we can just use the original home page there
+    USE_TMP_HOME =
+      !ENV["CI"] ||
       begin
         pwd = Dir.pwd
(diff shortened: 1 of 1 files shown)


st0012:
@znz Sorry that @ko1 reminded me this may be happening but I forgot to follow up.

But I'm confused by GH Action's behavior because its document says [`ENV["CI"]`  should always be true](https://docs.github.com/en/actions/learn-github-actions/environment-variables), which applies to this repo. And we can also see it set in the build's `Run env | sort` step. However it's not picked up by the tests 🤔

znz:
This fails because `chmod -w $HOME` at <https://github.com/ruby/actions/blob/9d0cf8fffecac5f9c8b8600e452021b87458d974/.github/workflows/snapshot-master.yml#L143> to check that tests does not affect normal environments.
So I want to write nothing under `$HOME` on CI too.

st0012:
Thanks for explanation 🙏 

However, this change also allows modifying `$HOME` when it's run locally, which was the patch originally trying to prevent.

```
❯ cat ~/.rdbgrc
foo

❯ be ruby test/console/rdbg_option_test.rb -n /rc_file/
Calling `DidYouMean::SPELL_CHECKERS.merge!(error_name => spell_checker)' has been deprecated. Please call `DidYouMean.correct_error(error_name, spell_checker)' instead.
Calling `DidYouMean::SPELL_CHECKERS.merge!(error_name => spell_checker)' has been deprecated. Please call `DidYouMean.correct_error(error_name, spell_checker)' instead.
Loaded suite test/console/rdbg_option_test
Started
..
Finished in 0.49796 seconds.
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
2 tests, 12 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
4.02 tests/s, 24.10 assertions/s

❯ cat ~/.rdbgrc
cat: /Users/st0012/.rdbgrc: No such file or directory

```

I think we can summarize our needs like this:

- If it's a local environment, we don't want to modify `$HOME` 
- If it's a CI environment, it's fine to write files in `$HOME`
    - However, in `ruby/actions`'s CI we also don't want `$HOME` to be writable

So may be the conditions should be this instead?

```rb
USE_TMP_HOME = !ENV["CI"] || HOME_CANNOT_CHANGE

# startup
@pty_home_dir =
  if USE_TMP_HOME
    Dir.mktmpdir
  else
    Dir.home
  end

# shutdown
if USE_TMP_HOME
  FileUtils.remove_entry @pty_home_dir
end
```

znz:
Thank you for your summarize.
I update the conditions.

st0012:
I can confirm it's not affecting local `~/.rdbgrc` anymore. Sorry again for the trouble and thanks for the fix 😄

PR #733, opened 2022-08-10
Set path to debug.so

nobu:
## Description
When starting from rdbg, the debug gem is not activated, and cannot find the debug/debug extension library under the gem extensions directory.
As well as `rdbg` passes the path to debug.rb, also the path to the extension library should be passed.

### DIFF
--- a/exe/rdbg
+++ b/exe/rdbg
@@ -13,9 +13,24 @@ when :start
   start_mode = config[:remote] ? "open" : 'start'
   cmd = config[:command] ? ARGV.shift : (ENV['RUBY'] || RbConfig.ruby)
 
+  if defined?($:.resolve_feature_path)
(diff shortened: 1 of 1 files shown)


exe/rdbg
```diff
@@ -13,9 +13,22 @@ when :start
...
 
+  if defined?($:.resolve_feature_path)
+    begin
+      _, sopath = $:.resolve_feature_path('debug/debug.so')
+    rescue LoadError # before 3.1
+    else
+      sopath = File.dirname(File.dirname(sopath)) if sopath
+    end
+  else
+    so = "debug/debug.#{RbConfig::CONFIG['DLEXT']}"
```
ko1, line 24: can you add a comment like that "for 2.6"?

Issue #739, opened 2022-08-26
Feature request: Stop debugger until first client connects


jvilk-stripe:
**Your proposal**

Make it possible for the debugger to pause the program until the first client connects:

```ruby
require "debug/open_wait_for_client"
# Program is now paused until the first debugger client connects
```

Once the first client connects, the program can proceed.

**Additional context**

We are working to integrate `debug.rb` into Stripe's Ruby test runner. The test runner needs to start the server and wait for VS Code to connect before running the tests.

`debug/open` does not work for this use case. `debug/open` pauses the program when the debugger starts, but does not resume the program after VS Code connects. Instead, VS Code reports the program as paused on a breakpoint in our test runner code. The user then has to hit "resume" to run their tests and hit the breakpoints that they actually set.

st0012:
In the use case you described, do you always have breakpoints set in your tests? 

If the answer is yes, then using `debug/open_nonstop` will skip the initial stop, but will wait for attaching once it hits a breakpoint.
If the answer is no, then I don't get the need to connect to the VSCode since the program will just finish.

jvilk-stripe:
> In the use case you described, do you always have breakpoints set in your tests?

Yes -- in VS Code.

> If the answer is yes, then using debug/open_nonstop will skip the initial stop, but will wait for attaching once it hits a breakpoint.

The problem is that the breakpoints get set just after VS Code attaches. VS Code is doing the breakpoint setting.

ko1:
Hi, I understand your request that

* On the VSCode, you setup breakpoints on the editor. You want to stop on that.
* You don't want to launch tests from VSCode (F5 key to launch debug process) maybe because there are some reasons.
* You want to invoke your test runner in your terminal or somewhere you want.
* You want to attach VScode just before your tests run.
* You don't want to stop the program at attaching timing, but you want to stop at breakpoints you setup on VSCode.

Right?

Can I ask something?

* There are some options to solve this issue.
  * 1. add an option to `#open` method like #747
  * 2. add an option to VSCode launch.json (nonstop on attach)
  * 3. add an option to load breakpoints setup by VSCode on loading timing of debug.gem 

Maybe (3) is most easy way because you don't need to launch VSCode and attach to the debuggee if no breakpoints are touched. 
In fact, vscode-rdbg supports to export breakpoints.

![image](https://user-images.githubusercontent.com/9558/190585019-280ea47e-b06c-4ea5-926c-af9138702b14.png)

[...]

PR #747, opened 2022-09-02
✨ Add on_client_connected to Session.open, to allow for synchronization

marianosimone:
## Description
Add a way for callers to get notified when a client gets connected to the debugger.

This solves [Feature request: Stop debugger until first client connects · Issue #739 · ruby/debug](https://github.com/ruby/debug/issues/739).

A caller can now do something like:

```ruby
client_connected = Concurrent::IVar.new
DEBUGGER__.open(nonstop: true, on_client_connected: -> {client_connected.set(true)})
client_connected.wait()
```

## Additional context

We are working to integrate `debug.rb` into Stripe's Ruby test runner. The test runner needs to start the server and wait for VS Code to connect before running the tests.

`debug/open` does not work for this use case, as it pauses the program when the debugger starts, but does not resume the program after VS Code connects. Instead, VS Code reports the program as paused on a breakpoint in our test runner code. The user then has to hit "resume" to run their tests and hit the breakpoints that they actually set.

### DIFF
--- a/lib/debug/server.rb
+++ b/lib/debug/server.rb
@@ -7,10 +7,12 @@ require_relative 'version'
 
 module DEBUGGER__
   class UI_ServerBase < UI_Base
-    def initialize
+    # @param on_client_connected [lambda] a function to be called every time a client connects to the debugger
+    def initialize(on_client_connected: nil)
       @sock = @sock_for_fork = nil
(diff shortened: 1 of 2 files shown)


marianosimone:
Closing in favor of #792

Issue #748, opened 2022-09-06
Feature request: respect `skip_path` when returning backtraces


marianosimone:
**Your proposal**
The documentation for `skip_path` says:

> Skip showing/entering frames for given paths

However, it only works for entering frames while doing step-by-step debugging and displaying the backtrace through `backtrace` (as implemented [here](https://github.com/ruby/debug/blob/2cb44483c681f6806bf6843a8f996d3a10d0e42c/lib/debug/thread_client.rb#L665)). However, it does not work when requesting the stacktrace through DAP [here](https://github.com/ruby/debug/blob/2cb44483c681f6806bf6843a8f996d3a10d0e42c/lib/debug/server_dap.rb#L699-L732).

ko1:
#749 fixed it? Please reopen if I misunderstood.

PR #749, opened 2022-09-06
🐛 Respect skip_path when returning the backtrace in ServerDAP

marianosimone:
This makes the DAP implementation respect `skip_path`, to only return frames that are not skipped.

See https://github.com/ruby/debug/issues/748

### DIFF
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -702,9 +702,10 @@ module DEBUGGER__
         frames = []
         @target_frames.each_with_index do |frame, i|
           next if i < start_frame
-          break if (levels -= 1) < 0
 
           path = frame.realpath || frame.path
+          next if skip_path?(path)
+          break if (levels -= 1) < 0
           source_name = path ? File.basename(path) : frame.location.to_s
 
           if (path && File.exist?(path)) && (local_path = UI_DAP.remote_to_local_path(path))
--- a/test/protocol/call_stack_with_skip_dap_test.rb
+++ b/test/protocol/call_stack_with_skip_dap_test.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+require_relative '../support/protocol_test_case'
+
+module DEBUGGER__
+  class CallStackWithSkipDAPTest < ProtocolTestCase
+    def program(path)
+      <<~RUBY
+        1| require_relative "#{path}"
+        2| with_foo do
+        3|  "something"
+        4| end
+      RUBY
+    end
+
+    def extra_file
+      <<~RUBY
+        def with_foo
+          yield
+        end
+      RUBY
+    end
+
+    def req_stacktrace_file_names
+      response = send_dap_request('stackTrace', threadId: 1)
+      stack_frames = response.dig(:body, :stackFrames)
+      stack_frames.map { |f| f.dig(:source, :name) }
+    end
+
+    def test_it_does_not_skip_a_path
+      with_extra_tempfile do |extra_file|
+        run_protocol_scenario(program(extra_file.path), cdp: false) do
+          req_add_breakpoint 3
+          req_continue
+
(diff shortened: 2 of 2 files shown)


lib/debug/server_dap.rb
```diff
@@ -685,6 +685,8 @@ def register_vars vars, tid
   end
 
   class ThreadClient
+    include SkipPathHelper
```
st0012, line 688: The `ThreadClient` here is a patch to the one defined in `thread_client.rb`, so this should not be needed.

lib/debug/server_dap.rb
```diff
@@ -706,6 +708,7 @@ def process_dap args
 
           path = frame.realpath || frame.path
```
st0012, line 709: nit: we can call `skip_path?` right after this so we don't need to compute `source_name` and throw it away

lib/debug/server_dap.rb
marianosimone: Oh, you are right! Thanks for catching :)

lib/debug/server_dap.rb
marianosimone: Yes, we can :) changed it

marianosimone:
@ko1 there's a failing test, but I'm not sure it's related to my changes, as the same test failed in all my other PRs 😕

lib/debug/server_dap.rb
```diff
@@ -702,9 +702,10 @@ def process_dap args
         frames = []
         @target_frames.each_with_index do |frame, i|
           next if i < start_frame
-          break if (levels -= 1) < 0
 
           path = frame.realpath || frame.path
+          next if skip_path?(path)
+          break if (levels -= 1) < 0
```
marianosimone, line 708: This now needs to be after deciding to skip or not, or we'd return less frames that required

test/protocol/call_stack_with_skip_dap_test.rb
```diff
@@ -0,0 +1,467 @@
+# frozen_string_literal: true
+
+require_relative '../support/protocol_test_case'
+
+module DEBUGGER__
+  
+  class CallStackWithSkipDAPTest1664318269 < ProtocolTestCase
```
st0012, line 7: Can this be a request based test? I think we only need to cover the `stackTrace` request for this fix but using recorded test case adds a lot of noise.

You can send a `stackTrace` like [this](https://github.com/ruby/debug/blob/19b4dde3308f532943e4234d1588d4fa26c52345/test/support/protocol_test_case.rb#L176-L180).

test/protocol/call_stack_with_skip_dap_test.rb
```diff
@@ -0,0 +1,467 @@
...
+        def with_foo
+          yield
+        end
+      RUBY
+    end
+    
+    def execute_test(skip_path:)
+      with_extra_tempfile do |extra_file|
+        if skip_path
+          ENV['RUBY_DEBUG_SKIP_PATH'] = extra_file.path
```
st0012, line 28: I think this should be reset in a `ensure` block?

test/protocol/call_stack_with_skip_dap_test.rb
marianosimone: I went back and forth on this a couple of times. I ended up deciding for this way, because that's what [CONTRIBUTING.md](https://github.com/ruby/debug/blob/master/CONTRIBUTING.md#tests-for-dap-and-cdp) and `ProtocolTestCase` seem to imply, by not making `send_dap_request` part of the API... But 🤷🏽 I agree that this is much better (and probably the docs need to be updated)

st0012: We were in the process of converting most of the DAP/CDP tests into request-based, so we don't have all request helpers yet.

Issue #394, opened 2021-11-16
Change the log level for the fork interception


ccdredzik:
**Your proposal**

Stop logging it as warnings, start logging it as trace or debug.

**Additional context**

It obscures normal test execution.

st0012:
If you're talking about these messages: `"Detaching after fork ...."` and `"Attaching after process ..."`, I think they can be downgraded to `debug` in the future. But at the moment, the forking mechanism isn't completely stable and can cause issues like #336. Having this message can help users be aware of process forking and identify/report the problem.

ko1:
Does it hurts your test or your favorite?

JunichiIto:
When I run RSpec against Rails 7 application, I see this warning. It's noisy a little to read the result of tests.

```
$ bundle exec rspec -f progress
...................................................................DEBUGGER: Attaching after process 69034 fork to child process 69055
...

Finished in 3.26 seconds (files took 1.79 seconds to load)
70 examples, 0 failures
```

In addition, this warning is displayed in red when I run tests within RubyMine's terminal. I sometimes misunderstand that something is wrong with my application.

<img width="587" alt="Screen Shot 2021-12-31 at 10 10 05" src="https://user-images.githubusercontent.com/1148320/147803817-18cc5f65-9470-4141-8780-3eb73b1a8a56.png">

I'd be glad if it is suppressed by default.

PR #750, opened 2022-09-06
🔊 Reduce breakpoint activation logging from warn to info

marianosimone:
Similar to https://github.com/ruby/debug/issues/394, these logs are a bit noisy and not that useful for the average user:

```
DEBUGGER:  BP - Line  /<redacted>.rb:<line> (call) is activated.
```

Reducing to `info` makes it less noisy.

### DIFF
--- a/lib/debug/breakpoint.rb
+++ b/lib/debug/breakpoint.rb
@@ -196,7 +196,7 @@ module DEBUGGER__
       enable
 
       if @pending && !@oneshot
-        DEBUGGER__.warn "#{self} is activated."
+        DEBUGGER__.info "#{self} is activated."
       end
     end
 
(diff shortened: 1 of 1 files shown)


ko1:
`b Foo#bar` can be issued when the `Foo#bar` is not defined.
When `Foo#bar` is defined, the breakpoint is enabled.
This message is printed when the breakpoint is enabled.

The purpose of this message is BP is activated correctly.
For example, typo like `b Foo#baz` (typo of `bar`) doesn't show the activation message and you can know the typo.

I think average users *need* this messages.

I agree it should be `info` and I think default loglevel should be `info`.
Advanced users like you can set loglevel as a `warn`.

marianosimone:
Closing in favor of https://github.com/ruby/debug/pull/752

PR #752, opened 2022-09-16
🔊 Make INFO the default logging level

marianosimone:
As proposed by @ko1 in https://github.com/ruby/debug/pull/750, make `INFO` the default level, so we can move down some messages like:

```
DEBUGGER:  BP - Line  /<redacted>.rb:<line> (call) is activated.
```

At the same time, there seem to be a lot of current logs that are more suited for the `DEBUG` level than `INFO`, so I'm moving those down to not make the new default more noisy.

### DIFF
--- a/lib/debug/breakpoint.rb
+++ b/lib/debug/breakpoint.rb
@@ -196,7 +196,7 @@ module DEBUGGER__
       enable
 
       if @pending && !@oneshot
-        DEBUGGER__.warn "#{self} is activated."
+        DEBUGGER__.info "#{self} is activated."
       end
     end
 
@@ -500,7 +500,7 @@ module DEBUGGER__
         retried = false
 
         @tp.enable(target: @method)
-        DEBUGGER__.warn "#{self} is activated." if added
+        DEBUGGER__.info "#{self} is activated." if added
 
         if @sig_op == '#'
           @cond_class = @klass if @method.owner != @klass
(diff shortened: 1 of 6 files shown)


lib/debug/breakpoint.rb
```diff
@@ -8,7 +8,7 @@ class Breakpoint
 
     attr_reader :key
 
-    def initialize cond, command, path, do_enable: true
+    def initialize(cond, command, path, do_enable: true)
```
ko1, line 11: does it related to the main goal?

ko1:
Thank you for the patch.
Viewing the changes, I want additional info command :p

* 1. Convenience information
  * enabling breakpoint 
  * forking process
* 2. Less-convenience information (but sometimes it is useful)
  * loading a ruby file
* 3. debugger Internal events

I agree 1 and 3 is INFO and DEBUG. Not sure for 2. Your patch makes 2 as DEBUG (and I agree almost it is enough).
hmm.

The current design

* WARN - for 1
* INFO - for 2
* DEBUG - for 3

is better? Let me take a time for consideration.
Another tools/debuggers examples are also welecome.

lib/debug/breakpoint.rb
marianosimone: ugh, sorry... This was my IDE doing some autoformatting. I'll revert

marianosimone:
I think `WARN`, `INFO` and `DEBUG` are pretty standard in terms of logging levels, and they roughly match with my patch here.

Feel free to edit this PR as you see fit to match your logging philosophy, though :)

ko1:
fix with #823

PR #754, opened 2022-09-20
Enable "step back <n>" while replay

ono-max:
Currently, `step back <n>` doesn't work in replay mode. This PR fixes it.

### DIFF
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -1061,12 +1061,14 @@ module DEBUGGER__
         else
           leave_subsession [:step, type, arg&.to_i]
         end
-      when /\Aback\z/, /\Areset\z/
+      when /\A(back)\z/, /\A(back)\s+(\d+)\z/, /\A(reset)\z/
         if type != :in
           @ui.puts "only `step #{arg}` is supported."
           :retry
         else
-          request_tc [:step, arg.to_sym]
+          type = $1.to_sym
+          iter = $2&.to_i
+          request_tc [:step, type, iter]
         end
       else
         @ui.puts "Unknown option: #{arg}"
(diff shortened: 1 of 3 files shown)


lib/debug/thread_client.rb
```diff
@@ -1187,8 +1188,11 @@ def enabled?
         @tp_recorder.enabled?
       end
 
-      def step_back
-        @index += 1
+      def step_back iter
+        @index += iter
+        if @index > @log.size
+          @index = @log.size
```
ono-max, line 1194: When n of "step  back <n>" is larger than current `@index`, `@index` will be size of log.

lib/debug/session.rb
```diff
@@ -1056,12 +1056,14 @@ def step_command type, arg
         else
           leave_subsession [:step, type, arg&.to_i]
         end
-      when /\Aback\z/, /\Areset\z/
+      when /\A(back)\z/, /\A(reset)\z/, /\A(back)\s(\d+)\z/
```
ko1, line 1059: \s -> \s+

ko1, line 1059:
```suggestion
      when /\A(back)\z/, /\A(back)\s+(\d+)\z/, /\A(reset)\z/
```

lib/debug/session.rb
```diff
@@ -1056,12 +1056,14 @@ def step_command type, arg
...
         if type != :in
           @ui.puts "only `step #{arg}` is supported."
           :retry
         else
-          request_tc [:step, arg.to_sym]
+          type = $1.to_sym
+          iter = $2.to_i unless $2.nil?
```
ko1, line 1065:
```suggestion
          iter = $2&.to_i
```

PR #755, opened 2022-09-20
Enable "step <n>" while replay

ono-max:
Currently, `step <n>`doesn't work in replay mode. This PR fixes it.

### DIFF
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -841,8 +841,9 @@ module DEBUGGER__
 
           case step_type
           when :in
+            iter = iter || 1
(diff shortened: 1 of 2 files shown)


lib/debug/thread_client.rb
```diff
@@ -1191,8 +1192,11 @@ def step_back
         @index += 1
       end
 
-      def step_forward
-        @index -= 1
+      def step_forward iter
+        @index -= iter
+        if @index < 0
+          @index = 0
```
ono-max, line 1198: When n of "step <n>" is larger thank current index, `@index` will be 0.

PR #758, opened 2022-09-21
🐛 Support displaying variables with non-unicode data

### LINKED ISSUE #756: DAP Server fails when dealing with some binary data
**Your environment**

* `ruby -v`:  2.7.2p137
* `rdbg -v`: 1.6.2

**Describe the bug**
With certain binary data in a variable, the DAP server fails to dump the frame data and crashes

**To Reproduce**

I originally detected this when using protobuf, as the `inspect` method uses interpolates its `value` (which is binary data) into the string that describes it.

A simple way to reproduce is:

1. Create a file like:
```ruby
class PassthroughInspect
  def initialize(data)
    @data = data
  end

  def inspect
    @data
  end
end

with_binary_data = PassthroughInspect.new([8, 200, 1].pack('CCC'))
puts("Put a breakpoint here!")
```
3. Use VSCode, add a breakpoint to the last line, and start debugging
4. Immediately after reaching the breakpoint, the session will end

**Expected behavior**
The debugging session shouldn't crash

**Additional context**
To debug this, you can add something like this to the same file you are debugging:

```
require 'json'

JSON.instance_eval do
  class << self
    alias unsafe_dump dump
  end

  def dump(obj, *args)
    unsafe_dump(obj, *args)
  rescue StandardError => e
    puts("There was an exception dumping the state: #{e}")
    puts("The backtrace is: #{e.backtrace&.join('\n')}")
    puts("The offending object is #{obj}")
  end
end
```

This shows that the problem is `source sequence is illegal/malformed utf-8`, and the trace is:

```
 <redacted>/json-2.3.1/lib/json/common.rb:430:in `generate'
 <redacted>/json-2.3.1/lib/json/common.rb:430:in `generate'
 <redacted>/json-2.3.1/lib/json/common.rb:629:in `dump'
 target.rb:11:in `dump' # this is the test file
 <redacted>/debug-1.6.2/lib/debug/server_dap.rb:206:in `send'
 <redacted>/debug-1.6.2/lib/debug/server_dap.rb:220:in `send_response'
 <redacted>/debug-1.6.2/lib/debug/server_dap.rb:441:in `respond'
 <redacted>/debug-1.6.2/lib/debug/server_dap.rb:650:in `dap_event'
 <redacted>/debug-1.6.2/lib/debug/session.rb:307:in `process_event'
 <redacted>/debug-1.6.2/lib/debug/session.rb:204:in `session_server_main'
 <redacted>/debug-1.6.2/lib/debug/session.rb:174:in `block in activate'
```

The problem is that the `any` variable in my example now has binary data as its value, and when ruby/debug tries to dump it into a String that can send over DAP, JSON complains about it not being a utf-8 string.

marianosimone:
## Description
This fixes #756

DAP is JSON-based, so the state of all variables is dumped into a string that needs to be unicode. When a variable contains binary data that is NOT unicode, this breaks with: `source sequence is illegal/malformed utf-8`.

This patch makes sure that we can display at least part, and show the user that the data they are seeing is not exactly the one that is in memory.

## Testing
I tried this in VSCode (which uses DAP), and saw:

1. Before this patch: the debugging session crashes
2. After this patch:

image

As a side note, `vscode-ruby-debug` doesn't even try to show the variable content and just shows `[Invalid encoding]`, which, IMO, is a worse UX.

### DIFF
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -687,22 +687,12 @@ module DEBUGGER__
   class ThreadClient
     def value_inspect obj
       # TODO: max length should be configuarable?
-      value = DEBUGGER__.safe_inspect obj, short: true, max_length: 4 * 1024
+      str = DEBUGGER__.safe_inspect obj, short: true, max_length: 4 * 1024
 
-      # Given that this is going to be transmitted in a JSON string, it needs to be unicode-encoded
-      # (and "UTF-8" is the default).
-      if value.encoding != Encoding::UTF_8
-        # If the string we got is frozen, we need to make a copy first
-        value = value.dup if value.frozen?
-        value.force_encoding(Encoding::UTF_8)
-      end
-
-      if value.valid_encoding?
-        value
+      if str.encoding == Encoding::UTF_8
+        str.scrub
       else
-        # If a variable contains non-unicode data, at least we can send it partially and signal that
-        # the encoding was unexpected.
-        "[Invalid encoding] #{value.encode("UTF-8", invalid: :replace, undef: :replace)}"
+        str.encode(Encoding::UTF_8, invalid: :replace, undef: :replace)
       end
     end
 
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -687,7 +687,23 @@ module DEBUGGER__
   class ThreadClient
     def value_inspect obj
       # TODO: max length should be configuarable?
-      DEBUGGER__.safe_inspect obj, short: true, max_length: 4 * 1024
+      value = DEBUGGER__.safe_inspect obj, short: true, max_length: 4 * 1024
+
+      # Given that this is going to be transmitted in a JSON string, it needs to be unicode-encoded
+      # (and "UTF-8" is the default).
+      if value.encoding != Encoding::UTF_8
+        # If the string we got is frozen, we need to make a copy first
+        value = value.dup if value.frozen?
+        value.force_encoding(Encoding::UTF_8)
+      end
+
+      if value.valid_encoding?
+        value
+      else
+        # If a variable contains non-unicode data, at least we can send it partially and signal that
+        # the encoding was unexpected.
+        "[Invalid encoding] #{value.encode("UTF-8", invalid: :replace, undef: :replace)}"
+      end
     end
 
     def process_dap args
--- a/test/protocol/binary_data_dap_test.rb
+++ b/test/protocol/binary_data_dap_test.rb
@@ -25,7 +25,7 @@ module DEBUGGER__
         assert_locals_result(
           [
             { name: '%self', value: 'main', type: 'Object' },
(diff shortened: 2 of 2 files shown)


test/protocol/binary_data_dap_test.rb
```diff
@@ -0,0 +1,460 @@
...
+                  name: "%self",
+                  value: "main",
+                  type: "Object",
+                  variablesReference: 8,
+                  indexedVariables: 0,
+                  namedVariables: /\d+/
+                },
+                {
+                  name: "with_binary_data",
+                  value: "[Invalid encoding] \b�\u0001",
```
marianosimone, line 423: This is the interesting part! Without the fix, this would have crashed instead

ko1:
Let me check another debuggers on VSCode (DAP servers)

lib/debug/server_dap.rb
```diff
@@ -687,7 +687,23 @@ def register_vars vars, tid
...
+        value.force_encoding(Encoding::UTF_8)
+      end
+
+      if value.valid_encoding?
+        value
+      else
+        # If a variable contains non-unicode data, at least we can send it partially and signal that
+        # the encoding was unexpected.
+        "[Invalid encoding] #{value.encode("UTF-8", invalid: :replace, undef: :replace)}"
+      end
```
ko1, line 706:
```suggestion
      str = DEBUGGER__.safe_inspect obj, short: true, max_length: 4 * 1024

      if str.encoding == Encoding::UTF_8
        str.scrub
      else
        str.encode(Encoding::UTF_8, invalid: :replace, undef: :replace)
      end
```

is enough. could you check it?

marianosimone: This might solve the UTF-8 problem, but (in order of importance):

1. Doesn't take into account the frozen string problem
2. Doesn't have any comments explaining to future developers *why* we are doing the manipulation
3. I don't believe this is the right thing to do from a UX standpoint: This gives no indication to the user that what they are seeing is *not* the data (which the `[Invalid encoding]` prefix tries to communicate)

ko1: 1. could you give me an example? it doesn't modify the generated string.
2. IMO the code is obvious, and `git blame` is enough.
3. `[Invalid encoding]` is too much because it is ambiguous what is "invalid". JSON doesn't support non-UTF-8 string but the program is valid (the program isn't invalid code). And most of case `inspect` returns valid UTF-8 encoding, I don't think we need care more.

marianosimone: 1. You are right, just double-checked with the test I added for frozen strings, and it works

The other 2 are more of a personal/UX preference, so I defer to you, as the maintainer.

IMO, it's not obvious for someone reading the code without context/without knowing that JSON doesn't support non-unicode strings (the specs don't talk about UTF-8).

I'm accepting the proposed changes, and we can think about better documenting in the future if needed

ko1:
Thank you!

PR #766, opened 2022-09-27
🐛 Fix test generator for protocol tests

marianosimone:
## Description
These generators were wrong, as they were inheriting from `TestCase`, but then trying to use `run_<protocol>_scenario`, which is only available in `ProtocolTestCase`


ono-max:
Thank you 👍

PR #772, opened 2022-10-22
Allow debugger to be attached from Inspector page in Chrome

ono-max:
## Description

https://user-images.githubusercontent.com/59436572/197338015-42f0ece1-86eb-42e4-bb19-74ac8ab17408.mov

When node.js users use Chrome DevTools debugging, they open `chrome://inspect`, then click the link on the page and start debugging it.

https://medium.com/@paul_irish/debugging-node-js-nightlies-with-chrome-devtools-7c4a1b95ae27

This PR supports this feature.

## How to use it

1. Execute debugger as a TCP/IP debuggee. `9229` is the default port on the Inspector page in chrome.

```
$ rdbg target.rb -O --port 9229
```

2. Open `chrome://inspect`, then target will be shown up.
Screen Shot 2022-10-22 at 21 12 33

3. Click "inspect" and the window will be opened!

### DIFF
--- a/lib/debug/server.rb
+++ b/lib/debug/server.rb
@@ -21,6 +21,7 @@ module DEBUGGER__
 
     class Terminate < StandardError; end
     class GreetingError < StandardError; end
+    class RetryConnection < StandardError; end
 
     def deactivate
       @reader_thread.raise Terminate
@@ -77,6 +78,8 @@ module DEBUGGER__
           next
         rescue Terminate
           raise # should catch at outer scope
+        rescue RetryConnection
+          next
         rescue => e
           DEBUGGER__.warn "ReaderThreadError: #{e}"
           pp e.backtrace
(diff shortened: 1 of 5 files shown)


lib/debug/server.rb
```diff
@@ -4,6 +4,8 @@
 require 'etc'
 require_relative 'config'
 require_relative 'version'
+require 'json'
```
ko1, line 7: Please require them in line 165 because they are not used on other type connection.

lib/debug/server.rb
```diff
@@ -157,7 +161,35 @@ def greeting
...
+          url: "file://#{File.absolute_path($0)}",
+          webSocketDebuggerUrl: "ws://#{addr}/#{@uuid}"
+        }]
+        send_http_res body
+        raise RetryConnection
+
+      when /^GET \/#{@uuid} HTTP\/1.1/
         require_relative 'server_cdp'
 
         self.extend(UI_CDP)
```
ko1, line 195:
```suggestion
      when /^GET \/json\/version HTTP\/1.1/, /^GET \/json HTTP\/1.1/,  /^GET \/#{@uuid} HTTP\/1.1/
        require_relative 'server_cdp'
        self.extend(UI_CDP)
        self.cdp_launch g
```

lib/debug/server.rb
```diff
@@ -4,6 +4,7 @@
 require 'etc'
 require_relative 'config'
 require_relative 'version'
+require 'securerandom'
```
ko1, line 7: load only CDP.

PR #792, opened 2022-11-03
`"nonstop": true` option (default: false) on launch.json

### LINKED ISSUE #739: Feature request: Stop debugger until first client connects

ko1:
attach configuration in launch.json file `"nonstop": true` option
is supported. If nonstop is true, do not pause on attach timing.

fix #739

### DIFF
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -273,17 +273,22 @@ module DEBUGGER__
         when 'launch'
           send_response req
           UI_DAP.local_fs_map_set req.dig('arguments', 'localfs') || req.dig('arguments', 'localfsMap')
-          @is_launch = true
+          @nonstop = true
 
         when 'attach'
           send_response req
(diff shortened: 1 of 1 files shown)


ko1:
https://user-images.githubusercontent.com/9558/199809093-536cd1f8-d1b0-4dbd-9dc1-4bc070e0965a.mp4

marianosimone:
I did some testing at Stripe with this, and it seems to work fine. I'll report back if we find any issues

PR #804, opened 2022-11-17
Use binding's location on irb command

k0kubun:
## Description
This makes the debugger's `irb` command work in the same location as the current frame's location, instead of `"(rdbg)/#{f}"`. With the current implementation, commands like `whereami` don't work. The debugger shouldn't modify the source location for `irb` command.

### Before
```rb
$ ruby ~/tmp/a.rb
[1, 3] in ~/tmp/a.rb
     1| require 'debug'
=>   2| debugger
     3| puts 'hello'
=>#0    <main> at ~/tmp/a.rb:2
(rdbg) irb    # command
irb(main)[01:0]> whereami
The current context doesn't have code.
=> nil
```

It doesn't show the current location, and `whereami` doesn't work.

### After
```rb
$ ruby ~/tmp/a.rb
[1, 3] in ~/tmp/a.rb
     1| require 'debug'
=>   2| debugger
     3| puts 'hello'
=>#0    <main> at ~/tmp/a.rb:2
(rdbg) irb    # command

From: /home/k0kubun/tmp/a.rb @ line 2 :

    1: require 'debug'
 => 2: debugger
    3: puts 'hello'

irb(main)[01:0]> whereami

From: /home/k0kubun/tmp/a.rb @ line 2 :

    1: require 'debug'
 => 2: debugger
    3: puts 'hello'

=> nil
```

It shows the current location, and `whereami` works as well.

### DIFF
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -391,15 +391,19 @@ module DEBUGGER__
       end
     end
 
-    def frame_eval_core src, b
+    def frame_eval_core src, b, binding_location: false
       saved_target_frames = @target_frames
       saved_current_frame_index = @current_frame_index
 
       if b
-        f, _l = b.source_location
+        file, lineno = b.source_location
 
         tp_allow_reentry do
-          b.eval(src, "(rdbg)/#{f}")
+          if binding_location
+            b.eval(src, file, lineno)
+          else
+            b.eval(src, "(rdbg)/#{file}")
+          end
         end
       else
         frame_self = current_frame.self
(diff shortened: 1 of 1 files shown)


ko1:
```
From: /home/k0kubun/tmp/a.rb @ line 2 :

    1: require 'debug'
 => 2: debugger
    3: puts 'hello
```

the location lines are needed? (it seems verbose)
if you know how to remove them, please tell me how.

k0kubun:
We could add an option like `binding.irb(show_code: false)` if you wish.

k0kubun:
If you want that behavior and agree with https://github.com/ruby/irb/pull/444, I think we can merge that on IRB's side.

ko1:
Thank you, after releasing with [#444](https://github.com/ruby/irb/pull/444) let's update with it.
Now I merge current patch.

Issue #620, opened 2022-04-10
Allow multiple levels of breakpoint command with options


st0012:
**Your proposal**

Allow nesting commands like this:

```rb
debugger(do: "break Foo#initialize do: (watch @ivar do: pp self)")
```

Or even:

```rb
debugger(do: {
  "break Foo#initialize" => {
    do: { "watch @ivar" => { do: "pp self" } }
  },
  "catch MyError" => { do: "bt 20" }
})
```

This will allow users to write a simple script, place it at one place (e.g. a Rails controller), and execute the entire debugging process without any automatically.

Right now it's not possible because the debugger assumes `do:` or `pre:` options only have 1 level of action.

ko1:
This request is, allowing "break do:" with debug commands, not Ruby expression, right?

st0012:
Sorry, I don't understand the question. I want to allow `break do:` in another `break do:` if that's what you're asking.

Currently this works:

```rb
debugger(do: "break Foo#initialize do: watch @ivar")
```

But this doesn't work because the debugger doesn't know how to parse the later `do:`.

```rb
debugger(do: "break Foo#initialize do: watch @ivar do: pp self")
```

ko1:
Ah, I see. I misunderstood the current `do:` spec.
To introduce parens we need to introduce parser to analyze the commands and we need to consider the syntax more.

PR #828, opened 2022-11-23
Update README.md to show multiple commands execution using binding.b

binarygit:
## Description
Describe your changes:
This addition to the README.md demonstrates how to specify multiple commands to run when using `binding.b`.

I didn't think it obvious that you needed to have two semicolons separating the commands (`cmd 1 ;; cmd 2`). The README does not document this which is why I thought I would add it there.  

I've also added an example to demonstrate how you could use `pre` and `do` arguments together to create  a more advanced/useful workflow.

P.S. I wouldn't have known how to do both of the above things if it hadn't been for @st0012 Ruby Kaigi talk on debugging. So Thank You for that!

### DIFF
--- a/misc/README.md.erb
+++ b/misc/README.md.erb
@@ -568,11 +568,11 @@ It is useful if you only want to call a debug command and don't want to stop the
 ```
 def initialize
   @a = 1
-  binding.b do: 'watch @a'
+  binding.b do: 'info \n watch @a'
 end
 ```
 
-On this case, register a watch breakpoint for `@a` and continue to run.
+On this case, execute the `info` command then register a watch breakpoint for `@a` and continue to run. You can also use `;;` instead of `\n` to separate your commands.
 
 If `pre: 'command'` is specified, the debugger suspends the program and run the `command` as a debug command, and keep suspend.
 It is useful if you have operations before suspend.
--- a/README.md
+++ b/README.md
@@ -832,11 +832,11 @@ It is useful if you only want to call a debug command and don't want to stop the
 ```
 def initialize
   @a = 1
-  binding.b do: 'watch @a'
+  binding.b do: 'info \n watch @a'
 end
 ```
 
-On this case, register a watch breakpoint for `@a` and continue to run.
+On this case, execute the `info` command then register a watch breakpoint for `@a` and continue to run. You can also use `;;` instead of `\n` to separate your commands.
 
 If `pre: 'command'` is specified, the debugger suspends the program and run the `command` as a debug command, and keep suspend.
 It is useful if you have operations before suspend.


misc/README.md.erb
```diff
@@ -586,6 +586,22 @@ end
 
 On this case, you can see the result of `bar()` every time you stop there.
 
+You can use `pre` and `do` together to create more useful workflows.
```
ko1, line 589: I'm not sure why this added block is needed.

ko1:
But it is already documented...

> RUBY_DEBUG_COMMANDS (commands): debug commands invoked at first stop. commands should be separated by ';;'

So I can accept to describe about it (first part of this patch)

misc/README.md.erb
binarygit: It wasn't very obvious to me, how I would achieve something like this. I think the technique used is handy in alot of situations but as I said, it wasn't obvious how I would go about doing it. So, I think it's useful to have this documented here.

I would have never thought that 
```
binding.b do: 'watch @a ;; break foo pre: info'
```
the `pre` inside the string above would be evaluated correctly, like an argument to a `binding.b` and thus suspend the program. I would have never thought this would work, so I think it's worth documenting. 

But If I missed something in the README that does suggest that this is possible and I missed it, please let me know :smile:

binarygit:
> ;; is temporary syntax and I'm not sure we should continue to maintain it.
This is why I didn't documented it.
For the pre:, do: command, I think "\n" (newline) is better.

if the `;;` syntax is temporary, and you recommend using the "\n" to separate the commands then I will change the example to use "\n".  Also a little info about `;;` being temporary might help other people who use it to adopt "\n", wdy think?

misc/README.md.erb
ko1: I'm curious why it is not clear without explanation. `do:` accepts debug command string (described) and `break foo pre: info` is debug command.

on this case, 

```ruby
debugger do: 'watch @a'
debugger do: 'break foo pre: info'

# or
debugger do: <<~DO
  watch @a
  break foo pre: info
DO
```

is more readable for me.
(I usually use `;;` for writability/easy to write, though)

ko1:
> if the ;; syntax is temporary, and you recommend using the "\n" to separate the commands then I will change the example to use "\n". Also a little info about ;; being temporary might help other people who use it to adopt "\n", wdy think?

+1

> I assume we can still write multiple commands in the same line?

Yes, but not sure it is readable.

st0012:
I think this feature has 2 types of usages:

1. Quickly replicate a debugging script (copy the line of script and paste it in other places)
    - For example, I sometimes put multiple `debugger(do: "i ;; bt 10")` as an advanced puts-debugging.
2. Record and share debugging flow with others

For 2), I think newline separator does help:

```rb
debugger do: <<~DO
  watch @a
  break foo pre: info
DO
```

But it's not convenient for 1) as we now need to copy and paste multiple lines instead of just one. Typing `\n` for inline usage could work but it's still more difficult to type than `;;`. If we want to move away from `;;`, I hope it'd be something easier to type than `\n`.

Another thing to consider is that `\n` still don't allow multi-level commands as mentioned in https://github.com/ruby/debug/issues/620, which I think should be included in this discussion as well.

binarygit:
I agree with @st0012 `\n` is more difficult to type than `;;`. If there's a way we can keep `;;`, I would vote for it!

misc/README.md.erb
binarygit: I updated my PR removing the above block and I added a sentence about how users can use `;;` instead of `\n`.
I think this will do, no?
Let me know if you want me to add something else.

Also hope `;;` stays around :smile:

PR #829, opened 2022-11-23
Fix ruby/vscode-rdbg#82

ono-max:
The reason for the failure to set breakpoints is that `@local_fs_map` is nil


ko1:
In general, `local_fs_map` is false because it can be different machine, so this patch is not acceptable.

PR #836, opened 2022-11-25
CDP: support to show global variables

ono-max:


lib/debug/server_cdp.rb
```diff
@@ -716,9 +737,28 @@ def process_protocol_request req
             frame_id = ref[1]
             fid = @frame_map[frame_id]
             request_tc [:cdp, :scope, req, fid]
+          when 'global'
+            vars = global_variables.sort.map do |name|
+              gv = eval(name.to_s)
+              prop = {
```
st0012, line 743: Perhaps we can extract `ThreadClient#propertyDescriptor` and use it here too?

ono-max: Yes, I plan to refactor them after releasing the new version. I added the comment. Thank you.

PR #850, opened 2022-11-30
🐛 Respect breakpoints even when in a skipped path

### LINKED ISSUE #849: "Skip path" configuration makes VSCode stop at breakpoint, but not jump to definition
**To Reproduce**
- Set a breakpoint
- Configure `skip_path` to include the path of the breakpoint
- Debug in VSCode

**Expected behavior**
The breakpoint should be respected, and VSCode should jump to the definition of where the issue is

**Actual behavior**
The debugger stops, but VSCode can't jump to the definition, because the path is skipped (starting in https://github.com/ruby/debug/pull/749)

marianosimone:
## Description
This makes sure that we also check that there's no breakpoint in the path we are about to skip, so VSCode can properly go to the definition.

Not doing so makes the debugger stop, but not go to the definition, which is confusing.

Fixes #849

### DIFF
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -711,7 +711,7 @@ module DEBUGGER__
           next if i < start_frame
 
           path = frame.realpath || frame.path
-          next if skip_path?(path)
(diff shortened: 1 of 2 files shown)


ko1:
I'll apply the prev comment. Thank you!

PR #856, opened 2022-12-02
Support Process.daemon

shunichi:
## Description
Threads could not be started after Process.daemon since the session server thread was gone.

The session must be activated like fork child processes.

### DIFF
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -2421,8 +2421,20 @@ module DEBUGGER__
       end
     end
 
-    private def __fork_setup_for_debugger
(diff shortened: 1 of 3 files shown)


ko1:
Thank you.
One thing we need is what's happens when it is using local debugging...
warning before calling `deaemon`?

shunichi:
@ko1  I'll add the following message before Process.daemon.
Is it OK?

```
DEBUGGER__.warn "Can't debug the code after Process.daemon locally. Use the remote debugging feature."
```

ko1:
Ok! Can you add the condition?

shunichi:
@ko1 I added the condition.

Issue #797, opened 2022-11-09
Feature Request: Push debugger activation to first breakpoint hit instead of `require "debug"`


st0012:
## Proposal

- `require "debug"` won't activate the debugger. Debugger will only be activated when the first breakpoint is triggered.
    - `require "debug"` will be the same as `require "debug/session"`
- `require "debug/open"` and `require "debug/open_nonstop"`'s behaviour will remain the same.

### Side-effects

In apps that have `Bundler.require`, like most Rails apps, having `gem "debug"` in Gemfile means the `debug` gem is always activated when the app is booted. 

However, with the debugger's activation, many things will happen in the background:

- Additional `TracePoint`s would be enabled ([example](https://github.com/ruby/debug/blob/a402e7304db8b78e8b7ee1e2142e47214b99cc72/lib/debug/session.rb#L141))
- [A new thread](https://github.com/ruby/debug/blob/a402e7304db8b78e8b7ee1e2142e47214b99cc72/lib/debug/session.rb#L172) will be spawned
- Additional computation and object allocation that come with the above

To most apps, having them in the background doesn't make a huge difference. But to some, having the debugger activated causes tests to fail. 

For example, I tested 5 mid-to-large Rails apps in Shopify, and 3 of them have test failures that only happen when `debug` is activated:

- Some tests mock `File.readline`, it could break when [this line](https://github.com/ruby/debug/blob/master/lib/debug/source_repository.rb#L98) is triggered in the background.
- There's also weird `NoMethodError` exceptions that only happens with the debugger activated.

Therefore, we always have `require: false` after `gem "debug"`.

### Problem Solved?

However, having `require: false` means users need to manually type `require "debug"` before start debugging with console. But in the meantime, `byebug` can be required by default without the same side-effects. So `byebug` doesn't need that manual require.

This extra require step is quite inconvenient for people considering switching from `byebug` to `debug`.

### Other Smaller Issues

- The activation and forking messages are noisy and disabling them project-by-project is not trivial.
- `require "debug"` locks the UI to console, so if users want to use the remote UI by requiring `debug/open` after the app is booted, `require: false` is also needed.

### Possible Change

```diff
diff --git a/lib/debug.rb b/lib/debug.rb
index 15ebccb..3489268 100644
--- a/lib/debug.rb
+++ b/lib/debug.rb
@@ -1,5 +1,3 @@
 # frozen_string_literal: true

 require_relative 'debug/session'
-return unless defined?(DEBUGGER__)
[...]

PR #860, opened 2022-12-02
`RUBY_DEBUG_LAZY` boot option

ko1:
With `RUBY_DEBUG_LAZY=1`, `require 'debug'` doesn't start a session
but start when `debugger` method is called (or `require 'debug/start'`
is called).

This feature is experimental.
https://github.com/ruby/debug/issues/797

### DIFF
--- a/lib/debug.rb
+++ b/lib/debug.rb
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
-require_relative 'debug/session'
-return unless defined?(DEBUGGER__)
-DEBUGGER__::start no_sigint_hook: true, nonstop: true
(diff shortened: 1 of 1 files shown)


st0012:
Thank you for this PR, I'll test it today.

st0012:
I've tested it and the result looks promising 🎉  I'll make it the default in one of our major apps next week and see if any dev report issues.

FWIW, `v1.7.0` itself is already a big improvement. 2 apps that used to break CI with debugger activated now works fine WITHOUT applying this feature 👍 Thanks for the work 🙏 

A few other apps are still failing though (with reduced number of errors), so we still need this lazy-activation for the near future.

PR #869, opened 2022-12-16
DAP: show `#dump` when pp'ed string is too long

ko1:
For String objects, show `#dump` when the String object is long
enough. However, the length of a String is not correct condition
because elimination with `...` is used for long `pp`ed result.

This patch shows `#dump` field if `pp'ed_str.end_with?('...')`
is true.


ko1:
image

PR #879, opened 2022-12-27
Add missing close backquote

makenowjust:
## Description

Some close backquotes for the `until` command are missing, so the display of `README.md` at GitHub is broken. This PR fixes it.

Thank you.


hsbt:
👍

PR #880, opened 2023-01-02
🔼 Update CI matrix Ruby versions

yamashush:
## Description

Update CI matrix ruby versions.

- Add 3.2 ✨ 
- ~~Remove 2.6~~ 🔥 

ref: https://www.ruby-lang.org/en/downloads/branches/

```
Ruby 3.2
status: normal maintenance
release date: 2022-12-25
```

```
Ruby 2.6
status: eol
release date: 2018-12-25
EOL date: 2022-04-12
```

### DIFF
--- a/.github/workflows/ruby-macos.yaml
+++ b/.github/workflows/ruby-macos.yaml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby-version: ['3.1', 'head']
+        ruby-version: ['3.2', 'head']
 
     steps:
     - uses: actions/checkout@v3
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby-version: ['2.7', '3.0', '3.1', '3.2', 'head', 'debug']
+        ruby-version: ['2.6, '2.7', '3.0', '3.1', '3.2', 'head', 'debug']
 
     steps:
     - uses: actions/checkout@v3
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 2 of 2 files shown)


st0012:
I think 2.6 shouldn't be removed because it's still [supported by this gem](https://github.com/ruby/debug/blob/master/debug.gemspec#L13). EOL of a version only means the version itself won't receive further updates (usually), but it doesn't mean gems have dropped support for it.

gwillcox-r7:
@st0012 My understanding is that Ruby 2.6 is no longer receiving any updates anymore and Ruby 2.7 is soon to enter EOL state and enter a similar condition. My understanding is Ruby goes from stable releases, to security releases and updates only, and then eventually to EOL where they stop releasing any updates at all.

If I'm incorrect in that though feel free to correct me but https://www.ruby-lang.org/en/downloads/ states that Ruby 2.6.10 is not maintained anymore and 2.7.x is soon to enter the same state so that is the info I'm going off of here.

st0012:
@gwillcox-r7 As I've said, a version of Ruby stop receiving updates != gems will drop support on it right away. Given that Ruby 2.6 is still supported by this particular gem, it's better to make sure it's tested on CI as well.

yamashush:
Sorry for late reply, and I've understood that it is necessary to leave Ruby 2.6.
Thank you :) :) :)

Is it ok to add Ruby 3.2?

st0012:
Of course, other libs like `IRB` or `reline` also has 3.2 in their test matrixes now 🙂

st0012:
cc @ko1 I think this could help distinguish if failures are caused by Ruby 3.2 and/or 3.3.

ko1:
Thank you for the patch and discussion.

* I don't have any opinion about Ruby 2.6, and I agree dropping Ruby 2.6 makes things easy.
* But just now we have no trouble on Ruby 2.6. So I keep Ruby 2.6 matrix.

If you find any trouble on Ruby 2.6 on CI, please tell me and drop it.

ko1:
(in other words) we shouldn't hesitate to introduce 2.6 incompatible code on debug.gem just now.

PR #884, opened 2023-01-10
Add test for global variable support in DAP and CDP

vinistock:
## Description

PR #835 added support for global variables. This PR just adds a test case to verify that request.

To avoid having to assert every single global variable using an array (which is a big number considering Ruby's built in globals), I extracted part of `assert_locals_result` into `gather_variables`, so that we can get all available globals in the test and just verify a few of them.


vinistock:
Closing in favour of #885.

PR #885, opened 2023-01-11
Add tests for various variable features

egiurleo:
This PR adds tests for the following PRs:
- #835
- #806 

They both rely on @vinistock's addition of the `gather_variables` method extract the behavior of collecting variables.

This method is explained in a comment from #884:
> To avoid having to assert every single global variable using an array (which is a big number considering Ruby's built in globals), I extracted part of assert_locals_result into gather_variables, so that we can get all available globals in the test and just verify a few of them.

### DIFF
--- a/test/protocol/variables_test.rb
+++ b/test/protocol/variables_test.rb
@@ -3,7 +3,7 @@
 require_relative "../support/protocol_test_case"
 
 module DEBUGGER__
-  class DAPVariablesTest < ProtocolTestCase
+  class DAPGlobalVariablesTest < ProtocolTestCase
     PROGRAM = <<~RUBY
       1| $a = 1
       2| $b = 2
@@ -18,19 +18,19 @@ module DEBUGGER__
         globals = gather_variables(type: "globals")
 
         # User defined globals
-        assert_includes(globals, { name: "$a", value: "1", type: "Integer" })
-        assert_includes(globals, { name: "$b", value: "2", type: "Integer" })
+        assert_includes(globals, { name: "$a", value: "1", type: "Integer", variablesReference: 0 })
+        assert_includes(globals, { name: "$b", value: "2", type: "Integer", variablesReference: 0 })
 
         # Ruby defined globals
-        assert_includes(globals, { name: "$VERBOSE", value: "false", type: "FalseClass" })
-        assert_includes(globals, { name: "$stdout", value: "#<IO:<STDOUT>>", type: "IO" })
+        assert_includes(globals, { name: "$VERBOSE", value: "false", type: "FalseClass", variablesReference: 0 })
+        assert_includes(globals, { name: "$stdout", value: "#<IO:<STDOUT>>", type: "IO", variablesReference: 0 })
 
         req_terminate_debuggee
       end
     end
   end
 
-  class CDPVariablesTest < ProtocolTestCase
+  class CDPGlobalVariablesTest < ProtocolTestCase
     PROGRAM = <<~RUBY
       1| $a = 1
       2| $b = 2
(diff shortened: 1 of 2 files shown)


test/protocol/variables_test.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+require_relative "../support/protocol_test_case"
+
+module DEBUGGER__
+  class DAPVariablesTest < ProtocolTestCase
```
st0012, line 6: Nitpick: since the program script for this class only has global variables, could we be more specific on the class name, like `DAPGlobalVariablesTest`?

test/protocol/variables_test.rb
```diff
@@ -0,0 +1,85 @@
...
+        # Ruby defined globals
+        assert_includes(globals, { name: "$VERBOSE", value: "false", type: "FalseClass" })
+        assert_includes(globals, { name: "$stdout", value: "#<IO:<STDOUT>>", type: "IO" })
+
+        req_terminate_debuggee
+      end
+    end
+  end
+
+  class CDPVariablesTest < ProtocolTestCase
```
st0012, line 33: Same as the above.

test/protocol/variables_test.rb
```diff
@@ -0,0 +1,85 @@
...
+      end
+    end
+  end
+
+  class DAPInstanceVariableTest < ProtocolTestCase
+    PROGRAM = <<~RUBY
+      1| @a = 1
+      2| @c = 3
+      3| @b = 2
+      4| @z = 4
```
st0012, line 65: Also a nitpick: I know it's because the breakpoint would stop the beginning of the line so `@z` wouldn't show up in the response. But at the first glance it's not very obvious. Maybe we can make it something else like `__LINE__` to avoid the confusion?

st0012 (commented):
Some small suggestions

st0012 (commented):
LGTM 👍

ko1:
@egiurleo Sorry for late response.

Thank you for the tests.
I can not run github actions (tests) on this PR.
I'm not sure why but could you rebase the PR and I'll check it later we can run GH actions (tests).

egiurleo:
@ko1 done!

ko1:
Tests seem failed 😢 

```
Error: test_binary_data_gets_encoded(DEBUGGER__::BinaryDataDAPTest): NoMethodError: undefined method `gather_dap_variables' for an instance of DEBUGGER__::BinaryDataDAPTest
/home/runner/work/debug/debug/test/support/protocol_test_case.rb:246:in `assert_locals_result'
/home/runner/work/debug/debug/test/protocol/binary_data_dap_test.rb:25:in `block in test_binary_data_gets_encoded'
     22:       run_protocol_scenario(program, cdp: false) do
     23:         req_add_breakpoint 12
     24:         req_continue
  => 25:         assert_locals_result(
     [26](https://github.com/ruby/debug/actions/runs/4355848307/jobs/7615161572#step:4:27):           [
     [27](https://github.com/ruby/debug/actions/runs/4355848307/jobs/7615161572#step:4:28):             { name: '%self', value: 'main', type: 'Object' },
     [28](https://github.com/ruby/debug/actions/runs/4355848307/jobs/7615161572#step:4:29):             { name: 'with_binary_data', value: [8, 200, 1].pack('CCC').encode(Encoding::UTF_8, invalid: :replace, undef: :replace), type: 'PassthroughInspect' }```

andyw8:
I've pushed a fix for this, however one check is failing because of the shorter timeout on `ruby-macos`:

> The job running on runner GitHub Actions 23 has exceeded the maximum execution time of 15 minutes

ko1:
Thank you!

PR #890, opened 2023-01-31
Always propagate keyword arguments in WebSocketServer

Maumagnaguagno:
Methods send_response, send_event and fire_event are checking if the keyword arguments are empty before propagating an empty hash, the same can be accomplished just propagating the original keyword arguments.

### DIFF
--- a/lib/debug/server_cdp.rb
+++ b/lib/debug/server_cdp.rb
@@ -450,11 +450,7 @@ module DEBUGGER__
     end
 
     def send_response req, **res
-      if res.empty?
-        @ws_server.send id: req['id'], result: {}
-      else
-        @ws_server.send id: req['id'], result: res
-      end
+      @ws_server.send id: req['id'], result: res
     end
 
     def send_fail_response req, **res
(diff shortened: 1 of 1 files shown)


Maumagnaguagno:
Can anyone enable the tests for this PR?
I also would like to know if we should alias/remove ``fire_event``, as it has the same behavior of ``send_event``.

ko1:
> I also would like to know if we should alias/remove `fire_event`, as it has the same behavior of `send_event`.

@ono-max san could you answer the question?

ono-max:
>I also would like to know if we should alias/remove `fire_event`, as it has the same behavior of `send_event`.

I defined `fire_event` method to distinguish methods called from Session.

https://github.com/ruby/debug/blob/master/lib/debug/server_cdp.rb#L675

Maumagnaguagno:
That means we could alias the method, let me know if you want a PR for that :+1:.

ono-max:
@Maumagnaguagno 

Thanks! I would appreciate it if you could do it.

PR #892, opened 2023-02-06
Remove redundant raw detach test

st0012:
The test's subject is the `disconnect` request with `restart: false,
terminateDebuggee: <true|false>` arguments, which has been covered by the
tests in [`disconnect_dap_test.rb`](https://github.com/ruby/debug/blob/master/test/protocol/disconnect_dap_test.rb). So this test case has become obsolete.

### Background

I was investigating the [master build failure](https://github.com/ruby/debug/actions/runs/3790505882/jobs/6445178180) with @andyw8. And we later found that this test case isn't designed to catch that error and actually is redundant. So by removing it we can also fix part of the master build failure.

If having a test to capture the stacktrace difference of the test failure is preferred, I'll be happy to open another PR for it. But IMO, it'd be testing Ruby's behaviour instead of testing this gem.


ko1:
Thank you for investigation.

PR #909, opened 2023-02-27
DAP: use Object#__send__ to avoid name conflicts

ono-max:
Since Object#send is used https://github.com/ruby/debug/blob/master/lib/debug/server_dap.rb#L456, we need to rename to avoid name conflicts.


Maumagnaguagno:
We could use ``Object#__send__`` instead.

ono-max:
Thank you, I'll change it.

PR #910, opened 2023-02-27
DAP: add tracer for vscode-rdbg

ono-max:
This PR supports a new feature "trace inspector" in vscode-rdbg. 

WIP

### DIFF
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -463,6 +463,10 @@ module DEBUGGER__
       send_event :terminated unless @sock.closed?
     end
 
+    def request_rdbgTraceInspector req
+      @q_msg << req
+    end
+
     ## called by the SESSION thread
 
     def respond req, res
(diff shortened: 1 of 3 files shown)


lib/debug/server_dap.rb
```diff
@@ -506,17 +511,28 @@ def event type, *args
...
     end
+
+    def notify_updated tid, ts
+      ts.values.each{|t|
+        if t.type == 'dap'
+          send_event 'rdbgTraceInspector', command: 'updated'
+          break
+        end
+      }
+    end
```
ko1, line 535:
```suggestion
    def notify_updated
          send_event 'rdbgTraceInspector', command: 'updated' if @use_trace_inspector
        end
      }
    end
```

lib/debug/server_dap.rb
```diff
@@ -463,6 +463,10 @@ def process
       send_event :terminated unless @sock.closed?
     end
 
+    def request_rdbgTraceInspector req
+      @q_msg << req
+    end
```
ko1, line 468: @use_trace_inspector = true

PR #913, opened 2023-03-04
DAP: allow custom request extension in ThreadClient class

ono-max:
This PR is for https://github.com/ruby/debug/pull/935.


ko1:
could you test this code or left it until we surely need it?

PR #914, opened 2023-03-06
DAP: use Mutex when sending response

ono-max:
When debug.gem tries to send response from multiple threads, the socket connection is closed. I confirmed this bug when using custom request from vscode-rdbg.

Also, because Errno::EPIPE is rescued while sending message to socket, protocol_test_case_test.rb did not pass. protocol_test_case_test.rb had been passed because ReaderThreadError was occurred and the debuggee process was still alive. Here is a scenario. After closing socket, terminated event was sent. However socket was closed, so debuggee process raised Errno::EPIPE and debugggee process was still alive. The test framework detected the status and failed.

Thus I fixed so that the test framework does not kill the debuggee process unexpectedly.

### DIFF
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -125,6 +125,7 @@ module DEBUGGER__
     def dap_setup bytes
       CONFIG.set_config no_color: true
       @seq = 0
+      @send_lock = Mutex.new
 
       case self
       when UI_UnixDomainServer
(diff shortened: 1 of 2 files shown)


lib/debug/server_dap.rb
```diff
@@ -212,7 +213,11 @@ def send **kw
       if sock = @sock
         kw[:seq] = @seq += 1
         str = JSON.dump(kw)
-        sock.write "Content-Length: #{str.bytesize}\r\n\r\n#{str}"
+        @mutex.synchronize do
```
ko1, line 216: please rescue Close exception instead of checking the `close?`

lib/debug/server_dap.rb
```diff
@@ -125,6 +125,7 @@ def self.local_fs_map_set map
     def dap_setup bytes
       CONFIG.set_config no_color: true
       @seq = 0
+      @mutex = Mutex.new
```
ko1, line 128: please use `@send_lock` instead of `@mutex` to make clear the purpose.

PR #915, opened 2023-03-06
Fix bug that "trace line" does not work after stopping trace once

ono-max:


st0012 (commented):
Good catch on this 👍 
I think the method could be simplified as:

```rb
    def add_tracer tracer
      if @tracers[tracer.key]&.enabled?
        tracer.disable
        @ui.puts "Duplicated tracer: #{tracer}"
      else
        @tracers[tracer.key] = tracer
        @ui.puts "Enable #{tracer}"
      end
    end
```

ono-max:
Thank you for your suggestion. I've updated it.

PR #916, opened 2023-03-06
DAP: introduce Rdbg Trace Inspector

ono-max:
We'll introduce a new feature Rdbg Trace Inspector. This is useful when tracing in VS Code.

https://user-images.githubusercontent.com/59436572/223106427-2b808adb-b38e-4c44-9050-dbb8cd0cc893.mov

# How to use it

1. Click on "Start Trace" in TRACE view
2. Execute the program

That's it! You can check the location and returned value by clicking the trace log.

## How to choose events

On the default, line, call, and return events are selected.

If you want to change the default, click "more items", then choose events in the context menu.

https://user-images.githubusercontent.com/59436572/223157318-70889b66-72fd-4bb1-b206-e20e415222a0.mov

### DIFF
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -9,6 +9,9 @@ module DEBUGGER__
   module UI_DAP
     SHOW_PROTOCOL = ENV['DEBUG_DAP_SHOW_PROTOCOL'] == '1' || ENV['RUBY_DEBUG_DAP_SHOW_PROTOCOL'] == '1'
 
+    # After loading UI_DAP module, load the custom DAP file.
+    require_relative 'dap_custom'
+
     def self.setup debug_port
       if File.directory? '.vscode'
         dir = Dir.pwd
--- a/lib/debug/tracer.rb
+++ b/lib/debug/tracer.rb
@@ -147,6 +147,78 @@ module DEBUGGER__
     end
   end
 
+  # MultiTracer is designated for DAP.
+  class MultiTracer < Tracer
+    MAX_LOG_SIZE = 4000
+    def initialize ui, evts, **kw
+      @evts = evts
+      @log = []
+      super(ui, **kw)
+    end
+
+    attr_reader :log
+
+    def setup
+      @tracer = TracePoint.new(*@evts){|tp|
+        next if skip?(tp)
+
+        case tp.event
+        when :call, :c_call, :b_call
+          append(call_trace_log(tp))
+        when :return, :c_return, :b_return
+          return_str = DEBUGGER__.safe_inspect(tp.return_value, short: true, max_length: 120)
+          append(call_trace_log(tp, return_str: return_str))
+        when :line
+          append(line_trace_log(tp))
+        end
+      }
+    end
+
+    def call_identifier_str tp
+      if tp.defined_class
+        minfo(tp)
+      else
+        "block"
+      end
+    end
+
+    def append log
+      if @log.size >= MAX_LOG_SIZE
+        @log.shift
+      end
+      @log << log
+    end
+
+    def call_trace_log tp, return_str: nil
+      log = {
+        depth: DEBUGGER__.frame_depth,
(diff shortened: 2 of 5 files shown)


st0012:
I've been wanting this feature for a while, this looks super cool 🎉 
How can we select the even type with this UI (e.g. exclude the line events)? From the video I don't see an option to do so.
Also, is it possible to have some protocol tests for it?

ono-max:
> how can we select the even type with this UI (e.g. exclude the line events)?

We can select the option as follows:

https://user-images.githubusercontent.com/59436572/223157318-70889b66-72fd-4bb1-b206-e20e415222a0.mov

ono-max:
> Also, is it possible to have some protocol tests for it?

Yeah, I think that we can add tests.

lib/debug/tracer.rb
```diff
@@ -148,6 +148,78 @@ def skip_with_pattern?(tp)
     end
   end
 
+  class MultiTracer < Tracer
+    MAX_LENGTH = 4000
```
ko1, line 152: `LENGTH` -> `RECORDS`

ko1: or `LOGS` if you u se `@log`

lib/debug/tracer.rb
```diff
@@ -148,6 +148,78 @@ def skip_with_pattern?(tp)
     end
   end
 
+  class MultiTracer < Tracer
```
ko1, line 151: `# Now it is only for DAP`

lib/debug/tracer.rb
```diff
@@ -148,6 +148,78 @@ def skip_with_pattern?(tp)
...
+      @evts = evts
+    end
+
+    attr_reader :log
+
+    def setup
+      @tracer = TracePoint.new(*@evts){|tp|
+        next if skip?(tp)
+
+        call_identifier_str =
```
ko1, line 165: cut off to the method

lib/debug/server_dap.rb
```diff
@@ -655,6 +659,49 @@ def process_protocol_request req
...
+          when 'line'
+            evts << :line
+          when 'call'
+            evts << :call
+            evts << :c_call
+            evts << :b_call
+          when 'return'
+            evts << :return
+            evts << :c_return
+            evts << :b_return
```
ko1, line 679: else
  raise "unknown trace type #{evt}"

lib/debug/server_dap.rb
```diff
@@ -655,6 +659,49 @@ def process_protocol_request req
       end
     end
 
+    def request_rdbgTraceInspector(req)
+      cmd = req.dig('arguments', 'command')
+      case cmd
+      when 'enable'
+        events = req.dig('arguments', 'events')
+        evts = []
+        events.each{|evt|
```
ko1, line 668: Another idea:

```ruby
EVENTS = {'line' => [:line], ...}

def ...
  evts = []
  events.each{|evt| evts.concat EVENTS[evt]}
  ...
```

or 

```ruby
  evts = events.flat_map{|evt| EVENTS[evt]}
```

But the current code makes sense because it is easy to read.

lib/debug/server_dap.rb
```diff
@@ -655,6 +659,49 @@ def process_protocol_request req
...
+        @ui.respond req, {}
+      when 'disable'
+        @tracers.values.each{|t|
+          if t.type == 'multi'
+            t.disable
+            break
+          end
+        }
+        @ui.respond req, {}
+      when 'logs'
```
ko1, line 692: other subcommands are verbs so I recommend to use `correct` (a verb word)

st0012:
> Yeah, I think that we can add tests.

@ono-max Thanks. Could we first have some tests in the same PR?

ono-max:
I'll try it, but I'm not sure if I can add tests in the same PR because I'm busy. Sorry.

PR #919, opened 2023-03-07
Alias Session send methods in WebSocketServer

Maumagnaguagno:
Methods ``respond``, ``respond_fail`` and ``fire_event`` can be aliased to ``send_response``, ``send_fail_response`` and ``send_event``, respectively.
Related to #890.

### DIFF
--- a/lib/debug/server_cdp.rb
+++ b/lib/debug/server_cdp.rb
@@ -668,17 +668,9 @@ module DEBUGGER__
 
     ## Called by the SESSION thread
 
-    def respond req, **result
-      send_response req, **result
-    end
-
-    def respond_fail req, **result
-      send_fail_response req, **result
-    end
-
-    def fire_event event, **result
-      send_event event, **result
-    end
+    alias respond send_response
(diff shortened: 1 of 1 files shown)


lib/debug/server_dap.rb
```diff
@@ -465,9 +465,7 @@ def process
 
     ## called by the SESSION thread
 
-    def respond req, res
-      send_response(req, **res)
-    end
+    alias respond send_response
```
ono-max, line 468: I cannot accept this change because this change breaks existing codes.
```ruby
def send_response(req, **res)
  puts req, res
end

# def respond req, res
#   send_response(req, **res)
# end

alias respond send_response

def send_response(req, **res)
  puts req, res
end

respond({}, {a: 1}) # => wrong number of arguments (given 2, expected 1) (ArgumentError)
```

ono-max: No problem. Thank you for your contribution 👍

ono-max (approved):

Issue #900, opened 2023-02-17
First breakpoint is skipped when running via `bin/rails runner` on a remote development environment


andyw8:
**Describe the bug**

First breakpoint is skipped when running via `bin/rails runner` on a remote development environment

**To Reproduce**

1. Go to https://github.com/github/codespaces-rails to start with a new Rails app.
2. Choose Use This Template -> Open in a Codespace
3. Create the following in `.vscode/launch.json`
    ```json
       {
        // Use IntelliSense to learn about possible attributes.
        // Hover to view descriptions of existing attributes.
        // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
        "version": "0.2.0",
        "configurations": [
            {
                "type": "rdbg",
                "name": "Run the current file with rails runner and rdbg",
                "useBundler": true,
                "request": "launch",
                "showProtocolLog": true,
                "command": "${workspaceRoot}/bin/rails",
                "script": "runner ${file}",
                "args": [],
                "askParameters": false,
                "env": {
                  "DISABLE_SPRING": "1"
                }
              },
              {
                "type": "rdbg",
                "name": "Run the current file with rdbg",
                "useBundler": true,
                "request": "launch",
                "showProtocolLog": true,
                "command": "ruby",
                "script": "${file}",
                "args": [],
                "askParameters": false,
                "env": {
                  "DISABLE_SPRING": "1"
                }
              },
        ]
    }
    ```

4. Create the following in `test.rb`.

    ```rb
    a=1
    binding.b
    binding.b
    ```
5. Ensure v0.1.0 of the vscode-rdbg extension is installed.
6. Switch to `test.rb` and choose "Run the current file with rdbg". Observe that it correctly stops at the first breakpoint.
7. Stop the debugger.
8. Choose "Run the current file with rails runner and rdbg"
9. Observe that it incorrectly skips the first breakpoint and stops at the second.

**Your environment**

* `ruby -v`: ruby 3.1.3p185 (2022-11-24 revision 1a6b16756e) [x86_64-linux]
* `rdbg -v`: rdbg 1.7.1

**Expected behavior**

It should stop at the first breakpoint.

**Additional context**

It does not happen locally, only when running on a remote environment such as Codespaces.

andyw8:
Some further findings on this:

It seems like a race condition. If I add `sleep 0.1` before this line:
[...]

PR #920, opened 2023-03-08
Improve dap command execution

st0012:
Fixes https://github.com/ruby/debug/issues/900

## Description

From investigating #900 with @andyw8, I found that the current implementation has 2 potential problems:

1. The command sent via `evaluate` request (e.g. `,b 5`) is actually executed AFTER the next session command (e.g. [enqueued request hashes](https://github.com/ruby/debug/blob/master/lib/debug/server_dap.rb#L447-L452)). This is because:
    1. `wait_command` would usually be waiting for [`@ui.readline`](https://github.com/ruby/debug/blob/733dd4b464191e147f00fe767f3c6dcd6ae2c6b9/lib/debug/session.rb#L399) when the request arrives. Therefore, the `@preset_command` added through `debugger do:` won't be picked up right away.
    1. Now, when the later request (e.g. `stackTrace`) comes in, it'll be picked up by `@ui.readline` and be executed right away.
    1. Only after the current `wait_command` call finishes and we enters the next one, the `evaluate` command would be picked up.
1. The `debugger do:`'s behaviour is not deterministic, at least in the `evaluate` request's case. Sometimes the `do` command would be added through [`SESSION.add_preset_commands`](https://github.com/ruby/debug/blob/733dd4b464191e147f00fe767f3c6dcd6ae2c6b9/lib/debug/session.rb#L2581) and sometimes through [`DEBUGGER__.add_line_breakpoint`](https://github.com/ruby/debug/blob/733dd4b464191e147f00fe767f3c6dcd6ae2c6b9/lib/debug/session.rb#L2584). We believe this is one of the causes of #900.

## Implementation

To address these 2 problems I think we can directly enqueue the sent command to `@q_msg`:

- It solves problem 1 because the command would be picked up by `@ui.readline` too so the execution order would be correct. The test added in the PR covers this case.
- It also avoids problem 2 because the behaviour won't be affected by whether we're in a subsession or not.

## Questions and Concerns

- I assume there was a reason that this feature was implemented with `debugger do:` in the first place, but we couldn't figure it out ourselves. @ko1 would you mind elaborating it?
- Is it safe for `wait_command` to take the command from both synchronous state change (`@preset_command`) and asynchronously through `@ui.readline`?
- I think problem 1 could also be addressed by improving `wait_command`. But I don't know the design decisions behind the current implementation + I think the impact radius is too big. So I didn't pursue it.


lib/debug/server_dap.rb
```diff
@@ -437,10 +437,12 @@ def process
           expr = req.dig('arguments', 'expression')
           if /\A\s*,(.+)\z/ =~ expr
             dbg_expr = $1
+            commands = dbg_expr.split(';;')
+            commands.each { |cmd| @q_msg << cmd }
+
```
Maumagnaguagno, line 442: String#split accepts a block since [Ruby 2.6](https://bugs.ruby-lang.org/issues/4780).
``dbg_expr.split(';;') { |cmd| @q_msg << cmd }``

st0012: TIL this usage! Thank you 😄

ko1:
All right, maybe I consider something but I couldn't remember the reason why I haven't choose direct manipulation of `@q_msg`.
Let's try.

PR #925, opened 2023-03-09
Improve test process' kill signalling

st0012:
1. `remote_info.failed_process` stores symbol but we only use the value's presence to check if the process is force-killed.
2. The force-killed status is directly controlled by `kill_safely` through `kill_remote_debuggee`, which is directly called right before we check the status with `remote_info.failed_process`.

Combining the two, we can just let `kill_safely` and `kill_remote_debuggee` return the force-killed status and not storing it in `remote_info`, which already contains a bunch of information.

This also eliminates the need to pass `test_info` to `kill_safely`, which makes related code easier to understand and maintain.

### DIFF
--- a/test/support/test_case.rb
+++ b/test/support/test_case.rb
@@ -14,7 +14,7 @@ require_relative 'assertions'
 module DEBUGGER__
   class TestCase < Test::Unit::TestCase
     TestInfo = Struct.new(:queue, :mode, :prompt_pattern, :remote_info,
-                          :backlog, :last_backlog, :internal_info, :failed_process)
+                          :backlog, :last_backlog, :internal_info)
 
     RemoteInfo = Struct.new(:r, :w, :pid, :sock_path, :port, :reader_thread, :debuggee_backlog)
 
--- a/test/support/test_case.rb
+++ b/test/support/test_case.rb
@@ -122,17 +122,17 @@ module DEBUGGER__
       true
     end
 
-    def kill_safely pid, name, test_info
-      return if wait_pid pid, TIMEOUT_SEC
-
-      test_info.failed_process = name
+    def kill_safely pid
+      return false if wait_pid pid, TIMEOUT_SEC
 
       Process.kill :TERM, pid
-      return if wait_pid pid, 0.2
+      return true if wait_pid pid, 0.2
 
       Process.kill :KILL, pid
       Process.waitpid(pid)
+      true
     rescue Errno::EPERM, Errno::ESRCH
+      true
     end
 
     def check_error(error, test_info)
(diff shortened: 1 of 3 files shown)


ono-max:
> remote_info.failed_process stores symbol but we only use the value's presence to check if the process is force-killed.

The reason for storing the symbol is to use in console tests such as https://github.com/ruby/debug/blob/c3bfcc29377ac13367370c26efece0c109e8724e/test/support/console_test_case.rb#L210. (Sorry, I think that I accidentally deleted this part before)

But it's ok to check each killing method(kill_remote_debuggee and kill_safely). Do you think that it'll be better?

st0012:
> But it's ok to check each killing method(kill_remote_debuggee and kill_safely). Do you think that it'll be better?

Yeah definitely. Since we always check the state right after calling kill methods, using return value will be more readable and make refactoring easier. 
However, I think in those cases we can also write fixed error messages based on conditions?

ono-max:
> However, I think in those cases we can also write fixed error messages based on conditions?

Yes, that's what I said here:
"But it's ok to check each killing method"

We can write fixed error messages by checking the state in all killing methods.

ono-max:
> Yeah definitely. Since we always check the state right after calling kill methods, using return value will be more readable and make refactoring easier.

All right, Let's go with that idea.

ono-max:
I'll review the changes tomorrow. Thank you for the explanation.

test/support/test_case.rb
```diff
@@ -142,13 +144,15 @@ def check_error(error, test_info)
     end
 
     def kill_remote_debuggee test_info
-      return unless r = test_info.remote_info
+      needs_kill = false
+      return needs_kill unless r = test_info.remote_info
```
ono-max, line 148:
```suggestion
      return false unless r = test_info.remote_info
```

test/support/test_case.rb
```diff
@@ -122,16 +122,18 @@ def wait_pid pid, sec
       true
     end
 
-    def kill_safely pid, name, test_info
-      return if wait_pid pid, TIMEOUT_SEC
+    def kill_safely pid
+      needs_kill = false
+      return needs_kill if wait_pid pid, TIMEOUT_SEC
```
ono-max, line 127:
```suggestion
      return false if wait_pid pid, TIMEOUT_SEC
```

ono-max:
How about renaming `needs_kill` to `signal_sent`?

st0012:
> How about renaming needs_kill to signal_sent?

Yeah `needs_kill` is not the best name. But since we also send signal in both cases, I feel `signal_sent` isn't clear either.
I feel `force_killed` matches the state's purpose better. WDYT?

ono-max (approved):
Great! Thank you 👍

PR #927, opened 2023-03-10
Fix test builder (undefined method `debug_code`)

andyw8:
`debug_code` is defined only in `ConsoleTestCase`, not `TestCase`, so tests generated using `bin/gentest` were failing with `NoMethodError: undefined method `debug_code'`.


ono-max (approved):
👍

PR #928, opened 2023-03-12
Run the test for assert_line_text only locally

ono-max:
`test_the_helper_takes_a_string_expectation_and_escape_it` is sufficient to meet the requirements for local testing only.


ono-max:
This is not a bottleneck for slow tests.

PR #930, opened 2023-03-14
Run PseudoTerminalTest in each mode separately

ono-max:
debug_code method runs local and remote tests in multi threads. To test them correctly, we should test them separately.

### DIFF
--- a/test/support/test_case_test.rb
+++ b/test/support/test_case_test.rb
@@ -12,7 +12,7 @@ module DEBUGGER__
 
     def test_the_test_fails_when_debugger_exits_early
       assert_raise_message(/Expected all commands\/assertions to be executed/) do
-        debug_code(program) do
+        debug_code(program, remote: false) do
           type 'continue'
           type 'foo'
         end
@@ -35,7 +35,7 @@ module DEBUGGER__
 
     def test_the_test_fails_when_the_repl_prompt_does_not_finish_even_though_scenario_is_empty
       assert_raise_message(/Expected the REPL prompt to finish/) do
-        debug_code(program) do
+        debug_code(program, remote: false) do
         end
       end
     end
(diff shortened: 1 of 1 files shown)


st0012:
I think it's probably worth running them on remote tests as well. If we look at [all `debug_code_on_*` methods](https://github.com/ruby/debug/blob/master/test/support/console_test_case.rb#L248-L266), we can see that only remote scenarios have `test_info.remote_info`, which does trigger different code path in [`run_test_scenario`](https://github.com/ruby/debug/blob/a9bec2540e68088be9d0ffcddb7d6bc0cf03929b/test/support/console_test_case.rb#L180-L182).
What's the benefit of skipping them? Does it reduce the test duration significantly?

ono-max:
> Does it reduce the test duration significantly?
Yes.

And I understand your opinion. Let me change it.

ono-max:
I'll create other PRs to make tests faster.

ono-max:
Thank you for advising me.

st0012 (commented):
Sorry for nitpicking: but neither `prepare_test_environment` nor `debug_code_on_unix_domain_socket` is public API. Testing them directly is probably not the best idea as they'll make future refactoring harder.

Can we test them as:

- `debug_code(program, remote: false)`
- `debug_code(program, remote: :remote_only)`

It'll make testing styles more consistent too (e.g. we can pass steps as the block argument instead of a block object)

Also, since they're all testing pseudo terminal and uses the same test setup, do we need to create another test class for remote scenarios?

ono-max:
> debug_code(program, remote: :remote_only)

I'd like to test if the debugger exits early in the UNIX domain socket and TCP/IP mode correctly. If we test in ":remote_only", we can not guarantee both mode works. Also, testing private API is not bad because these tests are for the test framework.

> Also, since they're all testing pseudo terminal and uses the same test setup, do we need to create another test class for remote scenarios?

I felt that PseudoTerminalTest class seems to be for local because PseudoTerminalTestForRemoteDebuggee is defined. That's why I defined a new class, but it's ok if you want to add these test cases to PseudoTerminalTest class. Do you want to do it?

ono-max:
Let me close this PR. I'll reopen it if necessary.

PR #938, opened 2023-03-22
Change the method name dap_event to process_protocol_request_event

ono-max:
Rename to be consistent with process_protocol_request method


ko1:
:dap/cdp_result also should be changed. Let me take a try.

PR #939, opened 2023-03-22
DAP: support custom request in session class

ono-max:
UI_DAP -> Session: custom_request_...

Session -> ThreadClient: custom_request_event_...

Add "request_event" prefix to clarify it is a response (not Events in DAP)


lib/debug/server_dap.rb
```diff
@@ -656,7 +656,7 @@ def process_protocol_request req
           fail_response req
         end
       else
-        if respond_to? mid = "request_#{req['command']}"
+        if respond_to? mid = "custom_request_#{req['command']}"
```
st0012, line 659: Extension APIs like this should be considered a breaking change and will require bumping a major version number (e.g. `2.0`) as apps and libraries extending `debug` through it will suddenly stop working.
But given now there probably isn't anyone using it yet, it **may** not break anything yet.

ko1: Now we don't publish this feature in the doc so now it is not public API.

Issue #877, opened 2022-12-23
Debugger freezes when triggering timeout inside a timeout block with Puma server


st0012:
**Your environment**

* `ruby -v`: `3.1.3`
* `rdbg -v`: `1.7.1`

**Describe the bug**

When the following conditions all match, the debugger has a high possibility to freeze:

- Inside a Rails app with Puma web server
- The breakpoint is inside a timeout call, like

    ```rb
      def hello
        Timeout.timeout(3600) do
          binding.b
        end
      end
    ```

- From the breakpoint, executes something that calls `Timeout.timeout` underneath

**To Reproduce**

I've documented reproduction steps in [this repo](https://github.com/st0012/debug-freeze-issue).

**Expected behavior**

The debugger doesn't freeze.

**Additional context**

It may not be the root cause, but `Timeout`'s thread being stopped by the debugger plays a major role here. 

If we somehow modify [this method](https://github.com/ruby/debug/blob/master/lib/debug/session.rb#L1654-L1664) to skip pausing the `Timeout` thread, like

```rb
    private def thread_stopper
      TracePoint.new(:line) do
        # run on each thread
        tc = ThreadClient.current
        next if tc.management?
        next unless tc.running?
        next if tc == @tc
        next if tc.thread.name && tc.thread.name.match?(/Timeout/)

        tc.on_pause
      end
    end
```

Then the issue doesn't occur.

ko1:
Thank you for the report.

* Is it difficult to repro within a single file? (without puma)
* Can you show the backtrace when thread_stopper's TP stoppes?

by modifying like that:

``` ruby
   private def thread_stopper
      TracePoint.new(:line) do
        # run on each thread
        tc = ThreadClient.current
        next if tc.management?
        next unless tc.running?
        next if tc == @tc
        STDERR.puts caller.inspect if tc.thread.name.match?(/Timeout/)

        tc.on_pause
      end
    end
```

ko1:
Anyway, I assume that it is difficult to fix in theory.

st0012:
> Is it difficult to repro within a single file? (without puma)

I tried it but didn't succeed. I suspect running it from a thread pool is one criteria and will experiment it later.

> Can you show the backtrace when thread_stopper's TP stoppes?

```rb
"/Users/hung-wulo/.gem/ruby/3.1.2/gems/timeout-0.3.1/lib/timeout.rb:112:in `block (2 levels) in create_timeout_thread'", 
[...]

PR #947, opened 2023-03-28
restart all threads on eval

ko1:
When a thread keeps a lock, and REPL runs a code which needs the
lock, other threads should make a progress to release the lock.

fix https://github.com/ruby/debug/issues/877


st0012:
This is amazing! Thank you for the fix 🙏

PR #950, opened 2023-03-29
Restart threads on DAP evaluation

st0012:
This makes sure DAP evaluation request also doesn't hang.

See #947 for the original console implementation.

### DIFF
--- a/test/protocol/eval_test.rb
+++ b/test/protocol/eval_test.rb
@@ -57,4 +57,27 @@ module DEBUGGER__
       end
     end
   end
+
+  class EvaluateThreadTest < ProtocolTestCase
+    PROGRAM = <<~RUBY
+      1| th0 = Thread.new{sleep}
+      2| m = Mutex.new; q = Queue.new
+      3| th1 = Thread.new do
+      4|   m.lock; q << true
+      5|   sleep 1
+      6|   m.unlock
+      7| end
+      8| q.pop # wait for locking
+      9| p :ok
+    RUBY
+
+    def test_eval_with_threads
+      run_protocol_scenario PROGRAM, cdp: false do
+        req_add_breakpoint 9
+        req_continue
+        assert_repl_result({value: 'false', type: 'FalseClass'}, 'm.lock.nil?', frame_idx: 0)
+        req_continue
+      end
+    end
+  end
 end
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -625,6 +625,7 @@ module DEBUGGER__
           expr = req.dig('arguments', 'expression')
 
           if find_waiting_tc(tid)
+            restart_all_threads
             request_tc [:dap, :evaluate, req, fid, expr, context]
           else
             fail_response req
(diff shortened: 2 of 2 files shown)


st0012:
@ono-max I tried implementing the same change for CDP as well but it still hangs. So I will leave that to you.

ko1:
```patch
[master]$ git diff
diff --git a/lib/debug/server_dap.rb b/lib/debug/server_dap.rb
index 8298428..bae1e4d 100644
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -625,6 +625,7 @@ module DEBUGGER__
           expr = req.dig('arguments', 'expression')

           if find_waiting_tc(tid)
+            restart_all_threads
             request_tc [:dap, :evaluate, req, fid, expr, context]
           else
             fail_response req
@@ -701,6 +702,7 @@ module DEBUGGER__
         register_vars result[:variables], tid
         @ui.respond req, result
       when :evaluate
+        stop_all_threads
         message = result.delete :message
         if message
           @ui.respond req, success: false, message: message
```

could you add `stop_all_threads`?

st0012:
Ah I thought it's handled in `Session#process_event` too. I've updated code but will try to put up a test case for thread-stopping too.

test/protocol/eval_test.rb
```diff
@@ -57,4 +57,57 @@ def test_eval_evaluates_arithmetic_expressions
...
+      run_protocol_scenario program, cdp: false do
+        req_add_breakpoint 11
+        req_continue
+        assert_repl_result({value: '1', type: 'Integer'}, '1', frame_idx: 0)
+        locals = gather_variables
+        count_var_value_1 = locals.find{|v| v[:name] == 'count'}[:value]
+        locals = gather_variables
+        count_var_value_2 = locals.find{|v| v[:name] == 'count'}[:value]
+        # if the thread is stopped, the value of count will not be changed.
+        assert_equal(count_var_value_1, count_var_value_2)
```
ko1, line 107: This test is not acceptable because `sleep 0.05` introducing nondeterministic results.

st0012: I've updated the test to not use `sleep`. And the test failure doesn't seem to be related (another [occurrence](https://github.com/ruby/debug/actions/runs/4516497233/jobs/7954891876))

st0012:
I think I found another issue: when I run the test, the result was not deterministic even without sleep and fails around 10% of the time on my machine. 

I initially thought it's because of my test, but then I found that it's because [this early return](https://github.com/ruby/debug/blob/master/lib/debug/session.rb#L1679) isn't always correct. Sometimes it doesn't reflect the actual running thread count and thus the debugger doesn't stop the threads. If I comment it out, the test doesn't fail intermittently anymore. Should I address that in this PR too?

ko1:
Now current behavior is already non-deterministic (by design).
Essentially this is hard to write a test for such cases.

st0012:
As the test could make CI unstable, I've dropped it.

PR #951, opened 2023-03-29
Improve assert_repl_result for CDP

st0012:
The current implementation doesn't work when the result is a boolean type because the helper calls `#inspect` on the values of `JAVASCRIPT_TYPE_TO_CLASS_MAPS`, which is an array for boolean types.

Example failure:

```
<["[TrueClass, FalseClass]"]> was expected to include
<"FalseClass">.
```

Since `JAVASCRIPT_TYPE_TO_CLASS_MAPS` is never used in other places, we can just store the String representation of the class name in the hash and avoid calling `#inspect` on the values.

### DIFF
--- a/test/protocol/eval_test.rb
+++ b/test/protocol/eval_test.rb
@@ -13,13 +13,14 @@ module DEBUGGER__
       6| f = 6
     RUBY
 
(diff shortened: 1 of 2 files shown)


ono-max (approved):
Thank you 👍

PR #957, opened 2023-04-03
Improve variable type name request

adisonlampert:
## Description

It is possible to override the `name` method in a class with a method that has arguments, resulting in an `ArgumentError` when calling `klass.name`. #790 addresses this by handing the exception and setting the variable's `type` to `"<Error: #{e.message} (#{e.backtrace.first}>"`.

This PR improves the handling of the case where the `class` method is overwritten and removes the need for the `type_name` method by using `M_NAME.bind_call(klass)` instead of calling the `name` method. Instead of the variable's type being the error message, it is now more accurate:
#### Before
```
{
          "name": "f",
          "value": "#<Foo:0x00000001090377a8>",
          "type": "<Error: wrong number of arguments (given 0, expected 1) (/var/folders/lf/pp4810hd7n5d0c_zs0skjxfh0000gn/T/debug-20230329-68605-bb7wh3.rb:5:in `to_s'>",
          "variablesReference": 4,
          "indexedVariables": 0,
          "namedVariables": 1
}
```

#### After 
```
{
          "name": "f",
          "value": "#<Foo:0x00000001090377a8>",
          "type": "Foo",
          "variablesReference": 4,
          "indexedVariables": 0,
          "namedVariables": 1
}
```

(This example is from the test included in this PR)

### DIFF
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -1033,14 +1033,14 @@ module DEBUGGER__
       if name
         { name: name,
           value: str,
-          type: type_name,
+          type: type_name || klass.to_s,
           variablesReference: vid,
           indexedVariables: indexedVariables,
           namedVariables: namedVariables,
         }
       else
         { result: str,
-          type: type_name,
+          type: type_name || klass.to_s,
           variablesReference: vid,
           indexedVariables: indexedVariables,
           namedVariables: namedVariables,
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -1010,16 +1010,6 @@ module DEBUGGER__
       variable nil, r
     end
 
-    def type_name obj
-      klass = M_CLASS.bind_call(obj)
-
-      begin
-        klass.name || klass.to_s
-      rescue Exception => e
-        "<Error: #{e.message} (#{e.backtrace.first}>"
-      end
-    end
-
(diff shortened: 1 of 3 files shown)


lib/debug/server_dap.rb
```diff
@@ -1010,16 +1010,6 @@ def evaluate_result r
       variable nil, r
     end
 
-    def type_name obj
-      klass = M_CLASS.bind_call(obj)
-
-      begin
-        klass.name || klass.to_s
-      rescue Exception => e
```
ko1, line 1018: why not drop `'klass.to_s`?

ko1: No. `klass.name` can return `nil`.

adisonlampert: Ohh, I see. Thanks for the catch! 
I updated the code to use `to_s` and added a test for anonymous class instances to ensure the `type` is `""`.

ko1: Do you really need `nil` for the name?
Previous one is `klass.name || klass.to_s` but your code uses `klass.name.to_s`.

ko1: The previous code means that if `klass.name` is given, use it, otherwise use `klass.to_s`.

ko1: In this case, the error because of `to_s` is ignored.
I'll fix it after merging.

PR #958, opened 2023-04-04
fix(console): Pass IO-like object to `pp` (pretty-print)

DerekStride:
## Description

Passing a String object to `pp` via `IRB::ColorPrinter.pp` raises an error if the object implementing `#pretty_print(pp)` expects the `pp.output` object to act like an IO object. It raises this error:

```
NoMethodError: private method `puts' called for "":String> rescued during inspection
```

### Example

[`TreeStand::Node#pretty_print`](https://github.com/Shopify/tree_stand/blob/ef153fe120632d43170a10f286bf9d79745c9e93/lib/tree_stand/node.rb#L211-L214) (calls [`io.puts`](https://github.com/Shopify/tree_stand/blob/main/lib/tree_stand/utils/printer.rb#L32))

### DIFF
--- a/lib/debug/color.rb
+++ b/lib/debug/color.rb
@@ -50,7 +50,7 @@ module DEBUGGER__
       def color_pp obj, width
         with_inspection_error_guard do
           if !CONFIG[:no_color]
-            IRB::ColorPrinter.pp(obj, "".dup, width)
+            IRB::ColorPrinter.pp(obj, StringIO.new, width)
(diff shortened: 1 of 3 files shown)


DerekStride:
Closing, looks like `PP#output` is expected to respond to `<<` instead.

https://ruby-doc.org/3.2.2/stdlibs/prettyprint/PrettyPrint.html

PR #959, opened 2023-04-08
DAP: introduce Rdbg Trace Inspector

ono-max:
# Summary

We introduce Rdbg Trace Inspector. This is useful when tracing in VS Code. Users can search tracing logs easily and get a lot of useful information by using Rdbg Trace Inspector. For example, users can know which method is called, which line is executed, local variables at some point, and so on.

# How to use it

1. Click on "Start Trace" in TRACE INSPECTOR view
2. Execute the program

That's it! You can check the location and returned value by clicking the trace log.

# Several demos

## Trace Call/Return

Users can know which method is called, its parameters, and return values.

https://user-images.githubusercontent.com/59436572/232519376-f6970c81-90d9-4312-9b0e-b76bfe2ebff2.mov

## Record & Replay

This is slow, but Users can get a wealth of information by enabling Record & Replay. For example, users can know the backtrace and local variables during the execution.

https://user-images.githubusercontent.com/59436572/232519356-fc12fb49-8249-46c1-829c-8fd959dbe780.mov

# Other

Trace Call and Trace Line are enabled in the default. Users can select Record & Replay from the menu.
Also, users can enter Regular Expressions to trace only matched events as follows.

https://user-images.githubusercontent.com/59436572/232519404-6d0ce171-e5bc-442f-8627-cdef03fd21e3.mov

Users can copy the specific log as follows:

https://user-images.githubusercontent.com/59436572/234153138-93a05fce-4a3a-44be-a11d-dd1a43a88d64.mov

Some Notes: This PR is the combination of https://github.com/ruby/debug/pull/946 and https://github.com/ruby/debug/pull/916. After discussing with @ko1-san, we decided to combine record inspector and trace inspector.

### DIFF
--- a/lib/debug/dap_custom/traceInspector.rb
+++ b/lib/debug/dap_custom/traceInspector.rb
@@ -1,0 +1,334 @@
+module DEBUGGER__
+  module DAP_TraceInspector
+    class MultiTracer < Tracer
+      def initialize ui, evts, trace_params, max_log_size: nil, **kw
+        @evts = evts
+        @log = []
+        @trace_params = trace_params
+        if max_log_size
+          @max_log_size = max_log_size
+        else
+          @max_log_size = 50000
+        end
+        @dropped_trace_cnt = 0
+        super(ui, **kw)
+        @type = 'multi'
+        @name = 'TraceInspector'
+      end
+  
+      attr_accessor :dropped_trace_cnt
+      attr_reader :log
+  
+      def setup
+        @tracer = TracePoint.new(*@evts){|tp|
+          next if skip?(tp)
+  
+          case tp.event
+          when :call, :c_call, :b_call
+            if @trace_params
+              params = parameters_info tp
+            end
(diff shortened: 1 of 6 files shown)


lib/debug/frame_info.rb
```diff
@@ -147,6 +147,17 @@ def local_variables
       end
     end
 
+    def get_params
```
ko1, line 150:
```ruby
def iseq_parameters_info
...
when :block, :method
  parameters_info
else
  nil
end
```

lib/debug/dap_custom/traceInspector.rb
```diff
@@ -0,0 +1,245 @@
+module DEBUGGER__
+  module DAP_TraceInspector
+    class MultiTracer < Tracer
+      MAX_LOG_SIZE = 4000
+      def initialize ui, evts, trace_params, **kw
```
ko1, line 5: max_log_size: 4000

PR #964, opened 2023-04-16
Move `Session#exnted_feature` into Session class

ono-max:
Currently, `exnted_feature` method is defined outside of Session class. This PR fixes it.


choosen:
@ono-max  @ko1  There is a typo here and in changelog: `#extend_feature` 
good that it is correct in code :D

Issue #975, opened 2023-05-05
trace method can't accept number as first argument


MasanoriOnishi:
**Your environment**

* `ruby -v`:  ruby 3.1.3p185 (2022-11-24 revision 1a6b16756e) [x86_64-linux]
* `rdbg -v`: rdbg 1.7.2

**Describe the bug**
In the following code, [Signal.#trap](https://docs.ruby-lang.org/ja/latest/method/Signal/m/trap.html
) method is overridden.
https://github.com/ruby/debug/blob/v1.7.2/lib/debug/session.rb#L2520

Although the original trap code can accept Integer, the above can't.

**To Reproduce**

```
irb(main):001:0> require 'debug'
=> true
irb(main):002:0> trap(3,1)
/usr/local/bundle/gems/debug-1.7.2/lib/debug/session.rb:2520:in `trap': undefined method `to_sym' for 3:Integer (NoMethodError)

      case sig&.to_sym
              ^^^^^^^^
Did you mean?  to_s
	from (irb):2:in `<main>'
	from /usr/local/bundle/gems/irb-1.6.4/exe/irb:9:in `<top (required)>'
	from /usr/local/bundle/bin/irb:25:in `load'
	from /usr/local/bundle/bin/irb:25:in `<main>'
```

**Expected behavior**
trace method should accept number as first argument.

PR #976, opened 2023-05-05
Enable trap function accept Integer(Non-String)

MasanoriOnishi:
## Description
In the following code, [Signal.#trap](https://docs.ruby-lang.org/ja/latest/method/Signal/m/trap.html) method is overridden.
https://github.com/ruby/debug/blob/v1.7.2/lib/debug/session.rb#L2520

Although the original trap code can accept Integer, the above can't.

That's why I enabled trap function accept Integer(Non-String).

Relates to https://github.com/ruby/debug/issues/975

### DIFF
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -2524,7 +2524,7 @@ module DEBUGGER__
 
   module TrapInterceptor
     def trap sig, *command, &command_proc
-      case sig&.to_sym
+      case sig&.to_s&.to_sym
       when :INT, :SIGINT
         if defined?(SESSION) && SESSION.active? && SESSION.intercept_trap_sigint?
           return SESSION.save_int_trap(command.empty? ? command_proc : command.first)


lib/debug/session.rb
```diff
@@ -2524,7 +2524,7 @@ def daemon(*args)
 
   module TrapInterceptor
     def trap sig, *command, &command_proc
-      case sig&.to_sym
+      case sig&.to_s&.to_sym
```
amomchilov, line 2527: This won't handle integers correctly, because they won't fall into `:INT` and `:SIGINT` below. This would do it:

```ruby
      case sig&.to_s&.to_sym
      when :INT, :SIGINT, :"2"
        if defined?(SESSION) && SESSION.active? && SESSION.intercept_trap_sigint?
          return SESSION.save_int_trap(command.empty? ? command_proc : command.first)
        end
      end
```

ko1: Sorry I missed your comment. Could you make an issue or PR for it with examples?

amomchilov: @ko1 Nevermind, you already fixed this in #976 :)

PR #977, opened 2023-05-05
work correctly with trap(Integer,...)

ko1:
follow up for https://github.com/ruby/debug/pull/976

### DIFF
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -2524,6 +2524,16 @@ module DEBUGGER__
 
   module TrapInterceptor
     def trap sig, *command, &command_proc
+      sym =
+        case sig
+        when String
+          sig.to_sym
+        when Integer
+          Signal.signame(sig)&.to_sym
+        else
+          sig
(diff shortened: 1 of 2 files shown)


lib/debug/session.rb
```diff
@@ -2524,6 +2524,16 @@ def daemon(*args)
...
+      sym =
+        case sig
+        when String
+          sig.to_sym
+        when Integer
+          Signal.signame(sig)&.to_sym
+        else
+          sig
+        end
+
```
MasanoriOnishi, line 2536: @ko1 Thank you for your fast check https://github.com/ruby/debug/pull/976.

And your code is very well cared for and excellent.

I think the following code will be more better if the fix is applied?
https://github.com/ruby/debug/pull/977/files#diff-c87adef8ace662a03249d206e6812162256bbcecabe18cd51861d1e1cd9cdfcaR2537

`case sig&.to_s&.to_sym` → `case sym`

How do you think?

PR #984, opened 2023-05-27
Improve the performance for record & replay

ono-max:
Here is a performance comparison.

Execution Time(s) | Ratio to existing implementation
-- | --
Existing implementation | 73.743 | 1.0
Current Proposal | 46.616 | 1.58

# Script for measuring performance
```ruby
def index
    binding.break # <- Start a program here
    1_000.times {
      respond_to do |format|
        format.html # index.html.erb
      end
    }
  end #<- Pause a program here
```

* ruby 3.2.2 (2023-03-30 revision e51014f9c0) [arm64-darwin21]
* MacBook Pro (16-inch, 2021)
* Chip: Apple M1 Max
* Memory: 32 GB

### DIFF
--- a/lib/debug/dap_custom/traceInspector.rb
+++ b/lib/debug/dap_custom/traceInspector.rb
@@ -101,6 +101,11 @@ module DEBUGGER__
     end
 
     class Custom_Recorder < ThreadClient::Recorder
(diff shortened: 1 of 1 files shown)


ko1:
Could you explain your idea?

PR #986, opened 2023-06-03
DAP: add the ability to hide the #class item for each object

ono-max:
We can hide the #class item for each object by setting true in hideClassItem.

Closes https://github.com/ruby/vscode-rdbg/issues/37.

# Before
Screen Shot 2023-06-03 at 13 34 01

# After

Screen Shot 2023-06-03 at 13 33 26

### DIFF
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -301,6 +301,14 @@ module DEBUGGER__
       send_event :terminated unless @sock.closed?
     end
 
+    def hide_class_item req
+      if req.dig('arguments', 'hideClassItem') == true
+        @hide_class_item = true
+      else
+        @hide_class_item = false
+      end
+    end
+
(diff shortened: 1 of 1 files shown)


amomchilov:
Perhaps there should be a special case for the most common values like:

* `true`/`false`/`nil`
* Integers
* Floats
* Strings

amomchilov:
I implemented this in https://github.com/ruby/debug/pull/1004/files, with a few improvements:

1. Because of the improvements in #1001, the implementation applies to both DAP and CDP.
2. The bug fixes in #1001 make the counts correct, so you don't see the disclosure triangles for items that have no members. (e.g. `⋁ a: 1` will now just be `a: 1`)
3. It still keeps `#class` when it helps.

I would love to hear your feedback!

ko1:
discuss on #1004 !

Issue #486, opened 2022-01-12
Debugging Async tasks yields `RuntimeError: running is given, but running`


machty:
**Your environment**

* `ruby -v`: ruby 3.1.0p0 (2021-12-25 revision fb4df44d16) [x86_64-darwin20]
* `rdbg -v`: rdbg 1.4.0

**Describe the bug**

Debugging programs with the Async library often leads to `RuntimeError: running is given, but running`

**To Reproduce**

Install `async` 2.0.0:

```
gem install async
```

Create ruby script `async_debug_test_case.rb` as follows:

```rb
require 'async'

Async do |t|
  t.async do
    puts "1\n"
  end
  t.async do
    puts "2\n"
  end
end
```

Run rdbg, set breakpoints on both `puts`s, and continue until error is raised. See output below:

```
rdbg scratch/async_debug_test_case.rb 
[1, 10] in scratch/async_debug_test_case.rb
=>   1| require 'async'
     2| 
     3| Async do |t|
     4|   t.async do
     5|     puts "1\n"
     6|   end
     7|   t.async do
     8|     puts "2\n"
     9|   end
=>#0    <main> at scratch/async_debug_test_case.rb:1
(rdbg) b 5    # break command
#0  BP - Line  /Users/machty/code/exc/rails/scratch/async_debug_test_case.rb:5 (line)
(rdbg) b 8    # break command
#1  BP - Line  /Users/machty/code/exc/rails/scratch/async_debug_test_case.rb:8 (line)
(rdbg) c    # continue command
[1, 10] in scratch/async_debug_test_case.rb
     1| require 'async'
     2| 
     3| Async do |t|
     4|   t.async do
=>   5|     puts "1\n"
     6|   end
     7|   t.async do
     8|     puts "2\n"
    10| end
=>#0    block in <main> (2 levels) at scratch/async_debug_test_case.rb:5
  #1    block in schedule at ~/.gem/ruby/3.1.0/gems/async-2.0.0/lib/async/task.rb:258

Stop by #0  BP - Line  /Users/machty/code/exc/rails/scratch/async_debug_test_case.rb:5 (line)
(rdbg) c    # continue command
[3, 11] in scratch/async_debug_test_case.rb
     3| Async do |t|
     4|   t.async do
     5|     puts "1\n"
     6|   end
     7|   t.async do
=>   8|     puts "2\n"
     9|   end
    10| end
    11| 
[...]

PR #987, opened 2023-06-04
Fix compatibility with Fiber Scheduler.

ioquatix:
This is a first pass at improving the compatibility with the fiber scheduler and contains two changes:

- Ensure the `ThreadClient` is actually per-thread and not per-fiber.
- Ensure that waiting for events (blocking) is done in a blocking context, otherwise it will be redirected in the fiber scheduler which is probably not desirable.

It only works on 3.2 but could be made to work on 3.0 and 3.1 if some tests were changed.

`Fiber.blocking{...}` was introduced in Ruby 3.2. It can be emulated by `Fiber.new(blocking: true) {...}` but the backtrace is changed. This causes some test failures.

My personal opinion is that we can ignore 3.0 and 3.1 but there is an option to support them with the compatibility wrapper outlined above, if we are willing to change some of the tests that check backtraces.

I cannot see any other obvious issues with the `debug` code. If there are other places where the `debug` code is executing blocking operations, we should add more `Fiber.blocking{...}` to avoid entering the fiber scheduler.

### DIFF
--- a/test/support/cdp_utils.rb
+++ b/test/support/cdp_utils.rb
@@ -63,7 +63,7 @@ module DEBUGGER__
       MSG
     end
 
-    TIMEOUT_SEC = (ENV['RUBY_DEBUG_TIMEOUT_SEC'] || 10).to_i
+    TIMEOUT_SEC = (ENV['RUBY_DEBUG_TIMEOUT_SEC'] || 30).to_i
     HOST = '127.0.0.1'
 
     def run_cdp_scenario program, &msgs
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -865,12 +865,15 @@ module DEBUGGER__
       private def fiber_blocking
         ::Fiber.blocking{yield}
       end
+    elsif ::Fiber.method_defined?(:blocking?)
+      private def fiber_blocking
+        ::Fiber.new(blocking: true){yield}.resume
+      end
     else
       private def fiber_blocking
         yield
       end
     end
-    
 
     def wait_next_action
       fiber_blocking{wait_next_action_}
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -862,8 +862,19 @@ module DEBUGGER__
     class SuspendReplay < Exception
     end
 
+    if ::Fiber.respond_to?(:blocking)
+      private def fiber_blocking
+        ::Fiber.blocking{yield}
+      end
+    else
+      private def fiber_blocking
+        yield
+      end
+    end
+    
+
     def wait_next_action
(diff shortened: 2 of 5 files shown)


ioquatix:
@ko1 can you please review and also if you want tests, can you suggest what kind of tests to add? The original issue <https://github.com/ruby/debug/issues/486> has a complete example and this PR fixes it. However, as you said, you may not want to pull in `async` to run that sample.

louim:
Hello! First, thank you @ioquatix for this patch. We are running ruby 3.1 and encountered the bug mentioned in #486. Since this patch is for ruby 3.2+, I went ahead an applied your suggestion to make it work with Ruby 3.1. [Here's the branch, rebased on master and with the patch you mentioned](https://github.com/ruby/debug/compare/master...didacte:debug:bug/fiber-scheduler-31?expand=1). It fixed our problem 🎉 .

I'd be happy to help get this in the debug gem itself. What would be the way forward to make this happen? @ko1, I'd be happy to help in any way I can to get this merged 🚀 .

louim:
Seems like I spoke too soon 🙃. @ioquatix using `Fiber.new(blocking: true) {yield}` does not seem to stop the debugger. Do we need to wrap it with something else?

ioquatix:
@louim can you try with Ruby 3.1.3?

There were some bug fixes which I hope were backported.

louim:
@ioquatix sorry, I didn’t mention it, but we are in fact using ruby 3.1.3.

ioquatix:
@louim Can you test with this?

```ruby
    if ::Fiber.respond_to?(:blocking)
      private def fiber_blocking
        ::Fiber.blocking{yield}
      end
    else
      private def fiber_blocking
        ::Fiber.new{blocking: true}{yield}.resume
      end
    end
```

ioquatix:
With the latest change, the test failures on < 3.2 are stack trace related, as we can't control the stack when using `Fiber.new(blocking: true){yield}.resume`. So we should skip those tests?

test/support/cdp_utils.rb
```diff
@@ -63,7 +63,7 @@ def connect_to_cdp_server
       MSG
     end
 
-    TIMEOUT_SEC = (ENV['RUBY_DEBUG_TIMEOUT_SEC'] || 10).to_i
+    TIMEOUT_SEC = (ENV['RUBY_DEBUG_TIMEOUT_SEC'] || 30).to_i
```
ko1, line 66: why do you extend the timeout?
this extension will introduce huge slowdown on some cases.

ioquatix: I found the test runner was timing out on GitHub actions even in HEAD.

PR #995, opened 2023-07-07
Allow starting debug server without opening apps

amomchilov:
## Description

Hey there!

I'm developing my own GUI debugger that uses the Debug Adapter Protocol. This gem supports it, but it's coupled to VSCode: you can only start the DAP server if you also open VSCode automatically.

In the modular spirit of DAP and CDP, it should be possible to start debug servers that use those protocols, without actually opening VSCode or Chrome, so other debugger clients can be connected instead.

## TODO

- [ ] Tests
- [ ] Ensure this works for starting a CDP session without starting Chrome
- [ ] Update documentation, readme, etc.


ko1:
I don't have any objection to have this feature, but please describe the user facing UI changes.

johnhamelink:
@ko1 Hi, I'm an Emacs user and we have two DAP clients available:

 - https://github.com/emacs-lsp/dap-mode
 - https://github.com/svaante/dape

If a user wanted to try running these modes with ruby debug, they'd want to use a configuration item like the one in this PR, since they won't necessarily have VSCode on their system.

amomchilov:
> please describe the user facing UI changes.

@ko1 I updated the README. Let me know if that answers your questions.

PR #1001, opened 2023-07-26
Share variable inspection logic between CDP and DAP

amomchilov:
Thanks for your Pull Request 🎉 

**Please follow these instructions to help us review it more efficiently:**

- Add references of related issues/PRs in the description if available.
- If you're updating the readme file, make sure you followed [the instruction here](https://github.com/ruby/debug/blob/master/CONTRIBUTING.md#to-update-readme).

## Description
Describe your changes:


lib/debug/server_dap.rb
```diff
@@ -875,48 +876,18 @@ def process_dap args
...
-            unless NaiveString === obj
-              vars += M_INSTANCE_VARIABLES.bind_call(obj).sort.map{|iv|
-                variable(iv, M_INSTANCE_VARIABLE_GET.bind_call(obj, iv))
-              }
-              vars.unshift variable('#class', M_CLASS.bind_call(obj))
-            end
+            VariableInspector.new.named_members_of(obj)
           end
+
+          vars = members.map { |member| variable(member.name, member.value) }
```
amomchilov, line 890: There's nothing implemented like CDP's `internalProperty` yet, but I think support for  `presentationHint` can be added here eventually to achieve a similar effect.

lib/debug/variable_inspector.rb
```diff
@@ -0,0 +1,75 @@
...
+    def value_inspect obj, short: true
+      # TODO: max length should be configurable?
+      str = DEBUGGER__.safe_inspect obj, short: short, max_length: MAX_LENGTH
+
+      if str.encoding == Encoding::UTF_8
+        str.scrub
+      else
+        str.encode(Encoding::UTF_8, invalid: :replace, undef: :replace)
+      end
+    end
```
amomchilov, line 73: I just copied this snippet for now, but my eventual goal is to move all presentation logic out of `ThreadClient` and into a shared class like this.

lib/debug/variable_inspector.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    def indexed_members_of obj, start:, count:
+      (start ... (start + count)).map { |i| Member.new(name: i.to_s, value: obj[i]) }
+    end
+
+    def named_members_of obj
+      members = case obj
+      when Array then obj.map.with_index { |o, i| Member.new(name: i.to_s, value: o) }
+      when Hash then obj.map { |k, v| Member.new(name: value_inspect(k), value: v) }
```
amomchilov, line 30: This used to be inconsistent before. VSCode would format Hash keys with `value_inspect`, but not Chrome. Now they will both us it.

lib/debug/server_cdp.rb
```diff
@@ -1112,46 +1113,25 @@ def process_cdp args
...
+
+          result = members.filter_map do |member|
+            next if member.internal?
+            variable(member.name, member.value)
           end
 
-          result += M_INSTANCE_VARIABLES.bind_call(obj).map{|iv|
-            variable(iv, M_INSTANCE_VARIABLE_GET.bind_call(obj, iv))
-          }
-          prop += [internalProperty('#class', M_CLASS.bind_call(obj))]
```
amomchilov, line 1152: For Chrome, `#class` is appended as the last prop, whereas for VSCode it was `unshift`ed to the top.

ko1:
* I like an idea to unify CDP and DAP because of implementation.
* However I'm not sure we can merge it because of difference between CDP and DAP (I don't know the CDP details). Do you have a time to trouble shooting for CDP?
* The name "variable.rb" doesn't make sense for me because it's only for the representation (the name is too general). debug.gem has a policy to reduce the number of "require" to reduce the boot time. Can we merge it with variable_inspector.rb?

amomchilov:
> Do you have a time to trouble shooting for CDP?

I've already run this with CDP, it works great.

> Can we merge it with variable_inspector.rb?

Sure! I'll still call it `class Variable`. It's a general name, but I don't have any better ideas. Please suggest if you have any.

lib/debug/variable_inspector.rb
```diff
@@ -0,0 +1,138 @@
+# frozen_string_literal: true
+
+require_relative 'limited_pp'
+
+module DEBUGGER__
+  class VariableInspector
+    def indexed_members_of(obj, start:, count:)
+      return [] if start > (obj.length - 1)
```
ono-max, line 8: In this implementation, the `#class` will disappear when displaying Array. Is this intentional behavior?

ono-max:
Also, some tests are failed in protocol part. Can you fix them?

lib/debug/variable_inspector.rb
amomchilov: That makes sense (in the same spirit as https://github.com/ruby/debug/pull/1004), though perhaps we should keep the `#class` if it's a non-standard subclass of `Array`. I'll make that in a separate PR.

amomchilov:
@ono-max I can have a look over the holiday break next week :)

lib/debug/server_dap.rb
```diff
@@ -867,56 +862,28 @@ def process_dap args
...
         vars = collect_locals(frame).map do |var, val|
-          variable(var, val)
+          render_variable Variable.new(name: var, value: val)
         end
 
         event! :protocol_result, :scope, req, variables: vars, tid: self.id
       when :variable
         vid = args.shift
-        obj = @var_map[vid]
-        if obj
```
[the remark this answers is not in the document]
amomchilov: ... this change allocates one more Member than before:

`#<Member name="#class" value=NilClass internal>`

Which bumps some of the `variablesReference` numbers in `test/protocol/step_back_raw_dap_test.rb` by 1.

amomchilov:
Hey @ono-max, Happy new year!

I've fixed the protocol tests, and rebased onto the latest master.

There's 2 test failures in `test/console/nested_break_test.rb` left, but I don't see how they could be related.

lib/debug/server_cdp.rb
```diff
@@ -1112,46 +1113,31 @@ def process_cdp args
...
-            result = obj.instance_variables.map{|iv|
-              variable(iv, obj.instance_variable_get(iv))
-            }
-            prop = [internalProperty('%ancestors', obj.ancestors[1..])]
-          when Range
-            prop = [
-              internalProperty('#begin', obj.begin),
-              internalProperty('#end', obj.end),
-            ]
+        if @obj_map.key?(oid)
```
ono-max, line 1117: Good catch. I missed the edge case for falsy value.

lib/debug/variable_inspector.rb
```diff
@@ -0,0 +1,138 @@
...
+
+      capped_count = [count, obj.length - start].min
```
ono-max, line 10: Can you add the comment why we need to compare  `count` and `obj.length - start`? Sorry, I could not understand the reason.

amomchilov: It's to keep the iteration from running off the end of the array. This used to give extra `nil`s at the end. See https://github.com/ruby/debug/pull/1001/commits/68374596f78c9d32141427d6aec5462a9d118b88

I can add a comment.

Could you please run the CI?

ono-max: Done.

ono-max:
It seems to be that some protocol tests are still failed. Can you please check them? If you need help, please let me know 👍

amomchilov:
@ono-max Hmmm I think we have one of those "it works on my machine" situations

Say the line bart... sigh... it works on my machine.

```
$ bundle exec rake test_protocol

Loaded suite /Users/alex/.gem/ruby/3.3.0/gems/rake-13.0.6/lib/rake/rake_test_loader
Started
Finished in 1.438602 seconds.
2 tests, 690 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
1.39 tests/s, 479.63 assertions/s
```

I'm running on Ruby 3.3.0 on macOS Sonoma 14.2.1.

I suspect these errors are coming from the ancestors chains being in different in CI, but I don't know why that would be. Could you please help me investigate this? I don't have access to a Linux machine to compare against.

ono-max:
I investigated the cause of the failed tests. I believe that the reason is because ancestor classes are changed from Ruby 3.1. Thus, we can skip the test if the Ruby version is lower than 3.1, e.g. [6f154ac](https://github.com/ruby/debug/pull/1059/commits/6f154ac0583d8d7722394c58c6a0ba66b806dee0)

amomchilov:
@ono-max Fantastic find! I pulled that commit into this PR, so you get credit for it :)

Is this ready to merge?

ono-max (approved):

ono-max:
I appreciate your dedicated efforts and hard work. @ko1-san will merge this PR 👍

amomchilov:
@ono-max Thanks!

Could you run the CI one more time, so we're sure it's green before merging?

amomchilov:
@ono-max Aw man, there's still more issues. Are these flakey tests? I've ran `bundle exec rake test_protocol` with Ruby 3.1.4 and 3.2.2, and get all tests passing.

amomchilov:
@ko1 Can you please review and merge this?

amomchilov:
@ko1 Are these flaky tests? I don't understand what went wrong, but it doesn't seem related to these changes

PR #1003, opened 2023-07-27
Double colon in Hash keys (`:sym:` → `sym:`)

amomchilov:
This PR depends on the changes in #1001. Only the last commit of this PR is unique to this PR.

* Both Chrome and VSCode automatically insert a `:` after the key, so the final result looks closer Ruby's Hash syntax, like: `{ sym: 1 }`
* This is consistent with how both Chrome and VSCode render the keys of JavaScript objects

Here's an example for this hash:

```ruby
h = {
  sym1: 1,
  sym2: 2,
  "str" => 3,
}
```

<table>
  <thead>
    <tr>
      <th/>
      <th>Before</th>
      <th>After</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>Chrome</td>
      <td>
        chrome before
      </td>
      <td>
        chrome after
      </td>
    <tr>
    <tr>
      <td>VSCode</td>
      <td>
        vscode before
      </td>
      <td>
        vscode after
      </td>
    <tr>
  </tbody>
</table>


st0012:
Are VS Code's pictures reversed?

amomchilov:
@st0012 Yes they are 😅 

Fixed :)

ono-max (approved):

PR #1004, opened 2023-07-28
Omit `#class` for simple values

amomchilov:
Thanks for your Pull Request 🎉 

**Please follow these instructions to help us review it more efficiently:**

- Add references of related issues/PRs in the description if available.
- If you're updating the readme file, make sure you followed [the instruction here](https://github.com/ruby/debug/blob/master/CONTRIBUTING.md#to-update-readme).

## Description
Describe your changes:

PR #1005, opened 2023-07-28
Debug representation hook

amomchilov:
This PR depends on the changes in #1004. Only [this last commit](https://github.com/ruby/debug/pull/1003/979a16497d27e2f4021ae84f094768627e1) of this PR is unique to this change.

## Description

I propose we add a hook method that developers can implement in their classes to customize the structural representation of their objects in a debugger. Used judiciously, this has the potential to really improve the clarify of certain types of objects.

I found it particularly useful for container-like objects, like `OpenStruct` and `ActionController::StrongParameters`.

| Before | After |
|--------|-------|
| Screenshot 2023-07-28 at 10 12 48 AM | Screenshot 2023-07-28 at 10 17 11 AM |

### DIFF
--- a/test/debug/variable_inspector_test.rb
+++ b/test/debug/variable_inspector_test.rb
@@ -47,7 +47,6 @@ module DEBUGGER__
       )
 
       expected = [
-        Variable.internal(name: '#class', value: Hash),
         Variable.new(name: ':sym', value: "has Symbol key"),
         Variable.new(name: '"str"', value: "has String key"),
         Variable.new(name: '1', value: "has Integer key"),
@@ -73,7 +72,6 @@ module DEBUGGER__
 
     def test_named_members_of_string
       expected = [
-        Variable.internal(name: '#class', value: String),
         Variable.internal(name: '#length', value: 5),
         Variable.internal(name: '#encoding', value: Encoding::UTF_8),
         # skip #dump member for short strings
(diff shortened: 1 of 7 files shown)


test/debug/variable_inspector_test.rb
```diff
@@ -227,5 +252,15 @@ def initialize(x:, y:)
         @y = y
       end
     end
+
+    class ClassWithCustomDebugRepresentation
+      def initialize(debug_representation)
+        @debug_representation = debug_representation
+      end
+
+      def debug_representation
```
amomchilov, line 261: Perhaps it should take unused rest args, in case any need to be added in the future?

```suggestion
      def debug_representation(*, **)
```

jabamaus:
A big +1 for this feature. This would be great. Surprised this wasn't in from the start.  Being able to control what's on the debug inspector without overloading to_s and inspect (which may be needed for something else) seems like a very sensible idea. 

When are some of these PRs going to get applied? It's been quiet the last few months...

jabamaus:
I'm not sure what the content of this PR exactly is but what I'd simply like is the ability to define a method on my objects called something like "to_rdbg" (or something I don't really mind) and have rdbg call that preferentially to to_s/inspect. In this way I could have to_s/inspect freed up for just my use and control over what appears in the debug window.

jabamaus:
Thanks @amomchilov that makes it easier to follow. What will happen if the object being inspected is a BasicObect? BasicObject does not implement respond_to?

amomchilov:
@jabamaus Please have a look at the implementation. It [always uses the `Kernel` implementation of `respond_to?`](https://github.com/ruby/debug/commit/89427979a16497d27e2f4021ae84f094768627e1#diff-6b56f95d5462f9655b8697972b970f713681fd490f9364ccd20590ed2f1d81d7R21), so that it works even on `BasicObject`, or any objects with a bad override of `#respond_to?`.

ko1:
> @ko1 The API is really simple: If your object responds to #debug_representation, the debugger will call it and use the result as the representation of your object.

Could you explain the specification of `#debug_representation`?

PR #1008, opened 2023-08-02
Disable cloned breakpoints trace point events

vinistock:
## Description

We're currently not invoking `bp.disable` or `bp.delete` on the cloned breakpoint. This means that the associated trace point event continues to be enabled.

Most of the time, that's not an issue because the target `@iseq` will not match. But there are some cases where old `@iseq` objects may continue to exist and we will continue having the execution suspended by the old trace point.

One example scenario of this is when using aliases, which hold references to the old `@iseq` (see the test, which reproduces the issue).

Depending on the order of events, this sometimes leads to execution being suspended by breakpoints that were already removed - because the trace point event is still enabled, even if the breakpoint was removed from the hash. Disabling the trace point for the removed breakpoint ensures that we can't get the execution suspended by these removed breakpoints.

### DIFF
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -1739,15 +1739,19 @@ module DEBUGGER__
       # check breakpoints
       if file_path
         @bps.find_all do |_key, bp|
-          LineBreakpoint === bp && bp.path_is?(file_path)
+          LineBreakpoint === bp && bp.path_is?(file_path) && (iseq.first_lineno..iseq.last_line).cover?(bp.line)
         end.each do |_key, bp|
           if !bp.iseq
             bp.try_activate iseq
           elsif reloaded
             @bps.delete bp.key # to allow duplicate
-            if nbp = LineBreakpoint.copy(bp, iseq)
-              add_bp nbp
-            end
+
+            # When we delete a breakpoint from the @bps hash, we also need to deactivate it or else its tracepoint event
+            # will continue to be enabled and we'll suspend on ghost breakpoints
+            bp.delete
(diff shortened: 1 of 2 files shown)


lib/debug/session.rb
```diff
@@ -1739,15 +1739,19 @@ def on_load iseq, src
       # check breakpoints
       if file_path
         @bps.find_all do |_key, bp|
-          LineBreakpoint === bp && bp.path_is?(file_path)
+          LineBreakpoint === bp && bp.path_is?(file_path) && (iseq.first_lineno..iseq.last_line).cover?(bp.line)
```
vinistock, line 1742: Here, we don't need to check breakpoints that are not covered by the iseq.

lib/debug/session.rb
```diff
@@ -1739,15 +1739,19 @@ def on_load iseq, src
...
-            if nbp = LineBreakpoint.copy(bp, iseq)
-              add_bp nbp
-            end
+
+            # When we delete a breakpoint from the @bps hash, we also need to deactivate it or else its tracepoint event
+            # will continue to be enabled and we'll suspend on ghost breakpoints
+            bp.delete
+
+            nbp = LineBreakpoint.copy(bp, iseq)
+            add_bp nbp
```
vinistock, line 1754: `LineBreakpoint.copy` always returns a breakpoint, so the `if` here is not necessary.

ko1:
I feel this kind of duplication check will introduce another issue (e.g. ISeq search order and so on) but try it.

PR #1017, opened 2023-09-13
Allow adding a suffix to sockets for easier multiple-socket debugging

andyjeffries:
- Add references of related issues/PRs in the description if available.

**No issue raised**

- If you're updating the readme file, make sure you followed [the instruction here](https://github.com/ruby/debug/blob/master/CONTRIBUTING.md#to-update-readme).

**Done**

## Description

In one of my projects (actually most, but this one in particular is what I'm working on it for) I use a Procfile with Overmind to launch both a web process and a worker process. However, at the moment I can't tell which socket is for which process.

This PR means that I can use a Procfile in the following format:

```
web: RUBY_DEBUG_OPEN=true RUBY_DEBUG_SOCK_SUFFIX=myproj-web PORT=3000 bundle exec puma -C config/puma.rb
worker: RUBY_DEBUG_OPEN=true RUBY_DEBUG_SOCK_SUFFIX=myproj-worker bundle exec sidekiq -t 10 -C config/sidekiq.yml
```

And the output of `bundle exec rdbg --util=list-socks` (which is used by VS Code for example) will change from:

```
/var/folders/gf/3z_681416nj3_zxgn2njdybm0000gn/T/ruby-debug-sock-501/ruby-debug-andy-32986
/var/folders/gf/3z_681416nj3_zxgn2njdybm0000gn/T/ruby-debug-sock-501/ruby-debug-andy-32998
```

to:

```
/var/folders/gf/3z_681416nj3_zxgn2njdybm0000gn/T/ruby-debug-sock-501/ruby-debug-andy-myproj-web-32986
/var/folders/gf/3z_681416nj3_zxgn2njdybm0000gn/T/ruby-debug-sock-501/ruby-debug-andy-myproj-worker-32998
```

This makes it much more obvious which socket is for which process. I'm going to raise a PR in the extension's project over there too (to trim out the matching stuff at the start), but this will be vital in making this use case work.

### DIFF
--- a/lib/debug/config.rb
+++ b/lib/debug/config.rb
@@ -47,6 +47,7 @@ module DEBUGGER__
     sock_path:      ['RUBY_DEBUG_SOCK_PATH',    "REMOTE: UNIX Domain Socket remote debugging: socket path"],
     sock_dir:       ['RUBY_DEBUG_SOCK_DIR',     "REMOTE: UNIX Domain Socket remote debugging: socket directory"],
     sock_suffix:    ['RUBY_DEBUG_SOCK_SUFFIX',  "REMOTE: UNIX Domain Socket remote debugging: socket suffix"],
+    sock_prefix:    ['RUBY_DEBUG_SOCK_PREFIX',  "REMOTE: UNIX Domain Socket remote debugging: socket prefix"],
     local_fs_map:   ['RUBY_DEBUG_LOCAL_FS_MAP', "REMOTE: Specify local fs map", :path_map],
     skip_bp:        ['RUBY_DEBUG_SKIP_BP',      "REMOTE: Skip breakpoints if no clients are attached", :bool, 'false'],
     cookie:         ['RUBY_DEBUG_COOKIE',       "REMOTE: Cookie for negotiation"],
@@ -498,6 +499,9 @@ module DEBUGGER__
   def self.create_unix_domain_socket_name_prefix(base_dir = unix_domain_socket_dir)
     user = ENV['USER'] || 'UnknownUser'
     filename = "ruby-debug-#{user}"
+    if !CONFIG[:sock_prefix].nil?
+      filename = "#{CONFIG[:sock_prefix]}-#{filename}"
+    end
     if !CONFIG[:sock_suffix].nil?
       filename += "-#{CONFIG[:sock_suffix]}"
     end
--- a/lib/debug/config.rb
+++ b/lib/debug/config.rb
@@ -44,10 +44,9 @@ module DEBUGGER__
     open:           ['RUBY_DEBUG_OPEN',         "REMOTE: Open remote port (same as `rdbg --open` option)"],
     port:           ['RUBY_DEBUG_PORT',         "REMOTE: TCP/IP remote debugging: port"],
     host:           ['RUBY_DEBUG_HOST',         "REMOTE: TCP/IP remote debugging: host", :string, "127.0.0.1"],
+    session_name:   ['RUBY_DEBUG_SESSION_NAME'  "REMOTE: Session name for differentiating multiple sessions"],
     sock_path:      ['RUBY_DEBUG_SOCK_PATH',    "REMOTE: UNIX Domain Socket remote debugging: socket path"],
     sock_dir:       ['RUBY_DEBUG_SOCK_DIR',     "REMOTE: UNIX Domain Socket remote debugging: socket directory"],
-    sock_suffix:    ['RUBY_DEBUG_SOCK_SUFFIX',  "REMOTE: UNIX Domain Socket remote debugging: socket suffix"],
-    sock_prefix:    ['RUBY_DEBUG_SOCK_PREFIX',  "REMOTE: UNIX Domain Socket remote debugging: socket prefix"],
     local_fs_map:   ['RUBY_DEBUG_LOCAL_FS_MAP', "REMOTE: Specify local fs map", :path_map],
     skip_bp:        ['RUBY_DEBUG_SKIP_BP',      "REMOTE: Skip breakpoints if no clients are attached", :bool, 'false'],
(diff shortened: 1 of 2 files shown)


ko1:
simple question: We can also use "prefix".
Suffix is more preferable? Do you know similar options for other tools?

andyjeffries:
No reason to not have both prefix and suffix. I'll update the PR with that.

andyjeffries:
I don't know of options for any other tools (I only use Ruby and don't know of any other similar facilities for Ruby).

I've added prefix now too though.

andyjeffries:
@ko1 I think this needs your approval again to run. Thanks.

ko1:
I don't ask to add PREFIX too. I want to know PREFIX or SUFFIX which is preferable. I think both is too much. To determine this kind of decision, checking other tools is good idea in general.

----

I understand you want to make clear that which UNIX domain socket is what you want to attach. So the main reason is "naming".

Now, debug.gem support to return information about UNIX domain socket with "info" request to debug server: https://github.com/ruby/debug/blob/master/lib/debug/server.rb#L132C11-L132C11
And if vscode-extension can support this information, we can see more rich information for each UNIX domain socket (connections).

But supporting on vscode-extension needs some more time.
So how about to add the "connection name" functionality?

My idea is:

* (1) Add `RUBY_DEBUG_NAME=<name>` or `RUBY_DEBUG_SESSION_NAME=<name>`
  * `--name=<name>` or `--session-name=<name>` with command line parameter.
* (2) The `<name>` will be returned by `info` request.
* (3) The `<name>` will be used as PREFIX (or SUFFIX. we don't need to decide as a spec) on UNIX domain socket.
  * `<name>` needs some translation because `<name>` can contain unacceptable characters for file systems like `/`. Or restrict `<name>` with Ruby's variable names (lower chars, digits and `_`) for example.
* (4) Option: REPL prompt can contain this `<name>`.

If vscode extension supports (2), we can see richer one.
With (3) we can see the name.
We don't need to specify PREFIX or SUFFIX any more.

andyjeffries:
I'm not clear, if we're doing (2) then why do we need (3)? I can then update the VS Code extension to prefix each line in the picker (based on my existing PR on that project) to list them in the format "name: .../filename". That does seem nicer, so I'll rework this PR to be that.

andyjeffries:
Having just looked at it, there's no way I understand the codebase enough to do this...

If you could give me any tips on a) where to add to the info command and b) how to test this locally without using VS Code as the client, I'd appreciate it (and could then continue).

Failing that, me and my team will just have to run with the forked gem and locally installed VS Code extension as it gets us moving nicely for now...

ko1:
> I'm not clear, if we're doing (2) then why do we need (3)?

We don't need to wait vscode extension update.

ko1:
> If you could give me any tips on a) where to add to the info command and b) how to test this locally without using VS Code as the client, I'd appreciate it (and could then continue).

On this PR, you only need to rename `PREFIX/SUFFIX` to `SESSION_NAME` (and maybe prefix will be nice).

andyjeffries:
Are you waiting for me to raise a separate PR to implement the changes to the info call, or are you going to do that? Thanks.

lib/debug/config.rb
```diff
@@ -496,7 +498,14 @@ def self.unix_domain_socket_dir
...
   def self.create_unix_domain_socket_name_prefix(base_dir = unix_domain_socket_dir)
     user = ENV['USER'] || 'UnknownUser'
-    File.join(base_dir, "ruby-debug-#{user}")
+    filename = "ruby-debug-#{user}"
+    if !CONFIG[:sock_prefix].nil?
+      filename = "#{CONFIG[:sock_prefix]}-#{filename}"
+    end
+    if !CONFIG[:sock_suffix].nil?
+      filename += "-#{CONFIG[:sock_suffix]}"
+    end
```
eugeneius, line 507: > On this PR, you only need to rename `PREFIX/SUFFIX` to `SESSION_NAME` (and maybe prefix will be nice).

I think what Koichi meant here was:

```suggestion
    if !CONFIG[:session_name].nil?
      filename = "#{CONFIG[:session_name]}-#{filename}"
    end
```

lib/debug/config.rb
```diff
@@ -46,6 +46,8 @@ module DEBUGGER__
     host:           ['RUBY_DEBUG_HOST',         "REMOTE: TCP/IP remote debugging: host", :string, "127.0.0.1"],
     sock_path:      ['RUBY_DEBUG_SOCK_PATH',    "REMOTE: UNIX Domain Socket remote debugging: socket path"],
     sock_dir:       ['RUBY_DEBUG_SOCK_DIR',     "REMOTE: UNIX Domain Socket remote debugging: socket directory"],
+    sock_prefix:    ['RUBY_DEBUG_SOCK_PREFIX',  "REMOTE: UNIX Domain Socket remote debugging: socket prefix"],
```
eugeneius, line 49:
```suggestion
```

andyjeffries:
I've got to be honest @eugeneius and @ko1 , I really don't understand what I'm doing here then. I've removed the changes from README.md (I believe they're now pulled in by README.md.erb), and removed the suffix/prefix stuff, so we just have a new "session name" setting, but is that really it?!

All this PR does is add a single line to a new config option? 

Feels like that could have been done as part of the PR (whoever is doing it, I don't mind doing it - if someone can point me roughly where to start looking) to add it to the info command.

eugeneius:
I was trying to suggest removing the changes related to `sock_prefix`/`sock_suffix`, and adding the `session_name` setting to the socket filename as a prefix when present. You can also keep the documentation changes in the README, but you have to apply them to `misc/README.md.erb` and then run `rake` to regenerate `README.md`, rather than edit it directly.

ko1:
Thank you @andyjeffries and @eugeneius to discuss. I make another PR to merge this feature.
Please check the PR if it is enough for your purpose.
And sorry @andyjeffries I couldn't guide your contribution well.

PR #993, opened 2023-06-18
Fix ruby warnings

y-yagi:
## Description

This fixes the following Ruby warnings:

```
lib/ruby/gems/3.2.0/gems/debug-1.8.0/lib/debug/session.rb:2527: warning: assigned but unused variable - sym
lib/ruby/gems/3.2.0/gems/debug-1.8.0/lib/debug/thread_client.rb:618: warning: assigned but unused variable - e
```

PR #1018, opened 2023-09-13
Removed unused variables

hsbt:
## Description

If we invoked `rails new railsapp` with `RUBYOPT=-w`, we will get the following warnings.

```
/path/to/debug-1.8.0/lib/debug/session.rb:2527: warning: assigned but unused variable - sym
/path/to/debug-1.8.0/lib/debug/thread_client.rb:618: warning: assigned but unused variable - e
```

I removed them.


st0012:
I think this is duplicated with #993

hsbt:
👍 Thanks. I'll close this.

PR #1019, opened 2023-09-13
Fix warnings for unused variables

andyw8:
Fixes the following warnings:

```
.../3.0.0/gems/debug-1.8.0/lib/debug/session.rb:2527: warning: assigned but unused variable - sym
.../3.0.0/gems/debug-1.8.0/lib/debug/thread_client.rb:618: warning: assigned but unused variable - e
```


st0012:
This is duplicated with #993 and #1018

PR #1025, opened 2023-10-12
Fix DAP's completion request with new IRB's completor

st0012:
In IRB 1.8.2, the completor's design and API has been changed and thus would break DAP server's completion request. This commit fixes the issue by introducing a `Completor` class that abstracts away the differences between the old and new completor.

cc @tompng

### DIFF
--- a/lib/debug/completor.rb
+++ b/lib/debug/completor.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+require "irb/completion"
+
+module DEBUGGER__
+  class Completor
+    class << self
+      # old IRB completion API
+      if defined?(IRB::InputCompletor)
+        def retrieve_completion_data(input, binding)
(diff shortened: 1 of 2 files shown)


st0012:
With https://github.com/ruby/irb/pull/730, the request should not be broken anymore once it's released. And while it's still better not to use the deprecated `InputCompletor`, I think we can wait a while to see how the new API will evolve before updating it.

ko1:
I'm not sure but also LSP may need completion feature so different library is another idea for completion?

st0012:
You're right, LSP also needs completion feature and we're building one in `ruby-lsp` with a [inhouse indexer](https://github.com/Shopify/ruby-lsp/tree/main/lib/ruby_indexer), which we hope to extract into a gem when it's feature-completed.

But unlike a LSP server, IRB and `debug` both have access to binding objects and runtime information, which means it can get great completion result without going through the expensive indexing process. So I think it's still worth exploring a different completor for IRB and `debug`.

PR #1030, opened 2023-10-26
Unfreeze threads for some object-evaluating commands

st0012:
There are several commands that evaluate objects and call methods on them, such as:
- info
- ls (outline)
- trace object
- display

If the called method acquires a mutex shared with another thread (e.g. when using Timeout), then it'd cause a deadlock as all threads are stopped.

For example, if there's an ActiveRecord Relation object stored as a local variable, and the user runs the `info` command, then it'd call `inspect` on it and trigger a database query, it could cause a deadlock as described in #877.

This commit fixes the issue by unfreezing all threads before evaluating those commands and freezing them again after receiving the result event.

Note: theoretically there are more commands that evaluate user program's objects, like `break Foo#bar` calls `Foo.instance_method`. But in that case the problem would only arise when `Foo.instance_method` is overridden with something that'd acquire the lock, which is a far unlikely to happen.

### DIFF
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -256,6 +256,11 @@ module DEBUGGER__
       @tc << req
     end
 
+    def request_tc_with_freed_threads(req)
+      restart_all_threads
+      request_tc(req)
+    end
+
     def process_event evt
       # variable `@internal_info` is only used for test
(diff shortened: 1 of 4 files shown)


ko1:
I'm not sure it is safe because current thread management technique does not have enough quality to control them, but try it.

PR #1039, opened 2023-11-28
Compile iseq_collector.c logic only on CRuby

eregon:
## Description

Compile iseq_collector.c logic only on CRuby

* iseq_collector.c relies on many CRuby internals, there is no point to compile it on other Rubies.
* It now fails to compile on TruffleRuby:
```
  ../../../../ext/debug/iseq_collector.c: In function ‘imemo_type’:
  ../../../../ext/debug/iseq_collector.c:18:13: error: implicit declaration of function ‘RBASIC’ [-Werror=implicit-function-declaration]
    18 |     return (RBASIC(imemo)->flags >> FL_USHIFT) & imemo_mask;
        |            ^~~~~~
```
See for example this GH action run: https://github.com/ruby/irb/actions/runs/7017641177/job/19091354167?pr=790#step:3:87

I also added a workflow to ensure the extension keeps building fine on TruffleRuby, and cleaned up a bit existing workflows.

### DIFF
--- a/ext/debug/extconf.rb
+++ b/ext/debug/extconf.rb
@@ -4,6 +4,7 @@ File.write("debug_version.h", "#define RUBY_DEBUG_VERSION \"#{DEBUGGER__::VERSIO
 $distcleanfiles << "debug_version.h"
 
 if defined? RubyVM
+  $defs << '-DHAVE_RB_ISEQ'
   $defs << '-DHAVE_RB_ISEQ_PARAMETERS'
   $defs << '-DHAVE_RB_ISEQ_CODE_LOCATION'
 
--- a/.github/workflows/protocol.yml
+++ b/.github/workflows/protocol.yml
@@ -1,10 +1,3 @@
-# This workflow uses actions that are not certified by GitHub.
-# They are provided by a third-party and are governed by
-# separate terms of service, privacy policy, and support
-# documentation.
-# This workflow will download a prebuilt Ruby version, install dependencies and run tests with Rake
-# For more information see: https://github.com/marketplace/actions/setup-ruby-jruby-and-truffleruby
-
 name: Protocol
 
 on:
@@ -26,12 +19,10 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: Set up Ruby
-    # To automatically get bug fixes and new Ruby versions for ruby/setup-ruby,
-    # change this to (see https://github.com/ruby/setup-ruby#versioning):
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: ${{ matrix.ruby-version }}
-        bundler-cache: true # runs 'bundle install' and caches installed gems automatically
+        bundler-cache: true
     - name: Run tests
       run: |
         bundle exec rake clobber
--- a/.github/workflows/ruby-macos.yaml
+++ b/.github/workflows/ruby-macos.yaml
@@ -1,10 +1,3 @@
-# This workflow uses actions that are not certified by GitHub.
-# They are provided by a third-party and are governed by
-# separate terms of service, privacy policy, and support
-# documentation.
-# This workflow will download a prebuilt Ruby version, install dependencies and run tests with Rake
-# For more information see: https://github.com/marketplace/actions/setup-ruby-jruby-and-truffleruby
-
 name: Ruby(macOS)
 
 on:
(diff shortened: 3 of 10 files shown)


eregon:
Unfortunately there is a rubygems.org issue currently which means stringio 3.1.0 might not be picked up in CI: https://github.com/rubygems/rubygems.org/issues/4247

eregon:
The Protocol workflow seems to randomly fail, already before: https://github.com/ruby/debug/actions/workflows/protocol.yml

eregon:
> Unfortunately there is a rubygems.org issue currently which means stringio 3.1.0 might not be picked up in CI: [rubygems/rubygems.org#4247](https://github.com/rubygems/rubygems.org/issues/4247)

That's fixed now, and the added truffleruby workflow passes.

ext/debug/extconf.rb
```diff
@@ -4,6 +4,7 @@
 $distcleanfiles << "debug_version.h"
 
 if defined? RubyVM
+  $defs << '-DCRUBY'
```
ko1, line 7: No C macros to determine CRuby or others?

eregon: I'm not aware of one.
We could be more specific and name it `HAVE_RB_ISEQ` or so maybe because it's all about iseq, but that may be confusing given we don't use `have_*` to check that.
Is `rb_iseq` exposed in CRuby headers?

ko1: `RUBY_RUBY_H` for example? (`include/ruby/ruby.h`)

ko1:
BTW now debug.gem doesn't support TruffleRuby and why should we merge it?

ext/debug/extconf.rb
eregon: TruffleRuby also has ruby/ruby.h and reuses CRuby headers: https://github.com/oracle/truffleruby/tree/master/lib/cext/include, so that wouldn't be a good term for it:

eregon:
> BTW now debug.gem doesn't support TruffleRuby and why should we merge it?

Because it breaks irb's CI: https://github.com/ruby/irb/actions/runs/7017641177/job/19091354167?pr=790#step:3:87

eregon:
And also it's important that Gemfile can just install the `debug` gem on TruffleRuby too.

eregon:
@ko1 Let me give you some context.

* `gem install debug` works on TruffleRuby 23.1.1
* But it does not work on truffleruby master, because truffleruby no longer (and basically cannot) support `RBASIC(obj)->flags` since it changed its C extension implementation: https://github.com/oracle/truffleruby/issues/3118#issuecomment-1804194112.
* C extension should not use those flags anyway: https://bugs.ruby-lang.org/issues/18059
* https://github.com/ruby/debug/blob/master/ext/debug/iseq_collector.c is clearly CRuby-specific, it seems good to guard it anyway. It would probably segfault or other bad error on non-CRuby (if it compiled fine).
* https://github.com/ruby/debug/blob/master/ext/debug/debug.c OTOH is more portable and in fact TruffleRuby implements `rb_debug_inspector_open()`.
* It's important the debug gem builds and gem-installs fine on TruffleRuby because it might be part of a Gemfile, or a gem dependency and also `debug` is a bundled gem since 3.1, so it is also built and shipped as part of truffleruby.
* `require 'debug'` fails (and on 23.1.1 too), that's fine for now, the debug gem does not support truffleruby yet. And in fact the next thing that breaks is TruffleRuby does not implement :call TracePoint yet.

ext/debug/extconf.rb
eregon: I changed it to `HAVE_RB_ISEQ`

ko1: Another `RB_ISEQ_...` means Ruby has corresponding functions, so `HAVE_RB_ISEQ` means having `rb_iseq()`.
So... anyway, I'll change it later.

ko1:
Ok I'll merge it.

Note that, this is repeating, I don't provide TruffleRuby support yet and I'll remove .github/workflows/truffleruby.yml if new feature breaks the test and I don't have any idea to fix in future with informing you.

eregon:
Thank you. Yes, that's fine.
I'm happy to help if you @-me on GitHub if there is any issue, but yes feel free to disable (e.g. `if: false`) or remove that workflow if not.

PR #1053, opened 2023-12-30
Update README.md

glaucocustodio:
- add example on how to set up a custom alias on IRB console

## Description
Although setting up alias on IRB might not seem the scope of this gem, the README of the debug gem was the place where myself (and possibly other devs) referred to improve my experience using this gem.


ko1:
I think it should be more useful to link to irb reference instead of writing only about alias'ing.

Issue #1031, opened 2023-10-30
Respect XDG basedir spec


jasonkarns:
**Your proposal**
I propose that debug respect the XDG Base Directory Specification such that files are written to locations controlled by user environment variables.

There are at least three relevant files:
- rdbg_history
- rdbgrc
- unix_domain_socket_dir

Presently, the history file may be _forced_ into a different location with the `RUBY_DEBUG_HISTORY_FILE` env var. While helpful, this does not adhere to the desire for single-configuration (setting XDG_* vars once and not needing to configure each of hundreds of different utilities). It should default to a file or directory in `XDG_STATE_HOME` (which itself defaults to `$HOME/.local/state`)

The unix_domain_socket_dir is already very nearly adhering to Xdg, as its first implicit location (after checking explicit config setting) uses `XDG_RUNTIME_DIR`.

The rdbgrc (and .rb variant) have no env var equivalents and so cannot even be set explicitly. However, they should be using files within `XDG_CONFIG_HOME` (which defaults to `$HOME/.config`). Perhaps `$XDG_CONFIG_HOME/rdbg/config` or `$XDG_CONFIG_HOME/rdbg/config.rb` ?

**Additional context**
The Xdg basedir spec is here: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

**Rationale**

- keeps user's homedirectory clean
- allows versioning of configuration files cleanly (ie, dotfiles repo to version-control .config but not cache or data files that also typically end up in homedir)
- allows single-configuration changes to move configuration to different directory tree (for testing, for simulated environments, and many other unknowable use-cases)

ko1:
We need to make migration plan.
`~/.rdbgrc` (and so on) are already there so we need to respect current files.

On the https://github.com/vim/vim/issues/4275, it was closed because of duplication of https://github.com/vim/vim/issues/2034 which has huge comments and not accepted (I can't read all of comments) but `#4275` proposes:

```
~/.vimrc
~/.vim/vimrc
$XDG_CONFIG_DIR/vim/vimrc (if the env var is unset, its default is ~/.config)
```

For rdbgrc, making searching order as

```
~/.rdbgrc
$XDG_RUNTIME_DIR/rdbg/config if `XDG_RUNTIME_DIR` is available
```

is one idea.

For `~/.rdbg_histroy` there are two options:

1. search `~/.rdbg_history` and `$XDG_STATE_HOME/rdbg/history` if `$XDG_STATE_HOME` is available
[...]

PR #1055, opened 2023-12-31
Support XDG directories for config and history files

jgarber623:
## Description

Addresses #1031 by implementing support for XDG directories in accordance with the rules discussed in the issue and summarized in [my comment here](https://github.com/ruby/debug/issues/1031#issuecomment-1855128422).

This change should be non-breaking in that the code continues to prefer existing default configuration files at `~/.rdbgrc` and `~/.rdbg_history`.

If those files _do not_ exist _and_ XDG environment variables are configured, then configuration and history files in the relevant XDG directories are used.

- `$XDG_CONFIG_HOME/rdbg/config` (if file is present)
- `$XDG_DATA_HOME/rdbg/history`

Note that `$XDG_CONFIG_HOME` is _typically_ `$HOME/.config` and that `$XDG_DATA_HOME` is _typically_ `$HOME/.local/share`. These are customizable by users, but I believe those are the customary/default paths.


lib/debug/session.rb
```diff
@@ -2293,11 +2293,19 @@ def skip?
...
-     [CONFIG[:init_script], false],
-     ].each{|(path, rc)|
+    rc_file_paths = [
+      [File.expand_path('~/.rdbgrc'), true],
+      [File.expand_path('~/.rdbgrc.rb'), true],
+      # ['./.rdbgrc', true], # disable because of security concern
+    ]
+
+    if (xdg_home = ENV["XDG_CONFIG_HOME"])
+      rc_file_paths << [File.expand_path(File.join(xdg_home, "rdbg", "config")), true]
```
ko1, line 2303: don't you need `config.rb` here?

jgarber623: Good catch. This change should support both `${XDG_CONFIG_HOME}/rdbg/config` and `${XDG_CONFIG_HOME}/rdbg/config.rb`. I'll make that update.

jgarber623: @ko1 Thank you again for the feedback. I've added support for `config.rb` in this path in 55f6624.

ko1: Thank you!

jgarber623:
Hello! I don't have the ability to re-run the failed checks and it's not immediately clear to me why those failed (and why Ruby 3.2 and 3.3 passed).

For someone more knowledgeable, do those failures look to be caused by changes in this pull request? Should I update this branch with a merge commit and have someone re-run the test suite?

ko1:
Sorry for keeping you waiting for a year. I'll merge it now.

One concern is existing user doesn't have chance to ignore the `~/.rdbg_history`, so show warning is one idea.
Moving (copying?) it when XDG directory doesn't have a history file yet, is one idea.
But it should be separate PR.

ko1:
ah, mentioned here. I have no strong opinion, but not sure which is the best.
do you have reasons to choose DATA?

jgarber623:
> do you have reasons to choose DATA?

@ko1 The conversation on #1031 documents how we got to using `XDG_DATA_HOME` for the history file. See: https://github.com/ruby/debug/issues/1031#issuecomment-1823767168.

> I _think_ `$XDG_DATA_HOME` (e.g. `~/.local/share`) is the more appropriate place for history files. Similar to my note above, the XDG page on the Arch wiki cites a number of examples using `$XDG_DATA_HOME` for history files.

It doesn't look like there were any concerns or objections to that approach which is likely why I implemented the change that way.

ko1:
okay, I'll move discussion to #1031

PR #1056, opened 2023-12-31
Correct the irb command's description

st0012:
After #1024, the `irb` command does more than starting a pure IRB session.


ko1:
Thank you!

PR #1057, opened 2024-01-02
Make irb_console toggleable with config update

st0012:
1. When the user sets `irb_console` to `true` through the command, the `irb:rdbg` will now be enabled the same way as typing the `irb` command.
2. When the user sets `irb_console` to `false` through the command, the `irb:rdbg` will now be disabled.
3. Users can now enable `irb:rdbg` by setting `irb_console` to `true` in their `.rdbgrc` file.


jhirn:
Have a question not worth opening up a new issue for. When running `bin/rspec ...` or `bin/rails ...` it does not seem to automatically pickup the `.rdbgrc` config when I just use `binding.break` in a file. Is there something else I have to do in my project to ensure debug gem picks up this config when setting a break point?

PR #1060, opened 2024-01-07
Fix one of flakey test in protocol tests

ono-max:


lib/debug/server_cdp.rb
```diff
@@ -489,7 +489,12 @@ def process
                         id: SecureRandom.hex
         when 'Runtime.terminateExecution'
           send_response req
-          exit
+          if SESSION.in_subsession?
+            @q_msg << 'kill!'
+          else
+            @q_msg << 'kill!'
+            pause
+          end
```
st0012, line 497: Feels like it could be simplified a bit?

```suggestion
          @q_msg << 'kill!'
          pause unless SESSION.in_subsession?
```

st0012:
Thanks for the fix btw!

PR #1065, opened 2024-01-19
Use ruby/actions workflow for ruby versions

m-nakamura145:
This PR use ruby/actions workflow for ruby versions

### DIFF
--- a/.github/workflows/protocol.yml
+++ b/.github/workflows/protocol.yml
@@ -7,14 +7,21 @@ on:
     branches: [ master ]
 
 jobs:
-  test:
+  ruby-versions:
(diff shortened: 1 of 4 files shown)


ko1:
Could you tell me the motivation of this PR?

m-nakamura145:
@ko1 Thank you. I believe there is a benefit in eliminating the need for someone to manually add the version each time a new Ruby version is released. Considering that other gems are using the same workflow, I thought it might be better to unify this approach. ref: https://github.com/ruby/irb/pull/516

ko1:
ok thanks, let's try.

PR #1071, opened 2024-01-29
Drop base64 dependency

Earlopain:
## Description
#1066 added `base64` as a dependency to silence warnings under Ruby 3.3
The digest gem already comes with its own base64 method. Just use that and drop the base64 gem dependency.

These methods are equivalent:
https://github.com/ruby/digest/blob/d39b684f9131186b994639be0e324e854eae4b1d/lib/digest.rb#L55-L57
https://github.com/ruby/base64/blob/9669a7d3b0e3b9a739969404daf58f912c58c6b3/lib/base64.rb#L273-L275


ko1:
ah, squash merge commit log becomes redirecting to this PR...
anyway thanks!

PR #1080, opened 2024-02-19
Support Ruby 3.4's new error message format

hsbt:
https://bugs.ruby-lang.org/issues/16495

This change is same as https://github.com/ruby/bigdecimal/pull/286


ko1:
I'll submit another PR to catch up Ruby 3.4.

PR #1092, opened 2024-03-20
Mention IRB console in readme

st0012:
I also took the liberty to format markdown tables in the readme (through my editor extension 😛 ), which will make them easier to read in the raw view.

For example:

Screenshot 2024-03-20 at 21 13 19


ko1:
Thank you. Could you consider to add about that remote debugging doesn't support it?

PR #1093, opened 2024-03-27
Use rb_iseqw_to_iseq to get iseq pointer

peterzhu2118:
Using DATA_PTR is internal implementation details so we should not use it.

It should fix this issue: https://github.com/ruby/ruby/pull/10336#issuecomment-2021861692


ko1:
Thanks!

PR #1094, opened 2024-04-04
Use default `max_lines` when calling `up` & `down`

HeyNonster:
## Description

Both `up` and `down` only show 1 source line when called, I think it would be more useful if they provided more context. This removes the `max_lines: 1` kwarg when they call `show_src`. `max_lines` in `show_src` will default to `CONFIG[:show_src_lines]`[^1].

[^1]:(https://github.com/ruby/debug/blob/0b77e8294b5b220b3b2a089bf09f94808654899a/lib/debug/thread_client.rb#L510)

Big thanks to @st0012 for showing me and @bquorning where this behavior is defined.

### DIFF
--- a/lib/debug/thread_client.rb
+++ b/lib/debug/thread_client.rb
@@ -1079,13 +1079,13 @@ module DEBUGGER__
           when :up
             if @current_frame_index + 1 < @target_frames.size
               @current_frame_index += 1
-              show_src max_lines: 1
+              show_src
               show_frame(@current_frame_index)
             end
           when :down
             if @current_frame_index > 0
               @current_frame_index -= 1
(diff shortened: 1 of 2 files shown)


HeyNonster:
Also probably useful to note that `byebug` displays more lines by default making it quite a difference switching from `byebug` to `debug`:

https://github.com/ruby/debug/assets/47484308/5f2174b9-1264-4753-be71-14203977c69f

HeyNonster:
Seems the protocol tests are flaky: https://github.com/ruby/debug/pull/1060

ko1:
Maybe we need to introduce another configuration.

PR #1097, opened 2024-04-11
Improve RUBYOPT's handling in tests

st0012:
When debugging the issue related to https://github.com/ruby/irb/pull/919, I noticed that debugger tests don't respect the IRB version I specified in the Gemfile. This is because console tests force override the RUBYOPT env, which will remove the `-rbundler/setup` injected by bundler.

Additionally, if tests use `run_rdbg` with the `rubyopt` option, the RUBYOPT will be overridden yet again.

So in this commit I did 2 improvements:

1. `run_rdbg` should append instead of override RUBYOPT
2. If tests are executed with bundler, we also run the debugger in PTY process with bundler by appending `-rbundler/setup` to RUBYOPT

### DIFF
--- a/test/support/console_test_case.rb
+++ b/test/support/console_test_case.rb
@@ -225,7 +225,12 @@ module DEBUGGER__
       @scenario = []
       test_steps.call
       @scenario.freeze
-      inject_lib_to_load_path
+
+      ENV['RUBYOPT'] = "-I #{__dir__}/../../lib"
+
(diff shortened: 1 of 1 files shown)


st0012:
Ruby head is failing because the commit it uses doesn't include the workaround in https://github.com/ruby/irb/pull/919 yet. Once it picks a newer commit it should pass.

ko1:
Thank you.

PR #1105, opened 2024-05-30
do not include `$FILENAME` in globals

nyz93:
`$FILENAME` is unsafe to include in the globals because it modifies ARGV and raises an error if it's a file that does not exist. This causes two issues if rdbg used to debug a command with arguments:
 * when using rdbg the server will crash when running `info globals`, as it does not catch the error
 * the DAP server catches this error, so when using a DAP client that requests globals before the debugged program was able to handle its arguments then the ARGV changes and the program will not do what's expected


ko1:
Thank you!

Issue #1117, opened 2024-09-18
Avoid port conflict in TCP/IP mode in processes that fork


crystianwendel:
Problem
--------

When using `rdbg` with the `--port` option, I encounter an issue when the process forks, leading to the error:
```log
Address already in use - bind(2) for IP:PORT (Errno::EADDRINUSE)
```
In my case, I'm using [solid_queue](https://github.com/samuelmullen/solid_queue), which in its basic configuration starts a Supervisor, Worker, and Dispatcher. This results in multiple processes, and since they all attempt to bind to the same specified port, a conflict occurs.

Feature Request
----------------

Change port option in rdbg so it allows specifying a range or list of ports. This would enable the debugger to iterate over the provided range, trying the next available port if one is already in use.

Proposed Solution
------------------

Allow rdbg to accept a port range or a list of ports, which it could iterate through to find an open one. For example:
```bash
rdbg --port 3003-3005 -n --open -c "./bin/jobs"
```
In this case, the UI_TcpServer class would attempt to bind to port 3003 first, and if that port is unavailable, it would continue trying 3004, 3005, and so on until it finds an open port.

This would greatly improve the usability of rdbg in environments where multiple processes may be spawned and need to communicate with the debugger.

Additional Context
------------------

This issue is particularly relevant in systems with forking processes, such as job queues (e.g., solid_queue), where multiple components (e.g., Supervisor, Worker, Dispatcher) are started. The ability to define a port range would help avoid conflicts and make debugging smoother.

f this enhancement is of interest to the maintainers, I would be happy to contribute by implementing it. Please let me know if this would be a welcome addition.

Thank you for considering this enhancement!

crystianwendel:
I found that the order in which `solid_queue` starts its processes is non-deterministic, so relying on a range of ports may not be a reliable solution. I wouldn't be able to predict which port my worker process would be running on.

An alternative approach could be to introduce a REGEX parameter, which would allow the TCP/IP connection to only start when the process information matches the given pattern. This way, the debugger could attach to the correct process based on more specific criteria, potentially avoiding the port conflict issue altogether.

gerymate:
Hopefully this PR would solve that.

marcelolx:
[...]

Issue #368, opened 2021-11-06
Remote debugging of `rake test` raises `Errno::EADDRINUSE`


suketa:
**Your environment**

* `ruby -v`: ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-linux]
* `rdbg -v`: rdbg 1.3.4

**Describe the bug**
remote debugging with `rake test` raises `Errno::EADDRINUSE`

**To Reproduce**
Rakefile:
```ruby
require 'rake/testtask'

Rake::TestTask.new(:test) do |t|
  t.test_files = ['test_example.rb']
end
```
test_example.rb:
```ruby
require 'minitest/autorun'

class TestExample < Minitest::Test
  def test_foo
    binding.break
    assert_equal(true, true)
  end
end
```

And run in console:
```
$ rdbg -O -p 12345 -c rake test
```

And connect from another console and type `c`:
```
$ rdbg -A 12345
[5, 14] in ~/.asdf/installs/ruby/3.0.2/bin/rake
     5| # The application 'rake' is installed as part of a gem, and
     6| # this file is here to facilitate running it.
     7| #
     8|
     9|
=>  10| version = ">= 0.a"
    11|
    12| str = ARGV.first
    13| if str
    14|   str = str.b[/\A_(.*)_\z/, 1]
=>#0    <main> at ~/.asdf/installs/ruby/3.0.2/bin/rake:10
(rdbg:remote) c    # continue command
```

Then `Errno::EADDRINUSE` was raised in server console.

```
$ rdbg -O -p 12345 -c rake test
DEBUGGER: Debugger can attach via TCP/IP (127.0.0.1:12345)
DEBUGGER: wait for debugger connection...
DEBUGGER: Connected.
/home/suke/.asdf/installs/ruby/3.0.2/lib/ruby/gems/3.0.0/gems/debug-1.3.4/lib/debug/server.rb:131: warning: assigned but unused variable - r
DEBUGGER: wait for debugger connection...
#<Thread:0x0000562eb6c48160@DEBUGGER__::Server::reader /home/suke/.asdf/installs/ruby/3.0.2/lib/ruby/gems/3.0.0/gems/debug-1.3.4/lib/debug/server.rb:43 run> terminated with exception (report_on_exception is true):
[...]

PR #1119, opened 2024-10-07
Add option to try to connect to a range of TCP ports

gerymate:
Add option to try to connect to a range of TCP ports

Fixes: #368 
Fixes: #1117

## Description
We can set an integer value with `--port-range` or `RUBY_DEBUG_PORT_RANGE` in addition to setting a TCP port for remote debugging. When `port` is unavailable, `rdbg` will try to open `port + 1`, `port + 2`, etc. until it finds a suitable port, or reaches the length of the range. This makes it possible to remote debug multiple debuggees in a multithreaded environment, but still use TCP ports in a determined way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,7 +330,9 @@ When `rdbg --attach` connects to the debuggee, you can use any debug commands (s
 
 NOTE: If you use the `quit` command, only the remote console exits and the debuggee program continues to run (and you can connect it again). If you want to exit the debuggee program, use `kill` command.
 
-If you want to use TCP/IP for the remote debugging, you need to specify the port and host with `--port` like `rdbg --open --port 12345` and it binds to `localhost:12345`.
+If you want to use TCP/IP for the remote debugging, you need to specify the port and host with `--port` like `rdbg --open --port 12345` and it binds to `localhost:12345`. You can add an optional `--port_range` option to try multiple ports in a reliable way. For example, `rdbg --open --port 12345 --port_range 10` will try to bind to 12345, 12346, 12347, ... until it finds an available port.
+
+```shell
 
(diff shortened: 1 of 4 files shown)


ko1:
Thank you for pinging me.

I'm not a network and security expert, but is it safe to try to connect not specific port?

My malicious scenario is:

1. Attacker open a port X
2. Debuggee uses port range X to X+10, and choose X+1 (because X is used by attacker's process)
3. Debugger connect with port range X to X+10 and connect to attacker's process
4. Attacker can snoop any messaging by proxying to debuggee port (X+1).

gerymate:
I'm not a security expert either, however, I don't see additional security risk from using a port range.

- The attacker can't open a port X without already being able to execute arbitrary code on the target machine with rights to open a port (as occupying a port can not be initiated from "outside", just from inside).
- When the above fails and the debugger connects to the port the attacker opened, it is still used as a one-to-one connection. I don't know the internals there, but I think the debugger has the responsibility to initiate the connection (at least in DAP), issues commands, and the attacker's process can only mimic a debuggee and reply according to the DAP, but can't query the debugger process, the flow of the communication is the other way around.

Unless there are security flaws in DAP or CDP, I don't see how could the attacker proxy traffic to another port. However, I consider the first reason to be the strongest: the developer's responsibility is to secure the dev box, it's not the task of the debug tool.

ko1:
I see. I thought the debugger client (`rdbg -A`) also use a port range feature and it can connect to the attacker's port.

However, your proposal is only server side (opening ports on debuggee) and the opening port is printed on the debuggee's console.
And the debugger client needs to select the port which is printed on the debuggee's console.

Issue #1098, opened 2024-04-23
Keystrokes between debug sessions seem to persist


joshuay03:
**Your environment**

* `ruby -v`: ruby 3.2.2 (2023-03-30 revision e51014f9c0) [arm64-darwin22]
* `rdbg -v`: rdbg 1.9.2

**Describe the bug**

After hitting a `debugger` break point once, and using `c` to move on, if you hit `Enter` before the next break point, the next breakpoint is skipped.

**To Reproduce**

Script:

```ruby
def my_method; 2.times { |i| p i; debugger; sleep 1 }; end
```

Without hitting enter:

https://github.com/ruby/debug/assets/54629302/2c5779ef-a054-4db9-97d0-1f09eaf05e49

With hitting enter during the first `sleep`:

https://github.com/ruby/debug/assets/54629302/808ccb30-456a-4888-9794-96f514f8189f

**Expected behavior**

The execution should stop at the next breakpoint regardless of what keystrokes are used between breakpoints.

**Additional context**

None.

ko1:
wow I see... Should we consume all input before REPL?

ko1:
advices are very welcome If someone knows a techniques about it on other implementations

joshuay03:
> wow I see... Should we consume all input before REPL?

Hmm possibly, is that considered unconventional? Might be worth looking at how `pry` handles this.

joshuay03:
The source of the issue:

https://github.com/ruby/debug/blob/0b77e8294b5b220b3b2a089bf09f94808654899a/lib/debug/session.rb#L1146-L1154

mattkuras:
I too recently encountered this same issue. makes it hard to clear extra lines in the console

dhh:
Been seeing the same issue. This didn't use to be the case. Maybe we can track down when this started?

joshuay03:
> Maybe we can track down when this started?

I think it was introduced [here](https://github.com/ruby/debug/pull/793). And [here's](https://github.com/ruby/debug/issues/764) a related follow-up issue.

I think a sensible middle ground would be to skip the repeat functionality on enter / empty line until the prompt is displaying. Or discard empty line input altogether until then.

dhh:
@ko1 What do you think about an option where Rails could turn off this repeater behavior? Right now ruby/debug is not all that usable in Rails because of this issue.

dhh:
When would you not want to have the no repeat thing on? I'd very much prefer if Rails could just run with the default Debug options in place.,

PR #1141, opened 2025-06-04
Clear `$stdin` before showing prompt

ko1:
Any input into `$stdin` before showing prompt will be appear on the next prompt, and entering new empty line will dispatch previous command.

This patch consumes all input to `$stdin` just before showing the prompt.


ko1:
https://github.com/ruby/debug/issues/1098

ko1:
This patch makes failure on remote connections, so I need to revisit it later.

PR #1146, opened 2025-06-13
fix hitory file path

ko1:
* Set default of `CONFIG[:history_file]` as nil, and ignore it if it is nil.
* Respect `CONFIG[:history_file]` even if the path doesn't exists.
* Respect `~/.rdbg_history` if exists for compatibility.
* Use `XDG_STATE_HOME` if there is no configuration about history file.


jasonkarns:
This is great! Do you think we need to document the fallback when XDG_STATE_HOME is unset? This PR does use the basedir specification's default (yay!), but that behavior isn't documented. Would it be welcome if I open a PR to modify the docs to that effect?

ko1:
> RUBY_DEBUG_HISTORY_FILE (history_file): history file (default: $XDG_STATE_HOME/rdbg/history)

is written. not enough?

PR #1148, opened 2025-06-17
Self-document the XDG_STATE_HOME fallback

jasonkarns:
Followup documentation/comments for #1146 

## Document XDG_STATE_HOME fallback value

The RUBY_DEBUG_HISTORY_FILE has a default value of `$XDG_STATE_HOME/rdbg/history` which is already documented. However, the fallback value for the env variable `XDG_STATE_HOME` is not documented. Users who are unfamiliar with XDG basedir spec may not know that the default value for it is `~/.local/state/`.

This change uses bash parameter expansion syntax to "self-document" the fallback value for the variable when it is unset.

Note: I used the `-` default syntax, not the `:-` syntax. This is because the current ruby code does not check if `XDG_STATE_HOME` is _empty_, it merely checks if it is _set_. So an empty-string value would not trigger the fallback location to be used.

This PR was spawned from the following discussion: https://github.com/ruby/debug/pull/1146#issuecomment-2980808756


ko1:
ok thank you!

it is another idea that adding a (sub)section "Files" and describe related files.

PR #1169, opened 2025-12-27
feat(dap): capture debuggee stdio

### LINKED ISSUE #1160: Missing DAP OutputEvents for debuggee stdout/stderr
**Your environment**

* `ruby -v`: `ruby 3.4.7 (2025-10-08 revision 7a5688e2a2) +PRISM [arm64-darwin25]`
* `rdbg -v`: `rdbg 1.11.0`

**Describe the bug**

When debugging a Ruby program with `rdbg` via DAP, the debuggee's `stdout` and `stderr` output is not sent as DAP OutputEvents. Instead, the output appears to be passed through to `rdbg` own `stdout`/`stderr` streams, which prevents it from viewing in the Debug Console of DAP clients that don't implement workarounds.

**To Reproduce**

The easiest way to reproduce is to debug any Ruby file with Zed (with installed Ruby extension):
I've created a small repo with an example: https://github.com/vitallium/rdbg-stdout-issue

Steps:

1. Clone the repo
2. Install gems with `bundle install`
3. Open in Zed
4. Open `test.rb` file
5. Press F4 to start the debugging session, and in the popup choose `Debug file`
6. Continue execution till the end.

According to the [Debug Adapter Protocol specification](https://microsoft.github.io/debug-adapter-protocol/specification#Events_Output), debug adapters should capture the debuggee's output streams and send them as `OutputEvents` with appropriate categories `stdout` and `stderr`

When a Ruby program prints to stdout/stderr, `rdbg` should send OutputEvent messages like:

```json
{
  "type": "event",
  "event": "output",
  "body": {
    "category": "stdout",
    "output": "Hello from stdout\n"
  }
}
```

**Expected behavior**

The `Console` panel should print: `Hello from stdout`, `Hello from stderr` and `After breakpoint` when the execution continues. Currently, `rdbg` does not send these OutputEvents. The debuggee's stdout/stderr is passed through to `rdbg` own stdout/stderr streams, which can be seen in adapter logs but not in the Debug Console. You can check that by opening the DAP logs in Zed via the command palette `dev: open debug adapter logs`.

**Additional context**

The `vscode-rdbg` extension actually does that: https://github.com/ruby/vscode-rdbg/blob/master/src/extension.ts#L994

I think I can work on a fix if you consider this an issue (or a feature request) with the `debug` gem. Thanks!

vitallium:
## Description

Hi! This PR adds support for capturing debuggee `stdout/stderr` output and forwarding it as [DAP output events](https://microsoft.github.io/debug-adapter-protocol/specification#Events_Output). On spawning (or attaching to a live process) spawn 2 background threads for reading from `stdout` and `stderr`, and redirecting data as DAP output events. 

## How to test

The linked issue and the fix can be easily reproduced with the Zed editor - https://zed.dev

1. Install Zed from https://zed.dev
3. Install the Ruby extension https://zed.dev/extensions/ruby
4. Clone the following repo https://github.com/vitallium/rdbg-stdout-issue
5. Install gems with `bundle install`
6. Open `test.rb` file and hit F4 to start the debugger
7. Choose `Debug file`
8. In the panels below with the original `debug` gem you should not see `Hello from stdout` and `Hello from stderr`
9. Open `Gemfile` and replace the original `debug` gem with the patched version.
10. Install gems again with `bundle install`
11. Go to the `test.rb` file again and hit F4 to start the debugger
12. Now you should see  `Hello from stdout` and `Hello from stderr` in the `Output` panel.

I understand that installing Zed and other things can be redundant so I recorded a small demo to demonstrate the fix:

https://github.com/user-attachments/assets/81fea7e4-18c7-4e85-9bf5-5081f373619a

## Additional comments

1. I've tested this in VSCode and everything seems to be working as usual.

Closes https://github.com/ruby/debug/issues/1160

Let me know what do you think. Thanks!

### DIFF
--- a/lib/debug/server_dap.rb
+++ b/lib/debug/server_dap.rb
@@ -202,7 +202,6 @@ module DEBUGGER__
       send_event 'initialized'
       puts <<~WELCOME
         Ruby REPL: You can run any Ruby expression here.
-        Note that output to the STDOUT/ERR printed on the TERMINAL.
         [experimental]
           `,COMMAND` runs `COMMAND` debug command (ex: `,info`).
           `,help` to list all debug commands.
@@ -298,6 +297,7 @@ module DEBUGGER__
         process_request(req)
       end
     ensure
+      restore_debuggee_stdio
       send_event :terminated unless @sock.closed?
     end
 
@@ -314,6 +314,7 @@ module DEBUGGER__
         UI_DAP.local_fs_map_set req.dig('arguments', 'localfs') || req.dig('arguments', 'localfsMap') || true
         @nonstop = true
 
+        capture_debuggee_stdio
         load_extensions req
 
       when 'attach'
@@ -326,6 +327,7 @@ module DEBUGGER__
           @nonstop = false
         end
 
+        capture_debuggee_stdio
         load_extensions req
 
       when 'configurationDone'
@@ -397,6 +399,7 @@ module DEBUGGER__
         terminate = args.fetch("terminateDebuggee", false)
 
         SESSION.clear_all_breakpoints
+        restore_debuggee_stdio
         send_response req
 
         if SESSION.in_subsession?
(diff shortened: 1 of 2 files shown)


lib/debug/server_dap.rb
frou: Should this part of the startup message be removed, because it implies the old behaviour and that the user manually runs the `rdbg` server in a separate terminal window/application?

vitallium: @frou yes, thanks for catching that!

Issue #1145, opened 2023-05-03
Deadlock after issuing `Net::HTTP.get` request and using a Ruby Debug breakpoint after the request


Chaseshak:
We noticed a thread deadlock issue with Ruby debug after our net/http version changed from `0.2.0` to `0.3.0`.

![Screenshot 2023-05-03 at 13 13 18](https://user-images.githubusercontent.com/13201458/236007195-52e9a07c-621c-4f3b-a786-44fbfcc4f03d.png)

When you are in the breakpoint _after_ making _any_ http request and you attempt to fetch your debug history (e.g. you arrow key up a few times), it crashes.

This error occurs in Ruby 3.1.3 which is using `net-http (0.3.0)`. It _does not_ occur in Ruby 3.1.2  (or earlier) which is using `net-http (0.2.0)`.

Reproduction script (using net-http 0.3.0 or higher):
```ruby
require "debug"
require "net/http"

Net::HTTP.get(URI.parse("https://google.com"))

binding.b
```

Stack trace:
```
["DEBUGGER Exception: .rbenv/versions/3.1.3/lib/ruby/gems/3.1.0/gems/debug-1.7.1/lib/debug/thread_client.rb:1233",
 #<fatal: No live threads left. Deadlock?
3 threads, 3 sleeps current:0x000000014a83b950 main thread:0x000000015a6048e0
* #<Thread:0x0000000100ec4d10 sleep_forever>
   rb_thread_t:0x000000015a6048e0 native:0x00000001faaf9b40 int:0

* #<Thread:0x000000010107e4a8@DEBUGGER__::SESSION@server .rbenv/versions/3.1.3/lib/ruby/gems/3.1.0/gems/debug-1.7.1/lib/debug/session.rb:177 sleep_forever>
   rb_thread_t:0x000000014a83b950 native:0x000000016f393000 int:0

* #<Thread:0x0000000104d64548@Timeout stdlib thread .rbenv/versions/3.1.3/lib/ruby/gems/3.1.0/gems/timeout-0.3.1/lib/timeout.rb:101 sleep_forever>
   rb_thread_t:0x000000014a9f1ce0 native:0x000000016f59f000 int:0

>,
 [".rbenv/versions/3.1.3/lib/ruby/gems/3.1.0/gems/debug-1.7.1/lib/debug/thread_client.rb:874:in `pop'",
[...]

PR #1175, opened 2026-04-05
fix: deadlock using IRB integration on Rails applications

jvlara:
## Related issues
Maybe https://github.com/ruby/debug/issues/1166
https://github.com/ruby/debug/issues/1145
I think there are others, but I'm not entirely sure

## Description

When using `debugger` in a Rails console with IRB integration (`RUBY_DEBUG_IRB_CONSOLE=true`), hitting a breakpoint causes a fatal deadlock: `"No live threads left. Deadlock?"`

The root cause is a race condition in `thread_stopper`: `leave_subsession` pops `@subsession_stack` and sets `@tc = nil` before `thread_stopper.disable` takes full effect. In that window, the `thread_stopper` TracePoint fires on the main thread (already back in IRB's eval loop), and since `@tc` is `nil`, the guard `tc == @tc` no longer protects it. The main thread is paused via `on_pause`, but the session server is already blocked on `@q_evt.pop` -- neither can wake the other.

The fix adds `next unless in_subsession?` as the first guard in `thread_stopper`. Once `leave_subsession` has popped the stack, `in_subsession?` returns `false` and the TracePoint becomes a no-op even if it hasn't been disabled yet.

## How to reproduce

Tested with Rails 8.1.3 and Ruby 3.3.9, but the issue is not Rails-version-specific -- it occurs whenever background threads are running (e.g. ActiveRecord Pool Reaper), which triggers the race condition in `thread_stopper`.

```bash
rails new my_app \
  --minimal \
  --api \
  --skip-git \
  --skip-docker \
  --skip-keeps \
  --skip-action-mailer \
  --skip-action-mailbox \
  --skip-action-text \
  --skip-active-job \
  --skip-active-storage \
  --skip-action-cable \
  --skip-asset-pipeline \
  --skip-javascript \
  --skip-hotwire \
  --skip-jbuilder \
  --skip-test \
  --skip-system-test \
  --skip-bootsnap \
  --skip-dev-gems \
  --skip-thruster \
  --skip-rubocop \
  --skip-brakeman \
  --skip-bundler-audit \
  --skip-ci \
  --skip-kamal \
  --skip-solid \
  --skip-bundle
```

Then modify the Gemfile:

```ruby
source "https://rubygems.org"

gem "rails", "~> 8.1.3"
gem "sqlite3", ">= 2.1"
gem "puma", ">= 5.0"
gem "tzinfo-data", platforms: %i[ windows jruby ]

gem "debug"
gem "irb"
```

Create a `script.rb` file:

```ruby
x = 1
debugger
```

Then run:

```bash
RUBY_DEBUG_IRB_CONSOLE=true bin/rails console
```

```ruby
load "script.rb"   # hits breakpoint
continue           # => "No live threads left. Deadlock?"
```

image

With this fix applied, the second `continue` returns to the prompt normally.

### DIFF
--- a/lib/debug/session.rb
+++ b/lib/debug/session.rb
@@ -1678,6 +1678,14 @@ module DEBUGGER__
 
     private def thread_stopper
       TracePoint.new(:line) do
+        # When leave_subsession pops the subsession stack and sets @tc = nil,
+        # there is a window before the TracePoint is disabled where this block
+        # can still fire on other threads. Without this guard, the main thread
+        # (back in IRB's eval loop) would be paused by on_pause, but the
+        # session server is already waiting on @q_evt.pop -- causing a mutual
+        # deadlock ("No live threads left. Deadlock?").
+        next unless in_subsession?
+
         # run on each thread
         tc = ThreadClient.current
         next if tc.management?


ko1:
Thank you！

PR #1176, opened 2026-04-14
Adapt color test expectation for IRB `>= 1.18.0`

hsbt:
## Description

https://github.com/ruby/irb/pull/1189 highlights struct member names in cyan. So, debug integration test is broken now.

https://github.com/ruby/irb/actions/runs/24364215336/job/71151532177?pr=1202

I added branch the expected value by IRB version so the test passes with both old and new IRB. 1.18.0 is TBD version, we may need to update that to `1.17.1` or something.

### DIFF
--- a/test/console/color_test.rb
+++ b/test/console/color_test.rb
@@ -33,7 +33,14 @@ module DEBUGGER__
         end
       end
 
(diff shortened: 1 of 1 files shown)


st0012:
I've updated the assertions in https://github.com/ruby/debug/pull/1177
I think the current failures are actually caused by ruby/ruby's irb still running 1.7.0

hsbt:
My fix has been submitted 1 month ago. But @ko1 didn't look this. I'm closing this.

st0012:
Oh for some reason I only got the notification today.
BTW, I think https://github.com/ruby/ruby/pull/16934 should fix ruby/ruby CI

